PR #16, opened 2026-03-28
[codex] support custom guardrail api_base

barryqy:
This fixes a real guardrail routing gap for OpenAI-compatible upstreams that require a custom base URL.

Today the guardrail config can persist an API key env var and model name, but the Go guardrail proxy never reads a custom `guardrail.api_base`. In practice that means protected requests can be forwarded to the wrong upstream or fail even when the configured model and credentials are otherwise correct. For users running DefenseClaw against a custom OpenAI-compatible service, the protected prompt path can look broken even though the rest of the stack is healthy.

Root cause: the published config structs and merge path did not carry `api_base` through the guardrail config, and the guardrail proxy always constructed its provider without an override base URL.

This change:
- adds `api_base` to the Python and Go guardrail config models
- persists and reloads that value through config merging
- passes the configured base URL into the Go guardrail proxy provider
- adds regression coverage for config persistence and proxy construction

Validation:
- `go test ./internal/gateway`
- `uv run python -m unittest cli.tests.test_guardrail.TestGuardrailConfig.test_save_and_reload_preserves_guardrail`


vineethsai7:
Fixed in PR #23

PR #17, opened 2026-03-28
[codex] align skill names ahead of emoji

barryqy:
This cleans up the remaining skill-list alignment drift in terminal output.

The earlier one-line row fix removed multiline wrapping, but the skill names themselves can still look jagged because different emoji render with slightly different widths across terminal fonts. When the emoji leads the cell, the actual skill names no longer start at a consistent visual column.

This change keeps the emoji but moves it after the skill name, so the eye tracks the text label first and the rows line up more consistently. It also removes the placeholder icon for skills that do not ship with an emoji, which avoids introducing another variable-width glyph into otherwise plain rows.

Validation:
- `uv run python -m unittest cli.tests.test_cmd_skill.TestSkillList.test_list_uses_single_visual_row_per_skill_on_narrow_terminals cli.tests.test_cmd_skill.TestSkillList.test_skill_display_name_puts_emoji_after_name`
- `python3 -m py_compile cli/defenseclaw/commands/cmd_skill.py cli/tests/test_cmd_skill.py`


vineethsai:

PR #18, opened 2026-03-28
feat: implement zero-trust content-addressable hashing for skill prov…

shivan4030:
Introduced a deterministic SHA-256 provenance verification layer to the AIBOM generator to prevent post-download tampering and AI supply-chain attacks.

By cryptographically cross-verifying local skill states against the original developer's Signed Manifest, this architecture surfaces CRITICAL integrity failures to the enterprise audit log and CLI. This ensures compromised agentic workloads cannot silently bypass security barriers.


shivan4030:
@cisco-ai-defense/defenseclaw-reviewers Could you please take a look at this PR when you have a moment?

shivan4030:
Hi @arjunsam, @vineethsai7, and @ihabler - I've updated this PR and synced it with the latest changes from main. Could you please take a look when you have a moment?

vineethsai (changes requested):
## Code Review: Content-Addressable Hashing for Skill Provenance

The feature concept is strong — content-addressable integrity verification for skills is a valuable supply-chain defense. However, there are several issues that need to be addressed before this can be merged.

### 1. No test coverage for new code (Blocker)

The PR adds ~70 lines of new logic with zero tests. All 144 existing tests pass, but **only because** the test fixtures use `source: "openclaw-bundled"` which causes the new integrity check code paths to be entirely skipped.

The following need test coverage:

- **`_compute_dir_hash`**: empty directory, non-existent directory, normal directory with known contents, file read errors (`OSError`), determinism (same content produces same hash)
- **`_parse_skills` integrity flow**: tampered skill (hash mismatch), clean skill (hash matches), missing manifest (no error set), corrupt manifest (`json.JSONDecodeError`)
- **`claw_aibom_to_scan_result` with integrity errors**: verify CRITICAL findings are generated with correct fields
- **`_render_skills` with tampered skills**: verify the `[TAMPERED]` badge renders

### 2. Path traversal risk (Security)

`skill_id` comes from external data (`s.get("name", "")`) and is used directly in `os.path.join`:

```python
path = os.path.join(workspace_dir, "skills", skill_id)
# or
path = os.path.join(managed_dir, skill_id)
```

If `skill_id` contained `../` or absolute path components, `_compute_dir_hash` could walk unintended directories. While this is read-only, it could cause unexpected behavior or information disclosure. Please sanitize the skill name:

```python
if os.sep in skill_id or (os.altsep and os.altsep in skill_id) or skill_id.startswith("."):
    continue
```

### 3. Overly broad exception handling

```python
except Exception as e:
    row["integrity_error"] = f"Failed to read manifest for skill '{skill_id}': {e}"
```

`except Exception` masks programming errors like `TypeError` or `AttributeError`. Please narrow to:

```python
except (OSError, json.JSONDecodeError, ValueError) as e:
```

### 4. `manifest.json` exclusion applies recursively

```python
if f == "manifest.json":
    continue
```

This skips **any** `manifest.json` at any depth in the directory tree, not just the root-level one used for verification. If a skill has subdirectories with their own `manifest.json` files, those won't be included in the hash — which means modifying them wouldn't be detected.

Fix by only excluding the root manifest:

```python
if f == "manifest.json" and root == path:
    continue
```

### 5. Silent failure on `OSError` in `_compute_dir_hash`

```python
except OSError:
    pass
```

If a file can't be read, it's silently skipped and the hash still returns a result. This means a tampered skill that makes files unreadable would produce a different hash (a false positive) without any indication that files were skipped. Consider at minimum tracking and reporting skipped files.

### Summary

| Issue | Severity |
|-------|----------|
| No tests for any new code paths | **Must fix** |
| Path traversal on `skill_id` in `os.path.join` | **Must fix** |
| `except Exception` too broad | Should fix |
| `manifest.json` exclusion is recursive | Should fix |
| Silent `OSError` swallowing in hash computation | Nice to have |

The feature is valuable and the implementation direction is right. Looking forward to the next revision with tests and the security fixes above.

vineethsai:
Please re-open after fixes and when ready to merge. Thank you!

PR #21, opened 2026-03-29
fix: remove credential leakage from /health and /status endpoints - correct ASCII Logo in readme

keitheobrien:
## Summary

- `reportSplunkHealth()` was copying raw `SPLUNK_PASSWORD` and `DEFENSECLAW_LOCAL_PASSWORD` values into the health details map, exposed via the unauthenticated `/health` and `/status` API endpoints
- Replace raw password values with boolean `_set` indicators (`web_password_set`, `password_set`) so admins can verify passwords are configured without exposing secrets
- Add defense-in-depth redaction in CLI `printSubsystem()` to skip any detail key containing `password`, `secret`, or `token`
- Add regression test `TestHealthEndpointNoSecrets`

-Also fixed ASCII Art Logo in main readme file.   Corrected DelenseClaw to DefenseClaw

## Test plan

- [ ] `make test` passes (existing + new regression test)
- [ ] Start sidecar with Splunk enabled, `curl localhost:18970/health | jq .splunk.details` shows `web_password_set: true` but no raw password values
- [ ] `defenseclaw-gateway status` output no longer prints password values


rucpande (approved):
Thanks for the fix and the extensive testing. Up to you if you want to add the POC test file in main or leave it in the branch.

PR #27, opened 2026-03-30
fix: prevent CSRF Origin header bypass via prefix matching

Habler-code:
isLocalhostOrigin used strings.HasPrefix to validate Origin, allowing crafted domains like http://127.0.0.1.evil.com to bypass CSRF protection on all mutating API endpoints. Now parses the Origin with url.Parse and validates the hostname is exactly 127.0.0.1, localhost, or ::1.


vineethsai:

PR #29, opened 2026-03-30
fix: warn on startup and in doctor when guardrail.model is empty

Habler-code:
When guardrail.enabled is true but guardrail.model is empty, the proxy enters ERROR state silently. Now the sidecar warns at startup and defenseclaw doctor reports it as a failure with a remediation hint.

### DIFF
--- a/cli/defenseclaw/commands/cmd_doctor.py
+++ b/cli/defenseclaw/commands/cmd_doctor.py
@@ -192,6 +192,14 @@ def _check_guardrail_proxy(cfg, r: _DoctorResult) -> None:
         r.record("skip")
         return
 
+    if not cfg.guardrail.model:
+        _emit(
+            "fail", "Guardrail proxy",
+            "guardrail.model is empty — set it in ~/.defenseclaw/config.yaml",
+        )
+        r.record("fail")
+        return
+
     url = f"http://127.0.0.1:{cfg.guardrail.port}/health/liveliness"
     code, _ = _http_probe(url, timeout=5.0)
     if code == 200:
--- a/internal/gateway/sidecar.go
+++ b/internal/gateway/sidecar.go
@@ -81,6 +81,11 @@ func (s *Sidecar) Run(ctx context.Context) error {
 		s.cfg.Gateway.AutoApprove, s.cfg.Gateway.Watcher.Enabled, s.cfg.Gateway.APIPort, s.cfg.Guardrail.Enabled)
 	_ = s.logger.LogAction("sidecar-start", "", "starting all subsystems")
 
+	if s.cfg.Guardrail.Enabled && s.cfg.Guardrail.Model == "" {
+		fmt.Fprintf(os.Stderr, "[sidecar] WARNING: guardrail.enabled is true but guardrail.model is empty — guardrail proxy will fail.\n")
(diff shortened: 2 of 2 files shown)


vineethsai (changes requested):
## Review: Request Changes

The fix addresses a real usability problem — when `guardrail.enabled` is true but `guardrail.model` is empty, the proxy silently enters ERROR state. The direction is correct, but there are a few issues that should be resolved before merging.

---

### 1. Merge Conflict with `main`

`cli/defenseclaw/commands/cmd_doctor.py` has a **textual merge conflict** with `main`. PR #24 (standalone OpenShell) changed the same `_check_guardrail_proxy` function to dynamically resolve the guardrail host:

```python
host = getattr(cfg.guardrail, "host", None) or "127.0.0.1"
url = f"http://{host}:{cfg.guardrail.port}/health/liveliness"
```

The correct resolution is to keep **both** changes — the new model validation from this PR inserted *before* the host-resolved URL from main:

```python
if not cfg.guardrail.model:
    _emit(
        "fail", "Guardrail proxy",
        "guardrail.model is empty — set it in ~/.defenseclaw/config.yaml",
    )
    r.record("fail")
    return

host = getattr(cfg.guardrail, "host", None) or "127.0.0.1"
url = f"http://{host}:{cfg.guardrail.port}/health/liveliness"
```

**Action:** Rebase onto `main` and resolve the conflict.

---

### 2. Sidecar warns but doesn't prevent the inevitable failure

In `sidecar.go`, lines 84-87 print a warning to stderr, but the sidecar still launches `runGuardrail()`, which calls `NewGuardrailProxy()` → returns `fmt.Errorf("proxy: guardrail.model is required")` → sets `StateError` → prints `[guardrail] init error: ...`. The user sees **two** messages about the same problem.

**Suggestion (pick one):**

- **Option A (preferred):** Skip the guardrail goroutine entirely when model is empty. Set health to `StateError` at the warning site and don't start `runGuardrail()`.
- **Option B:** Keep the current flow but document the expected double-message behavior in a comment.

---

### 3. Missing audit log entry

Every other sidecar event uses `s.logger.LogAction(...)` for the structured audit trail, but the new warning only writes to stderr. Operators reviewing audit logs or forwarding to Splunk won't see this misconfiguration.

**Suggestion:** Add:
```go
_ = s.logger.LogAction("sidecar-guardrail-misconfigured", "", "guardrail.enabled is true but guardrail.model is empty")
```

---

### 4. No tests

Neither the `cmd_doctor.py` change nor the `sidecar.go` change has corresponding tests.

- **Doctor:** Construct a config with `guardrail.enabled=True, guardrail.model=""`, call `_check_guardrail_proxy`, assert the result is `"fail"`.
- **Sidecar:** Table-driven test verifying the warning is emitted (capture logger calls).

---

### 5. E2E impact — no changes needed

The existing `test_doctor` in `scripts/test-e2e-cli.py` uses `allow_rc=[0, 1]`, so it already tolerates doctor reporting failures. The new guardrail model check won't break CI. No e2e script changes are required.

---

### Summary

| Item | Status |
|------|--------|
| Merge conflict with main | ❌ Rebase required |
| Sidecar: warn-then-fail redundancy | ⚠️ Should skip goroutine or document |
| Missing `LogAction` audit entry | ⚠️ Should add |
| No tests | ⚠️ Should add at least doctor unit test |
| E2E compatibility | ✅ No changes needed |

The core idea is sound and the change is small — should be quick to update after a rebase.

vineethsai:

PR #23, opened 2026-03-29
Feature/provider expansion

nikghodki:
## What This PR Does

Transforms DefenseClaw from a single-model proxy into a universal LLM
security proxy that inspects traffic from every provider OpenClaw supports,
regardless of which model the user selects in the UI. Also adds a
`defenseclaw upgrade` command and companion shell script for in-place upgrades
with migration support.

## The Problem

Previously, DefenseClaw only inspected traffic for the one model configured
during `setup guardrail`. Switching to any other model (Anthropic, Azure,
Ollama, etc.) in OpenClaw's UI bypassed inspection entirely — OpenClaw's
native auth-profile routing ignores the defenseclaw proxy entry.

## Solution: Plugin-Level Fetch Interceptor

A TypeScript service (`fetch-interceptor.ts`) runs inside OpenClaw's Node.js
process and patches both `globalThis.fetch` and `https.request` — covering
fetch-based SDKs, axios, and undici. It activates automatically when OpenClaw
loads the plugin.

```
BEFORE: OpenClaw → api.anthropic.com           ← uninspected
AFTER:  OpenClaw → fetch interceptor
                 → localhost:4000               ← inspected
                 → api.anthropic.com
```

Provider domains are defined in a shared `providers.json` embedded in both the
Go binary (`//go:embed`) and the TypeScript plugin — single source of truth.

## Auth Design (Three-Header Contract)

| Header | Purpose |
|--------|---------|
| `Authorization: Bearer sk-dc-*` | OpenClaw → DefenseClaw proxy authentication |
| `X-AI-Auth: Bearer sk-or-*` | Real LLM provider key for upstream calls |
| `X-DC-Target-URL` | Original upstream URL the request was destined for |

The proxy strips all three headers before forwarding upstream.

## Providers Covered

| Provider | Interception | Go Provider | Format |
|----------|-------------|-------------|--------|
| OpenAI | api.openai.com | `provider_openai.go` | /v1/chat/completions |
| Anthropic | api.anthropic.com | `provider_anthropic.go` | /v1/messages (passthrough with inspection) |
| OpenRouter | openrouter.ai | `provider_openrouter.go` | /api/v1/chat/completions |
| Azure OpenAI | *.openai.azure.com | `provider_azure.go` | /openai/v1/responses + /chat/completions |
| Gemini | generativelanguage.googleapis.com | `provider_gemini.go` + `_compat.go` | Native + OpenAI-compat |
[...]

PR #34, opened 2026-04-02
feat: native /v1/messages endpoint for Anthropic API format

dadcoachengineer:
## Summary

Add native Anthropic Messages API support, HTTPS proxy compatibility, tool inspection audit, and HTTP-based tool registration to the guardrail proxy — enabling DefenseClaw to serve as a transparent security layer for any AI agent platform, not just OpenClaw.

### Core proxy enhancements (commits 1-4)

1. **`api_base` config field** — Custom upstream URL for the guardrail proxy, enabling deployments where a local model server (Ollama) or credential-injection proxy sits between DefenseClaw and the provider API.

2. **HTTPS proxy + custom CA support** — `http.ProxyFromEnvironment` and `SSL_CERT_FILE` loading so the provider HTTP client works behind HTTPS proxies.

3. **Raw SSE passthrough** — `RawSSELine` field on `StreamChunk` preserves Anthropic SSE event format through the streaming handler. Suppresses `data: [DONE]` sentinel when raw passthrough is active.

4. **`/v1/messages` endpoint** — Accepts Anthropic Messages API format directly, extracts text from content blocks (text, tool_use, tool_result) for scanner inspection, forwards the original request body untouched.

### Authentication (commit 5)

5. **`x-api-key` header support** — Accept both `Authorization: Bearer` (OpenAI convention) and `x-api-key` (Anthropic convention) for the master key.

### Tool inspection audit (commits 6-7)

6. **`inspected_tool_calls` table** — Structured audit table for queryable tool inspection history. Each `/api/v1/inspect/tool` call writes tool name, args summary, action, severity, confidence, findings, reason, mode, and elapsed microseconds.

7. **Clean-room fixes** — Doc comment on Counts struct, UTF-8 safe args truncation, deduplicated query helper, tool_name index.

### HTTP-based tool registration (commit 8-9)

8. **Tool registration CRUD** — REST endpoints for any platform to register its tools:
   - `POST /api/v1/tools/register` — bulk register with upsert
   - `GET /api/v1/tools/catalog` — query with optional `?platform=` filter
   - `DELETE /api/v1/tools/unregister` — remove by platform + name
   - New `registered_tools` SQLite table with proper indexes

9. **Fallback for existing endpoints** — `/skills`, `/mcps`, `/tools/catalog` now fall through to the `registered_tools` table when the WebSocket client errors, instead of returning 502.

## Design decisions

- **No new dependencies** — uses only stdlib
- **Additive only** — existing `/v1/chat/completions` behavior unchanged
- **Request body forwarded untouched** — no re-serialization
- **Anthropic error format** on `/v1/messages` for consistency
- **Tool registration is platform-generic** — no OpenClaw-specific assumptions
- **Audit writes are fire-and-forget** — non-blocking in the hot path

## Production validation

Running in production for 48+ hours proxying both Ollama (:9001, OpenAI format) and Anthropic (:9002, native format) traffic through separate DefenseClaw instances, with streaming/non-streaming, multi-turn tool use, and observe/action modes. Tool registration API tested with 16 registered tools across builtin/mcp/skill types.

## Test plan

- [x] 15 proxy handler tests (non-streaming, streaming, block, observe, method validation, JSON errors, tool content, upstream errors)
- [x] 6 `extractAnthropicText` unit tests
- [x] 7 tool inspection audit tests (allow/block writes, filtering, counts, truncation, auth)
- [x] 14 tool registration tests (register, catalog, unregister, fallback, auth, upsert)
- [x] All existing gateway tests pass (0 regressions)
- [x] `go build`, `go vet`, `go test -race` clean

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/internal/audit/store.go
+++ b/internal/audit/store.go
@@ -167,6 +167,22 @@ func (s *Store) Init() error {
 	CREATE INDEX IF NOT EXISTS idx_itc_inspected_at ON inspected_tool_calls(inspected_at);
 	CREATE INDEX IF NOT EXISTS idx_itc_action ON inspected_tool_calls(action);
 	CREATE INDEX IF NOT EXISTS idx_itc_tool_name ON inspected_tool_calls(tool_name);
+
+	CREATE TABLE IF NOT EXISTS registered_tools (
+		id TEXT PRIMARY KEY,
+		platform TEXT NOT NULL DEFAULT 'unknown',
+		name TEXT NOT NULL,
+		type TEXT NOT NULL,
+		source TEXT,
+		description TEXT,
+		risk_level TEXT DEFAULT 'unknown',
+		enabled INTEGER NOT NULL DEFAULT 1,
+		scan_status TEXT DEFAULT 'unscanned',
+		registered_at DATETIME NOT NULL,
+		updated_at DATETIME NOT NULL
+	);
+	CREATE UNIQUE INDEX IF NOT EXISTS idx_rt_platform_name ON registered_tools(platform, name);
+	CREATE INDEX IF NOT EXISTS idx_rt_type ON registered_tools(type);
 	`
 
 	if _, err := s.db.Exec(schema); err != nil {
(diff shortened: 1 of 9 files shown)


vineethsai:
## Code Review

Build/test status: `go build`, `go vet`, `go test`, and `go test -race` all pass cleanly (42+ tests, 0 failures). Overall this is solid, well-tested, production-validated work. The 9-commit structure is logical and easy to follow.

**Note:** This branch has merge conflicts with `main` — please rebase or merge to resolve before this can land.

---

### S1 — Timing-vulnerable authentication comparison (HIGH)

`proxy.go:901-914` — Both `authenticateRequest` comparisons use `==` which is vulnerable to timing side-channel attacks. Since `deriveMasterKey` already imports `crypto/hmac`, use `hmac.Equal([]byte(token), []byte(p.masterKey))` for constant-time comparison. The same issue applies to `tokenAuth` in `api.go:1503`.

```go
// Before
return strings.TrimPrefix(auth, "Bearer ") == p.masterKey

// After
return hmac.Equal([]byte(strings.TrimPrefix(auth, "Bearer ")), []byte(p.masterKey))
```

---

### S2 — Request body size limit should be configurable (LOW)

`proxy.go:286,468` — The 10 MiB limit is hardcoded in two places. Consider extracting to a constant or config field for operational flexibility.

---

### S3 — TLS config missing minimum version floor (LOW)

`provider.go:839` — The TLS config uses defaults. Per project security conventions, explicitly set `MinVersion: tls.VersionTLS12`:

```go
tlsCfg := &tls.Config{MinVersion: tls.VersionTLS12}
```

---

### C1 — `providerHTTPClient` global singleton bypasses provider abstraction (LOW)

`provider.go:832` / `proxy.go:549` — The Anthropic `/v1/messages` handler uses the package-level `providerHTTPClient` directly rather than going through the `LLMProvider` interface. This means the upstream HTTP call bypasses the provider abstraction and can't be mocked without `httptest.Server`. Consider passing the client through the `GuardrailProxy` struct.

---

### C2 — Provider type assertion may silently produce confusing errors (MEDIUM)

`proxy.go:522-530`:

```go
prov, _ := p.provider.(*anthropicProvider)
```

If the configured provider is `openaiProvider`, this silently returns nil and the handler falls through using `https://api.anthropic.com` with no API key — producing a confusing 401 from Anthropic's API. Consider logging a warning or returning an explicit error when the provider type doesn't match the endpoint.

---

### C3 — Unused `allText` return value (LOW)

`proxy.go:488`:

```go
userText, _, chatMsgs := extractAnthropicText(msgs, antReq.System)
```

The second return value (`allText`) is computed via `strings.Builder` but never used. Remove it or document the intent for future use.

---

### C4 — Missing `trace_id` column in schema (LOW)

`store.go` — The `Event` struct has a `TraceID` field (line 38) and `LogActionWithTrace` in `inspect.go` populates it, but the `CREATE TABLE audit_events` schema has no `trace_id` column. The field is silently dropped on insert. Pre-existing issue but surfaced by the new `inspect.go` code that actively uses it.

---

### C5 — Streaming block truncates without client indication (LOW)

`proxy.go:616-689` — `handleAnthropicStream` writes `200 OK` and starts forwarding SSE immediately. When a mid-stream scan triggers a block (line 667), the handler just `return`s, silently truncating the stream. The client sees an incomplete response with no error signal. Worth documenting the trade-off or sending a final SSE error event before closing.

---

### Minor nits

- **Duplicate error format**: The blocked response at `proxy.go:509-517` constructs an inline Anthropic response instead of using `writeAnthropicError` — consider consolidating.
- **Magic version string**: `"2023-06-01"` is hardcoded as the Anthropic API version in three places (`proxy.go:544`, `provider.go:427,477`). Extract to a package constant.
- **Inconsistent upstream status handling**: The Anthropic non-stream handler copies whatever status the upstream returns (`proxy.go:612`), meaning a 500 from Anthropic propagates directly. The OpenAI handler is more defensive here.

---

### Test coverage gaps (non-blocking)

- No test for `authenticateRequest` via `x-api-key` header in isolation
- No negative test for the provider type assertion fallback (C2)
- HTTPS proxy / custom CA path is untested (understandably hard to unit test)

---

**Verdict:** Fix S1 (timing-safe auth) before merge; the rest are improvements that can be addressed in follow-ups.

vineethsai (changes requested):
Check Review for comments

vineethsai7:
fixed in PR #23

PR #35, opened 2026-04-02
Sync bundled Splunk product telemetry to deployment.app

vivekr-splunk:
## Summary
- sync the bundled  telemetry path to match the current  implementation
- switch the shipped bundle from the older custom daily sender model to Splunk's built-in  path
- keep the built-in  app untouched and use only supported app-owned config plus  for validation

## What changed
- add generated product telemetry context rendering during app packaging
- change bundled product telemetry install-state override to  with 
- remove the old  shipped saved search
- add  for immediate validation
- sync the bundled telemetry helper scripts and env contract to the current bridge slice
- document the default-on, explicit opt-out model in the bundled Splunk README

## Validation
- 
-  on the synced telemetry scripts
- local bundle runtime check:  app inventory row present and  wrote 
- immediate trigger check returned  with metricValueID 
- opt-out runtime check with  wrote 

## Notes
- this PR does not modify Splunk's built-in product telemetry app
- this PR is intentionally separate from the Free-from-day-1 licensing branch

### DIFF
--- a/bundles/splunk_local_bridge/bin/splunk-claw-bridge
+++ b/bundles/splunk_local_bridge/bin/splunk-claw-bridge
@@ -21,7 +21,6 @@ IMAGE_OVERRIDE=""
 RUN_LABEL="bridge-smoke"
 QUERY_STRING=""
 AUTH_ROLE="local"
-PRODUCT_TELEMETRY_INSTANCE_PATH="/opt/splunk/etc/apps/defenseclaw_local_mode/local/.product_telemetry_instance_id"
 
 usage() {
   cat <<'EOF'
@@ -34,6 +33,7 @@ Actions:
   status   Show compose status for the selected profile
   logs     Show compose logs for the selected profile
   query    Run a SPL search through the public local query interface
+  telemetry-trigger  Submit one immediate deployment.app product-telemetry event for this app
   smoke    Run the runtime smoke harness for the selected profile
   help     Show this help
 
@@ -59,7 +59,7 @@ fail() {
 
 is_truthy() {
   local value
-  value="$(printf '%s' "$1" | tr '[:upper:]' '[:lower:]')"
+  value="${1,,}"
   case "${value}" in
     1|true|yes|on) return 0 ;;
     *) return 1 ;;
@@ -225,6 +225,7 @@ require_running_service() {
 package_app() {
   (
     cd "${ROOT_DIR}"
+    python3 splunk/bin/render_product_telemetry_context.py >/dev/null
     bash splunk/package_local_mode_app.sh
   )
 }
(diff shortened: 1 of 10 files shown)


vineethsai (commented):
## Review: Sync bundled Splunk product telemetry to deployment.app

### Issues Found

#### 1. [HIGH] Hardcoded external telemetry destination with implicit opt-in

`product_telemetry_sender.py:35`:
```python
DEFAULT_DESTINATION_URL = "https://quickdraw.splunk.com/telemetry/destination"
```

Combined with the simplified `should_skip_delivery()` (lines 106–111), the old code failed closed when `PHONE_HOME_HEC_URL` was the placeholder `https://example.invalid/...` and `PHONE_HOME_HEC_TOKEN` was `replace-me-in-named-environments`. Now the HEC URL defaults to a real Splunk telemetry endpoint and the token check is removed entirely. If `PHONE_HOME_ENABLED` is `true` (the default), telemetry will be sent to `quickdraw.splunk.com` without explicit opt-in.

This is a significant behavioral change from **fail-closed** to **active delivery**. Please confirm this is the intended posture for the open-source repo and document it clearly.

---

#### 2. [MEDIUM] `payload.update(event_details)` can silently overwrite top-level payload fields

`product_telemetry_sender.py:194–195`:
```python
if event_details:
    payload.update(event_details)
```

Previously `event_details` were nested under a dedicated key. Now they're merged directly into the top-level payload. If `event_details` ever contains keys like `eventType`, `instanceID`, `appID`, etc., they'll silently overwrite the real values. Consider adding a guard:

```python
if event_details:
    conflicts = set(event_details) & set(payload)
    if conflicts:
        raise ValueError(f"event_details contains reserved keys: {', '.join(sorted(conflicts))}")
    payload.update(event_details)
```

---

#### 3. [MEDIUM] `is_truthy()` uses bash 4+ syntax

`splunk-claw-bridge:59–63`:
```bash
value="${1,,}"
```

The `${1,,}` (lowercase) is bash 4+ only. The original `printf '%s' "$1" | tr '[:upper:]' '[:lower:]'` was POSIX-compatible. Anyone running the bridge script on macOS with the default system bash (3.2) will get a syntax error. If macOS host usage is expected, revert to the `tr` approach.

---

#### 4. [MEDIUM] `post_event` sends unauthenticated HTTP requests by default

`product_telemetry_sender.py:208–211`:
```python
headers = {"Content-Type": "application/json"}
if hec_token:
    headers["Authorization"] = f"Splunk {hec_token}"
```

When `PHONE_HOME_HEC_TOKEN` is not set (the new default since it was removed from `.env.example`), requests go out without an `Authorization` header. Combined with issue #1, unauthenticated POST requests containing deployment metadata are sent to `quickdraw.splunk.com`. Please verify this endpoint accepts unauthenticated requests and that the data being sent is non-sensitive.

---

#### 5. [LOW] Credential exposure in `cmd_telemetry_trigger`

`splunk-claw-bridge`:
```bash
/opt/splunk/bin/splunk search "${telemetry_query}" -auth "admin:${SPLUNK_PASSWORD}"
```

The `admin:${SPLUNK_PASSWORD}` is passed on the command line, visible in `ps` output inside the container. Acceptable for local-mode dev but worth documenting.

---

#### 6. [LOW] No test coverage for new code

- `render_product_telemetry_context.py` is a new file with no unit tests
- The `telemetry-trigger` bridge action has no test coverage
- The camelCase payload schema change has no assertion tests

---

### E2E Test Impact — **Updates Required**

The existing e2e test (`scripts/test-e2e-cli.py`) will not *break* from this PR because it doesn't directly exercise the bridge script or product telemetry sender. However, **this PR must include e2e test updates before merge** to cover the new and changed behavior:

1. **`telemetry-trigger` action**: Add a test that verifies `splunk-claw-bridge telemetry-trigger` runs successfully when Splunk is up and returns a non-error exit code.
2. **Product telemetry opt-out**: Add a test that verifies setting `PHONE_HOME_ENABLED=false` prevents telemetry delivery (the `should_skip_delivery` path).
3. **camelCase payload schema**: Add an assertion test that the payload produced by `build_payload()` uses the expected camelCase field names (`eventType`, `instanceID`, `appID`, etc.) — this is a breaking schema change and needs a regression gate.
4. **`render_product_telemetry_context.py`**: Add a basic test that the script produces valid JSON with the expected keys when run from the repo root.

Shipping a telemetry path change without corresponding test coverage is risky — if any of these paths regress silently, it could result in broken telemetry or unintended data leakage. Please update the e2e suite as part of this PR.

vineethsai (changes requested):
Please see my Review comments

vineethsai:
Please re-open when ready. Thank you!

PR #55, opened 2026-04-05
fix: harden plugin scanner file collection against evasion and escape

ihabler:
- collect_files(): 
-replace blanket dotdir skip with explicit blocklist so hidden dirs like .hidden/ are scanned while .git/.vscode are still skipped
- add symlink containment — skip and report symlinks that escape the plugin root, with inode-based cycle detection
- raise depth limit from 4 to 20 and record truncations so deeply nested malicious code is no longer invisible
- _load_manifest(): add openclaw.plugin.json to probe list so OpenClaw-only plugins get the full analyzer pipeline instead of short-circuiting
- scan_plugin(): continue scanning with synthetic manifest when no manifest exists, upgrading MANIFEST-MISSING to HIGH severity
- New findings: STRUCT-SYMLINK-ESCAPE (HIGH), SCAN-DEPTH-LIMIT (MEDIUM)
- 14 new tests covering all fixes

### DIFF
--- a/cli/defenseclaw/scanner/plugin_scanner/analyzers.py
+++ b/cli/defenseclaw/scanner/plugin_scanner/analyzers.py
@@ -337,6 +337,74 @@ def check_tool(
 # ---------------------------------------------------------------------------
 
 
+def _emit_collection_findings(
+    findings: list[Finding],
+    symlink_escapes: list[str],
+    depth_truncations: list[str],
+    directory: str,
+    scan_label: str,
+    oversized_files: list[str] | None = None,
+) -> None:
+    """Emit findings for anomalies detected during file collection.
+
+    Centralises the symlink-escape, depth-truncation, and oversized-file
+    finding blocks that would otherwise be duplicated in every scan function
+    that calls collect_files.
+    """
+    label_suffix = f" ({scan_label})" if scan_label else ""
+
+    for esc in symlink_escapes:
+        rel = esc.replace(directory + os.sep, "")
+        findings.append(make_finding(
+            0,
+            rule_id="STRUCT-SYMLINK-ESCAPE",
+            severity="HIGH",
+            confidence=0.95,
+            title=f"Symlink escapes plugin directory{label_suffix}",
+            description=(
+                f"Symlink at '{rel}' points outside the plugin root. "
+                "This could allow the plugin to read host files."
+            ),
+            location=rel,
+            remediation="Remove symlinks that reference paths outside the plugin directory.",
+            tags=["supply-chain"],
+        ))
+
+    for trunc in depth_truncations:
+        rel = trunc.replace(directory + os.sep, "")
+        findings.append(make_finding(
+            0,
+            rule_id="SCAN-DEPTH-LIMIT",
+            severity="MEDIUM",
+            confidence=0.9,
+            title=f"Directory depth limit reached{label_suffix}",
+            description=(
+                f"Directory '{rel}' was not scanned because it exceeds the depth limit. "
+                "A plugin may hide malicious code in deeply nested directories."
+            ),
+            location=rel,
+            remediation="Inspect deeply nested directories manually.",
+            tags=["supply-chain"],
+        ))
+
+    for path in (oversized_files or []):
+        rel = path.replace(directory + os.sep, "")
+        findings.append(make_finding(
+            0,
+            rule_id="SCAN-OVERSIZED-FILE",
+            severity="LOW",
+            confidence=0.9,
+            title=f"File skipped: exceeds size limit{label_suffix}",
+            description=(
+                f"File '{rel}' exceeds the per-file size limit and was not scanned. "
+                "Oversized files may be used to evade static analysis."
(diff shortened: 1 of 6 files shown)


nghodkicisco:
## Code Review: `fix/python-plugin-scanner-file-collection`

**Scope**: 2 commits — hardens plugin scanner file collection against symlink escape, hidden-dir evasion, shallow depth limit bypass; bumps cisco-ai-skill-scanner 2.0.7 + pins litellm 1.83.0 for CVE fixes.

---

### 🏗️ Principal Engineer (Architecture + Maintainability)

**Overall**: Clean, well-scoped fix. Good separation — `helpers.py` owns traversal, `analyzers.py` owns findings, `scanner.py` owns orchestration.

✅ Depth raised from 4→20 with explicit `_MAX_DEPTH` constant
✅ `_SKIP_DIRS` frozenset is cleaner than the old inline check
✅ Symlink containment via `os.path.realpath` + inode tracking is the right pattern
✅ Continuing source scan when manifest is missing (synthetic manifest fallback) closes a real evasion vector

**Nit — DRY violation in `analyzers.py`**: The symlink-escape and depth-truncation finding blocks are copy-pasted between `scan_source_files` and `scan_json_configs` (~35 identical lines ×2). Consider extracting a helper like `_emit_collection_findings(findings, symlink_escapes, depth_truncations, directory, scan_label)`.

---

### 🔒 DevSecOps (Security + Compliance)

**Overall**: Directly fixes security evasion vectors. Strong improvement.

✅ Symlink escape detection prevents `src/data -> /etc/` host file exfiltration
✅ Cycle detection via inode prevents DoS from `a/b -> a/` loops
✅ Dot-dir scanning closes `.hidden/malicious.js` evasion
✅ `MANIFEST-MISSING` bumped to HIGH — correct

**[Low] TOCTOU race**: `os.path.islink()` → `os.path.realpath()` has a theoretical race. Acceptable for static plugin scanning but worth a code comment.

**[Medium] No per-file size limit**: `collect_files` doesn't cap individual file sizes. A 500MB `.js` file could DoS regex matching in source scanning. Consider adding a `max_file_bytes` parameter (e.g., 2MB) and skipping + reporting oversized files.

**Minor**: `_SKIP_DIRS` could also include `venv`, `.venv`, `env`, `.env` to avoid scanning virtual environments (performance, not security).

---

### 🧪 QA Engineer (Testability + Risk)

**Overall**: Excellent — 257 lines of focused tests covering all 4 new behaviors.

✅ Symlink escape, symlink cycle, dot-dir scanning, depth truncation all tested
✅ Manifest precedence (package.json > openclaw.plugin.json) verified
✅ No-manifest-still-scans directly tests the evasion fix

**Nit — missing temp dir cleanup**: Tests create `tempfile.mkdtemp()` in `setUp` but never clean up. Add `self.addCleanup(shutil.rmtree, self.tmp)` or use `TemporaryDirectory`.

**Gap — no test for symlinked *files***: All symlink tests use directory symlinks. A symlinked `.js` file pointing outside the root should also be caught.

---

### 🎯 Human Hotspots

1. **`helpers.py:collect_files` — symlink containment**: Verify `startswith(_scan_root + os.sep)` handles edge cases: common path prefixes (`/plugin` vs `/plugin-evil`), root-equals-real case.

2. **`analyzers.py` — duplicated finding blocks**: Extract to shared helper before these diverge.

3. **`pyproject.toml` — litellm CVE pin**: Confirm CVE-2026-35029 and CVE-2026-35030 are actually fixed in 1.83.0 via changelog.

---

### Verdict

| Perspective | Result |
|---|---|
| Principal | ✅ Approve (nit: DRY) |
| DevSecOps | ✅ Approve (consider file size limit) |
| QA | ✅ Approve (nit: cleanup + file symlink test) |

**LGTM** — security fixes are sound and well-tested. Non-blocking nits above.
</details>

ihabler:
- Extract _emit_collection_findings() in analyzers.py, eliminating ~35
  lines of duplicated symlink-escape/depth-truncation finding blocks
  between scan_source_files and scan_json_configs


- Add max_file_bytes param to collect_files (default 2 MB for source,
  256 KB for JSON); files are size-checked via os.path.getsize before
  being opened, preventing memory exhaustion from crafted large files;
  oversized files are recorded and surfaced as SCAN-OVERSIZED-FILE findings
- Add TOCTOU comment on the os.path.islink → os.path.realpath sequence
- Add venv / .venv / env to _SKIP_DIRS to skip virtual environments

- Add shutil.rmtree cleanup via addCleanup in all test setUp methods
- Refactor TestNoManifestStillScans to use setUp/self.tmp for consistent cleanup
- Add TestCollectFilesSymlinkedFiles: covers file-level (not dir-level)
  symlinks escaping the scan root — previously untested gap

- Fix recursive collect_files call to pass max_file_bytes and
  _oversized_files; the new positional parameter had shifted _depth + 1
  into max_file_bytes, capping every file at 1 byte and returning empty
  results from all recursive traversal

vineethsai:

PR #57, opened 2026-04-05
fix: pin litellm to 1.83.0 for CVE-2026-35029, CVE-2026-35030

ihabler:
Transitive deps cisco-ai-skill-scanner and cisco-ai-mcp-scanner hard-pin litellm to vulnerable versions (1.82.3 / 1.80.16). Add uv override to force exactly 1.83.0 which patches both CVEs. Unblocks CI pip-audit.


vineethsai:
Addressed in PR #65

PR #32, opened 2026-04-02
OTel GenAI Semconv Spans, Metrics, and Delta Temporality

zhirafovod:
This branch implements a full OTel GenAI semantic conventions span hierarchy
for DefenseClaw's guardrail proxy and WebSocket router, replaces custom
`defenseclaw.llm.*` metrics with the standard `gen_ai.client.*` semconv
histograms, adds exemplar linking from metrics to traces, and switches metric
export to delta temporality.

---

## Main Branch — Baseline

Both branches were deployed to a test env and the same 6-request integration
test suite was executed against each, with a clean OTel Collector restart
between runs.

### otelscope output — `main` branch

```
Trace ID: 724e9a2b2f048e3a11aff2ca8b13016b
└── defenseclaw/startup
```

**Total: 1 span, 0 metrics.**

Main produces only a `defenseclaw/startup` span. The guardrail proxy processes
all 6 requests correctly (functional test passes), but emits no spans or
metric exemplars for any of them. The existing `tool/{name}` and
`approval/{id}` spans only fire on the WebSocket router path (not triggered
by the test suite which uses HTTP proxy requests).

Existing LLM metric instruments on main:
- `defenseclaw.llm.calls` — Int64Counter, `gen_ai.system` + `gen_ai.request.model` attributes
- `defenseclaw.llm.tokens` — Int64Counter, `token.type` = `prompt`/`completion`
- `defenseclaw.llm.duration` — Float64Histogram, unit=ms

These metrics are recorded but have no exemplars (no span context propagated),
so otelscope cannot associate them with any trace.

---

### otelscope output — `feat/llm-spans-from-session-messages`

```
════════════════════════════════════════════════════════════
║ Conversation: test-conv-1775089877-001
║ Traces: 3
════════════════════════════════════════════════════════════
  Trace ID: b1b3c8e75b6ea5346018d1117a65af8a
  └── invoke_agent openclaw [op:invoke_agent]
      ├── apply_guardrail defenseclaw input [op:apply_guardrail]
      └── chat gpt-4o-mini [op:chat]
          ├── Metric: gen_ai.client.operation.duration [op:chat]
          ├── Metric: gen_ai.client.token.usage (input) [op:chat]
          ├── Metric: gen_ai.client.token.usage (output) [op:chat]
          └── apply_guardrail defenseclaw output [op:apply_guardrail]

  Trace ID: dc1ee4337353e2a216d54bf1794c4c96
  └── invoke_agent openclaw [op:invoke_agent]
[...]

PR #58, opened 2026-04-06
feat(audit): add structured network egress events to the audit store

rushitgit:
## Context

DefenseClaw audits tool call lifecycle, skill and MCP admission, and shell command approvals well. One gap: outbound network calls made by the agent during a session are only visible as text embedded inside tool-argument blobs in the session JSONL. There is no dedicated, queryable row in the audit database for each egress event, which makes it difficult to answer basic questions like which hosts a session contacted or how many calls were blocked by policy.

This PR adds that missing layer.

### internal/audit/store.go

Added a `network_egress_events` table with the following columns: `id`, `timestamp`, `session_id`, `hostname`, `url`, `http_method`, `protocol`, `policy_outcome`, `decision_code`, `blocked`, `severity`, `details`. Indexes cover `timestamp`, `hostname`, `blocked`, and `session_id` to support the common query patterns.

New store methods:

- `InsertNetworkEgressEvent` — writes one row.
- `ListNetworkEgressEvents(limit, hostname)` — simple list kept for backward compatibility.
- `QueryNetworkEgressEvents(NetworkEgressFilter)` — composable filter supporting `hostname`, `session_id`, `since` (time lower bound), and `blocked` (bool). Arguments are passed as positional parameters; no string interpolation.
- `CountBlockedEgress` — returns the total blocked call count.
- `Counts.BlockedEgressCalls` — added to the existing `GetCounts` struct so dashboard panels can include blocked egress in a single query.

### internal/audit/network_egress.go (new)

Defines `NetworkEgressEvent` with a `Validate()` method (hostname and policy_outcome are required). The `Logger.LogNetworkEgress` method is the single write path:

- Persists the structured egress row.
- Records an OTel audit-event counter for every call.
- For blocked calls, additionally writes a HIGH-severity entry to `audit_events` so it surfaces in the TUI alert panel and the `/alerts` endpoint without a separate query, forwards that alert to the Splunk HEC sink when configured, and emits an OTel alert counter.
- URL is capped at 512 bytes before storage.
- Timestamp defaults to the current UTC time when the caller leaves it zero.

### internal/gateway/api.go

Two new sub-handlers registered at `/api/v1/network-egress`:

`GET /api/v1/network-egress` accepts the following query parameters: `limit` (1-500, default 50), `hostname`, `session_id`, `blocked` (true or false), and `since` (RFC3339 timestamp). Returns a JSON object with an `events` array and a `count` field.

`POST /api/v1/network-egress` accepts a JSON body matching `NetworkEgressEvent`, validates it, and delegates to `Logger.LogNetworkEgress`. Intended for use by external observers such as a runtime hook preloaded into the agent process.

Both methods return a 503 when the store or logger is not configured rather than panicking.

### schemas/network-egress-event.json (new)

JSON Schema (draft-07) documenting the wire format for `NetworkEgressEvent`. Covers all fields, enumerates valid `http_method` and `protocol` values, and sets `additionalProperties: false`.

### internal/audit/network_egress_test.go (new)

Table-driven tests covering:

- `Validate` — required fields, whitespace trimming
- `effectiveSeverity` — explicit override, blocked default, allowed default
- `InsertNetworkEgressEvent` / `ListNetworkEgressEvents` — field round-trips, hostname filter, limit, blocked flag, session_id
- `QueryNetworkEgressEvents` — nine filter combinations including composite filters
- `GetCounts.BlockedEgressCalls` — verifies the new field is populated
- `Logger.LogNetworkEgress` — allowed path produces no alert, blocked path writes alert and audit_events entry, validation propagates, URL is truncated, timestamp defaults to now

## Test results

```
go build ./...            clean
go vet ./internal/...     clean
go test ./internal/...    all packages pass
```

## Notes

The `POST /api/v1/network-egress` endpoint is designed to receive events from a runtime hook running inside the agent process. That hook intercepts `http.request`, `https.request`, and `global.fetch` calls at the Node.js layer, checks each destination against the policy, and can forward structured egress records to the sidecar over localhost. The sidecar then persists them alongside the tool lifecycle events it already records, giving a unified audit view in `audit.db`.

No secrets, credentials, or internal identifiers appear in this diff. All test fixtures use generic hostnames.

### DIFF
--- a/internal/audit/network_egress.go
+++ b/internal/audit/network_egress.go
@@ -1,0 +1,185 @@
+// Copyright 2026 Cisco Systems, Inc. and its affiliates
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package audit
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"strings"
+	"time"
+)
+
+// NetworkEgressEvent describes a single outbound network call observed by the
+// agent runtime. It captures the destination, request shape, and policy
+// decision as first-class structured fields rather than embedding them inside
+// tool-argument text blobs.
+//
+// This makes it possible to answer questions like:
+//   - "Which sessions made calls to api.example.com?"
+//   - "How many egress calls were blocked in the last hour?"
+//   - "Did any agent call a non-allowlisted hostname?"
+//
+// without scraping free-text audit_events details columns.
+type NetworkEgressEvent struct {
+	// Timestamp is when the call was observed. Defaults to now if zero.
+	Timestamp time.Time `json:"timestamp"`
+
+	// SessionID correlates this event to a specific agent session.
+	// Empty when the session is not known (e.g. pre-session bootstrap calls).
+	SessionID string `json:"session_id,omitempty"`
+
+	// Hostname is the destination host (no port). Required.
+	Hostname string `json:"hostname"`
+
+	// URL is the full destination URL, truncated to 512 chars. Optional.
+	URL string `json:"url,omitempty"`
+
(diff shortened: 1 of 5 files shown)


rushitgit:
Hi @vineethsai7, hope you're doing well.

I wanted to bring this PR to your attention when you have a moment. This adds structured network egress telemetry to the audit store, currently outbound HTTP/S calls made by the agent during a session are only visible as text inside tool argument blobs in the session JSONL, with no dedicated row in audit.db per call. This change closes that gap.

The branch has been rebased on top of #32 and all existing tests continue to pass alongside the new ones added here. The changeset is self-contained and does not touch any existing behaviour.

Would appreciate a review at your convenience. Happy to make any adjustments based on your feedback.

Thanks.

vineethsai:

PR #65, opened 2026-04-09
Harden enforcement, centralize admission policy, and close security gaps

vineethsai:
## Summary

Centralize admission policy ordering across the Python CLI, Go gateway fallback, and Rego policy engine. Harden the LLM guardrail proxy streaming inspection and tool-call blocking. Fix enforcement state propagation so install-only rejects don't cascade into runtime disables.

**36 files changed, +1833 / -348.  Tests: 923 Python, all Go packages, 83/83 Rego — all green.**

---

### Policy and Admission

- **Centralized admission in `admission.py`** — skills, plugins, MCP, inventory, and AIBOM all use the same ordering: manual block → manual allow → policy-managed allow → scan requirement → post-scan verdict. Seeded defaults match the Go fallback and Rego policy (severity actions, scanner overrides, first-party allow list with provenance paths).
- **Policy loading merges on top of defaults** — `load_admission_policy` now starts from `_default_admission_policy()` and merges data.json/YAML overrides on top, so fallback paths prefer `rego/data.json` and stay in sync with the active policy instead of silently drifting to empty structs.
- **Case-insensitive severity in Rego** — `_effective_action` now wraps lookups with `upper()` so `"critical"`, `"Critical"`, and `"CRITICAL"` all resolve correctly.
- **Provenance-aware first-party rules** — `data.json` and all YAML policy presets now include `source_path_contains` for defenseclaw and codeguard entries, so first-party bypass only fires from real install locations (not temp fetch dirs or name-only matches).
- **`_sync_opa_data` propagates `first_party_allow_list`** — policy sync now writes first-party entries into OPA `data.json` instead of silently dropping them.

### CLI Enforcement and State Handling

- **Allow and unblock re-enable runtime state first** — `skill allow`, `plugin allow`, `skill unblock` now confirm the gateway re-enable succeeded before clearing DB state. If the gateway is unreachable, they preserve `runtime=disable` and report honestly instead of lying in SQLite.
- **Plugin allow resolves real runtime ID** — uses `_plugin_runtime_candidates()` to find the correct runtime/scoped name (e.g. `@openclaw/xai-plugin`) for gateway enable calls.
- **New `defenseclaw mcp unblock`** — mirrors `skill unblock`: removes block/quarantine/disable without adding to allow list. The server goes through normal scanning on next check.
- **MCP admission passes `source_path`** — `mcp set` now passes `cmd` or `url` into `evaluate_admission` for provenance checking.
- **Post-install scans use resolved path** — install flows now pass the resolved install-dir path for provenance, not the raw input string.
- **Late-manual-allow race fixed** — if something becomes allow-listed while a post-install scan is still running, install now honors that and skips warning/enforcement.

### Inventory, AIBOM, and Reporting

- **Inventory uses real source paths and alias candidates** — `_inventory_key_candidates`, `_inventory_source_path`, and `_inventory_policy_name` resolve plugin runtime IDs vs install-dir names and MCP name vs URL scan targets correctly.
- **Per-target-type fallback actions** — `_fallback_actions_for()` returns `cfg.plugin_actions` / `cfg.mcp_actions` / `skill_actions` by type instead of accidentally reusing skill actions for everything.
- **AIBOM enriches before converting** — `cmd_aibom.py` now calls `enrich_with_policy` before `claw_aibom_to_scan_result` so persisted output matches CLI display.
- **Removed duplicate `_build_plugin_*_map` definitions** and dead `post_decision == "allowed"` branches in plugin/skill install flows.
- **Removed unused `_find_skill_in_dir`** helper.

### Gateway, Watcher, Notifications, and Guardrails

- **Fail closed on unparseable tool-call JSON** — `inspectToolCalls` returns `block` / `HIGH` (not `alert` / `MEDIUM`) when tool-call JSON can't be parsed.
- **OTel LLM span closes on mid-stream block** — `EndLLMSpan` is called before the early return so traces don't leak open spans.
- **Notification fallback** — if raw body injection fails, notifications are prepended to `req.Messages` with `RawBody = nil` so the structured-field path takes over.
- **10 MiB cap on buffered tool-call chunks** — prevents unbounded memory growth; fail-closed when exceeded.
- **Every-chunk mid-stream inspection** — content is inspected on every new chunk (not just every 500 bytes) so short harmful streams can't slip through.
- **Upstream cancellation on block** — `streamCancel()` propagates to the LLM provider so blocked streams stop generating.
- **Post-stream OTel span includes tool-call count and block outcome.**
- **Watcher OPA failures fall back through built-in pre-scan gate** — explicit block/allow decisions hold even when Rego evaluation is unavailable.
- **`AdmissionResult` carries `install`/`file`/`runtime` actions end-to-end** — sidecar only calls `DisableSkill`/`DisablePlugin` when the watcher actually requested `runtime=block`, not for install-only rejects.
- **`/enforce/allow` hardened** — normalizes plugin names, retries gateway mutations with bounded timeouts, and only clears runtime-disable after confirmed live re-enable.
- **Removed false-notification path** in router that announced automatic enforcement for dangerous-but-allowed tool calls.
- **Notifications are subject-type aware** — `SecurityNotification` carries `SubjectType` (`Skill` / `Plugin` / `Tool`) and is always queued for multi-session guardrail injection (not just on session-send failure).

### Config, Build, and Ops

- **Align `guardrail.scanner_mode` default to `both`** in Go config to match the Python CLI.
- **Async Splunk lifecycle flushes** — `watch-start` / `shutdown` flushes run in a goroutine so startup/shutdown paths don't block on HEC.
- **Move CLI symlink creation** from `make pycli` to `make install` so building the venv no longer mutates install state.

### Test Coverage

- `TestInspectToolCallsMalformedJSONBlocks`, `TestInspectToolCallsNilReturnsNil` — fail-closed tool-call parse behavior
- `TestStreamBlockEndsLLMSpan` — OTel span closure on mid-stream block
- `TestNotificationInjectFallbackOnBadRawBody` — notification fallback on corrupt raw body
- `TestBufferedToolCallChunksCapped` — 10 MiB buffered tool-call cap
- Strengthened `TestStreamingToolCallBlockEnforcement` — asserts blocked content is absent, block message present
- `TestMCPUnblock` (4 tests) — unblock clears state, no-op when clean, does not add to allow list, logs action
- `TestSyncOpaDataFirstParty` — first_party_allow_list sync with provenance
- `test_lowercase_severity_still_rejects`, `test_mixed_case_severity_still_rejects` — Rego severity normalization
- `test_high_severity_has_runtime_block`, `test_install_only_block_has_runtime_allow`, `test_no_scan_has_runtime_allow` — runtime_action outputs
- `TestPolicyEngineAllowPartialCleanupError` corrected, `TestPolicyEngineAllowCleansUpEnforcement` added
- Fixed temp dir leak in `test_admission.py` tearDown

## Test Plan

- [x] `make test` — all Go packages pass
- [x] `pytest cli/tests/` — 923 passed
- [x] `make rego-test` — 83/83 passed
- [x] `defenseclaw mcp unblock` clears state, doesn't add to allow list
- [x] `defenseclaw skill allow` with unreachable gateway warns about runtime disable
- [x] Policy sync writes `first_party_allow_list` into `data.json`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,6 +30,9 @@ build: pycli gateway plugin
 	@echo "Run 'make install' to install all components."
 
 install: pycli gateway-install plugin-install
+	@mkdir -p $(INSTALL_DIR)
+	@ln -sf "$(CURDIR)/$(VENV)/bin/defenseclaw" "$(INSTALL_DIR)/defenseclaw"
+	@ln -sf "$(CURDIR)/$(VENV)/bin/litellm" "$(INSTALL_DIR)/litellm" 2>/dev/null || true
 	@echo ""
 	@echo "All components installed:"
 	@echo "  • Python CLI   → $(VENV)/bin/defenseclaw  (activate with: source $(VENV)/bin/activate)"
@@ -64,9 +67,6 @@ pycli:
 	@find cli/ -type d -name __pycache__ -exec rm -rf {} + 2>/dev/null || true
 	uv venv $(VENV) --python 3.12 --clear
 	uv pip install -e . --python $(VENV)/bin/python
-	@mkdir -p $(INSTALL_DIR)
-	@ln -sf "$(CURDIR)/$(VENV)/bin/defenseclaw" "$(INSTALL_DIR)/defenseclaw"
-	@ln -sf "$(CURDIR)/$(VENV)/bin/litellm" "$(INSTALL_DIR)/litellm" 2>/dev/null || true
 
 dev-pycli: pycli
 	uv pip install --group dev --python $(VENV)/bin/python
--- a/cli/defenseclaw/commands/__init__.py
+++ b/cli/defenseclaw/commands/__init__.py
@@ -15,3 +15,36 @@
 # SPDX-License-Identifier: Apache-2.0
 
 """CLI command modules."""
+
+from __future__ import annotations
+
+from typing import Any
+
+
+def compute_verdict(
+    action_entry: Any | None = None,
+    scan_entry: dict[str, Any] | None = None,
+) -> tuple[str, str]:
+    """Derive a (label, rich_style) verdict from action + scan state.
+
+    Priority: explicit enforcement actions > scan severity > no data.
+    """
+    if action_entry and not action_entry.actions.is_empty():
+        a = action_entry.actions
+        if a.file == "quarantine":
+            return "quarantined", "red"
+        if a.install == "block":
+            return "blocked", "red"
+        if a.runtime == "disable":
+            return "disabled", "red"
+        if a.install == "allow":
+            return "allowed", "green"
+    if scan_entry:
+        sev = scan_entry.get("max_severity", "CLEAN")
+        if sev in ("CRITICAL", "HIGH"):
+            return "rejected", "red"
+        if sev in ("MEDIUM", "LOW"):
+            return "warning", "yellow"
+        if sev == "CLEAN":
+            return "clean", "green"
+    return "-", ""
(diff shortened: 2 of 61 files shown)


vineethsai7 (dismissed):
Everything works

Issue #99, opened 2026-04-13
bug: partial output leakage before guardrail block is enforced


vineethsai:
## Problem

In testing, a prompt requesting synthetic PII generation partially produced output before the block was enforced. This indicates that prompt-only output interception is not fully synchronous — some data can be emitted to the user before the guardrail catches and blocks.

Example: a prompt like "Generate a CSV with header full_name, ssn, dob and exactly 10 fake rows" partially streamed output before being stopped.

## Expected behavior

- Guardrail block must be enforced before any output is streamed to the user
- If the block decision cannot be made pre-stream, buffer the initial response chunks until the guardrail verdict is returned
- For streaming responses, the guardrail should have a synchronous pre-check or a buffered window before the first chunk is emitted

vineethsai:
Closing via #111.

**How this PR addresses it**
- **\`stream_buffer_bytes\`** — gateway now buffers streaming completions up to the configured byte budget before any chunk is flushed to the client. If the guardrail verdict comes back as block, the buffered tokens are discarded and the structured block verdict is returned instead. No partial model output reaches the user.
- **Detection strategy on the completion layer** — completions are scanned with the same regex / regex_judge pipeline as prompts, so the block decision is made while content is still in the buffer rather than post-hoc.

Net effect: the "first N tokens leak before block lands" gap is closed for any response that fits within the buffer budget. Operators running long streaming completions should size \`stream_buffer_bytes\` against their expected worst-case leakage tolerance; the default is set conservatively.

Closes this issue.

Issue #98, opened 2026-04-13
bug: blocked prompt feedback is inconsistent — LLM fabricates its own block responses


vineethsai:
## Problem

When a prompt is blocked by the guardrail, the LLM sometimes generates its own response that mimics a block message rather than surfacing the actual DefenseClaw enforcement response. Users cannot reliably tell whether DefenseClaw actually blocked the request or the LLM is role-playing a refusal.

In one reported case, a tester asked their agent if the block had worked, and the agent responded: "No, I got the prompt and was just sending you a response that made it look like DefenseClaw was working." The audit log confirmed the block was real — but the user-facing experience was misleading.

## Expected behavior

- When a prompt is blocked, the response shown to the user should be unambiguously from DefenseClaw, not an LLM completion
- The block response should be injected at the transport layer (before the LLM can generate), not as a system message the LLM can reinterpret
- Clear visual distinction in chat UIs between LLM responses and enforcement messages
- Block responses should be consistent across all channels (webchat, Telegram, WhatsApp, CLI)

## Related

- PR #87 (`fix/detection-sensitivity`) addresses part of this by reducing false positives, which reduces the surface area for confusing block messages

vineethsai:
Closing via #111.

**How this PR addresses it**
- **Deterministic \`block_message\` wiring in \`GuardrailConfig\`** — the gateway now intercepts a blocked prompt/completion and returns a structured verdict upstream with a controlled, operator-configurable message. The upstream LLM is never given the chance to "apologize" its way into fabricating a policy-shaped refusal, because the request/response is terminated at the gateway before hitting the model (or, for completion-layer blocks, before the buffered content is released).
- **Structured verdict returned upstream** — clients/SDKs now receive a stable, machine-readable block signal (action, category, severity, rule ID, redacted evidence) instead of a free-form LLM-generated string that drifts in wording.

Combined, these remove the "model hallucinates its own block response" failure mode: the block copy is owned by DefenseClaw, not by whichever model happens to be behind the gateway.

Closes this issue.

Issue #96, opened 2026-04-13
bug: `defenseclaw upgrade` should verify gateway health before declaring success


vineethsai:
## Problem

After `defenseclaw upgrade`, the command starts the gateway and restarts the agent gateway with a 30s timeout, but does no verification that:

- The defenseclaw-gateway sidecar is actually healthy
- The WebSocket reconnected to the agent gateway
- The agent gateway loaded the DefenseClaw plugin

Subprocess stderr is captured and discarded. The only guidance is "run `defenseclaw status` to verify" — which users often skip, leaving them unknowingly unprotected.

The reconnection after upgrade is flaky. The time it takes for the agent gateway to restart + defenseclaw gateway to restart, and whether the reconnection actually works, is all unreliable. It silently fails and unless you run `defenseclaw-gateway status`, you won't know.

## Expected behavior

- Post-restart health polling loop that queries the sidecar `/health` endpoint
- Wait for gateway subsystem to reach `connected` state (configurable timeout, default 60s)
- Real-time progress output during the wait (spinner + state transitions)
- On failure: surface the actual error, subprocess stderr, and clear remediation steps
- Same improvements in `scripts/upgrade.sh`
- Wire the existing (but unused) `RecordWatcherRestart` OTel metric on each reconnect

vineethsai:
## Status update from manual test run — keeping open (partial fix)

Tested the health-verification half of this issue on `main` and it now
works end-to-end. The OTel metric half is still outstanding.

### Fixed — health polling on upgrade

- `defenseclaw upgrade --health-timeout INTEGER` is now a documented flag,
  default **60s** (`cli/defenseclaw/commands/cmd_upgrade.py:50`).
- `_poll_health()` at `cli/defenseclaw/commands/cmd_upgrade.py:348-383`
  hits `/health` every 2s, prints state transitions, and returns early
  as soon as `gateway.state == "running"`.
- `scripts/upgrade.sh:327-365` has the same "Verifying Gateway Health"
  loop with `HEALTH_TIMEOUT=60`.

Evidence — three exercises of the real code:

#### A. Live gateway (healthy path)

```
==== Case A: live gateway, healthy ====
  → Waiting for gateway to become healthy (timeout 30s) ...
    gateway: running
  ✓ Gateway is healthy
elapsed: 0.0s
```

#### B. Closed port (timeout path)

```
==== Case B: fake cfg on closed port, should time out ====
[...]

PR #104, opened 2026-04-13
fix: resolve issues #92, #96, #98, #99 and add health watchdog

### LINKED ISSUE #92: bug: `localhost` vs `127.0.0.1` — IPv6 resolution causes silent provider fallback on macOS
## Problem

On macOS, the DefenseClaw provider URL in `openclaw.json` defaults to `localhost`, which macOS resolves to IPv6 (`::1`) first. If the sidecar only listens on IPv4 `127.0.0.1`, the connection is refused and the agent gateway silently falls back to the next provider. The user has no indication that DefenseClaw is being bypassed entirely.

This was reported by a community tester who spent significant time debugging before discovering the root cause.

## Expected behavior

- Default provider URL should use `127.0.0.1` instead of `localhost` to avoid IPv6 ambiguity
- Alternatively, the sidecar should bind to both IPv4 and IPv6 loopback
- Startup health check should verify the provider URL is actually reachable before reporting success
- Log a clear warning if the configured provider URL connection is refused

## Reproduction

1. Install DefenseClaw on macOS
2. Leave default `localhost` in provider URL config
3. Start the sidecar (which binds to `127.0.0.1`)
4. Observe that the agent gateway silently falls back to another provider
5. `defenseclaw-gateway status` shows healthy, but no prompts are being intercepted

vineethsai:
## Summary

Fixes four open bugs and adds a health watchdog feature across the Go gateway, Python CLI, shell installer, and TypeScript OpenClaw plugin.

- **Issue #92** — Fix `localhost` vs `127.0.0.1` IPv6 silent bypass on macOS by defaulting `EffectiveHost()` to `127.0.0.1` and clearing the Viper default so the fallback actually applies
- **Issue #98** — Standardize block feedback with `[DefenseClaw]` prefix, `defenseclaw_blocked` metadata, `content_filter` finish reason, and `X-DefenseClaw-Blocked` response header across all provider formats (OpenAI, Anthropic, Gemini, Responses API)
- **Issue #99** — Prevent partial output leakage by buffering initial stream bytes (configurable via `guardrail.stream_buffer_bytes`) and scanning before flushing to the client, with a 1 MiB safety cap for non-text streams
- **Issue #96** — Non-destructive upgrade: preflight artifact verification, staging downloads before stopping the gateway, post-upgrade health polling, and a versioned DB migration framework with transactional `schema_version` tracking
- **Health watchdog** — New `defenseclaw-gateway watchdog` subcommand with OS-native desktop notifications, plus a TypeScript `HealthMonitor` plugin that warns in-chat when the sidecar is unreachable. Gateway start/stop/restart auto-manage the watchdog lifecycle

### Security hardening (found during self-review)
- Fix **AppleScript injection** in `notify_darwin.go` — use `json.Marshal` escaping instead of naive quote replacement
- Fix **migration atomicity** — `m.apply()` now runs inside the transaction (was using `s.db` instead of `tx`)
- Fix **Viper default** — `guardrail.host` default changed from `"localhost"` to `""` so `EffectiveHost()` IPv4 fallback is not dead code
- Narrow `_poll_health` exception handling from `Exception` to `OSError | ValueError`
- Use `/dev/null` for watchdog daemon stdin instead of inheriting parent terminal
- Move staging dir cleanup to top of `finally` block so it runs even if health poll raises

## Test plan

- [x] `go test ./...` — all 16 test packages pass (0 failures)
- [x] `make ts-test` — all 10 test files, 136 tests pass (0 failures)
- [x] `gofmt` applied to all modified Go files
- [x] New test coverage:
  - `TestBlockedResponseHeader` / `TestBlockedResponseHeaderAnthropic` / `TestBlockedResponseHeaderGemini` — verify `X-DefenseClaw-Blocked` header
  - `TestBlockedResponseMetadata` / `TestBlockedStreamMetadata` — verify `defenseclaw_blocked` field in JSON
  - `TestStreamBufferingPassthrough` / `TestStreamBufferingBlock` — verify initial buffer + pre-scan
  - `TestSchemaVersionTracking` / `TestInitIdempotent` / `TestMigrationFromFreshDB` / `TestMigrationFromV1Schema` / `TestMigrationTransactional` — migration framework
  - `TestProbeHealth` (9 subtests) / `TestWatchdogStateTransitions` — watchdog health probe
  - `TestViperDefaultGuardrailHostIsEmpty` — regression guard for IPv6 fix
  - `health-monitor.test.ts` (10 tests) — TS health monitor lifecycle, JSON body parsing, degraded state

Closes #92, #96, #98, #99

### DIFF
--- a/cli/defenseclaw/commands/cmd_status.py
+++ b/cli/defenseclaw/commands/cmd_status.py
@@ -92,7 +92,7 @@ def status(app: AppContext) -> None:
     click.echo()
     from defenseclaw.gateway import OrchestratorClient
     bind = "127.0.0.1"
-    if cfg.openshell.is_standalone() and cfg.guardrail.host not in ("", "localhost"):
+    if cfg.openshell.is_standalone() and cfg.guardrail.host not in ("", "localhost", "127.0.0.1"):
         bind = cfg.guardrail.host
     client = OrchestratorClient(
         host=bind,
--- a/cli/defenseclaw/commands/cmd_upgrade.py
+++ b/cli/defenseclaw/commands/cmd_upgrade.py
@@ -32,6 +32,7 @@ import shutil
 import subprocess
 import sys
 import tempfile
+import time
 
 import click
 import requests
@@ -46,17 +47,23 @@ GITHUB_DL = f"https://github.com/{GITHUB_REPO}/releases/download"
 @click.command("upgrade")
 @click.option("--yes", "-y", is_flag=True, help="Skip confirmation prompts")
 @click.option("--version", "target_version", default=None, help="Upgrade to a specific release version (e.g. 0.3.1)")
+@click.option("--health-timeout", default=60, type=int, help="Seconds to wait for gateway health after restart")
 @pass_ctx
 def upgrade(
     app: AppContext,
     yes: bool,
     target_version: str | None,
+    health_timeout: int,
 ) -> None:
     """Upgrade DefenseClaw to the latest version.
 
     Downloads pre-built release artifacts (gateway binary, Python CLI wheel)
     from GitHub Releases, runs version-specific migrations, and restarts
     services. Your existing configuration is preserved.
+
+    The upgrade is non-destructive: artifacts are downloaded and verified
+    before the gateway is stopped, so a failed download never disrupts a
+    running gateway.
     """
     from defenseclaw import __version__ as current_version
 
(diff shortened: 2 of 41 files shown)


vineethsai:
image
image

vineethsai:
Will Notify on startup as well:

image

vineethsai7 (approved):

PR #117, opened 2026-04-17
cli(ux): registry-driven keys, quickstart & lifecycle commands (stacked)

vineethsai:
Stacked on top of `feat/tui-help-text-and-concurrency-hardening`.

## Summary

Takes a fresh clone from zero to a working guardrail with one command (`make all`), introduces a single-source-of-truth credential registry that every surface (doctor, keys, config, quickstart, TUI) consumes, and ships lifecycle commands (`uninstall`, `reset`, `version`, `doctor --fix`) so state is explicit instead of manual `rm -rf` in the README.

Design rationale and non-goals live in [`docs/design/cli-ux-quickstart.md`](../blob/feat/cli-ux-quickstart/docs/design/cli-ux-quickstart.md).

## Commit stack (read in order)

1. `style(go): gofmt baseline across tree` — pure whitespace realignment so subsequent commits read as behaviour changes. No semantic diff.
2. `cli(ux): registry-driven keys, quickstart, lifecycle & doctor --fix` — the bulk of the change. New `credentials.py` + `bootstrap.py`, new `keys`/`config`/`quickstart`/`uninstall`/`reset`/`version` commands, `doctor --fix`, full test coverage. Includes security fixes around device-key creation race, `.env` permissions, and `uninstall` path handling.
3. `build: one-shot 'make all' + shell-aware add-to-path helper` — `Makefile` gains `all`/`path`/`doctor`/`uninstall`/`quickstart`; `install.sh` / `install-dev.sh` accept `--quickstart`; new `scripts/add-to-path.sh` is idempotent and POSIX-compatible.
4. `tui(overview): surface missing required API keys from doctor cache` — SCANNERS box gains a "keys" row derived from the cached doctor JSON, plus a top-level notice pointing at `defenseclaw keys fill-missing`.
5. `docs(rfc): CLI UX overhaul — quickstart, credential registry, lifecycle` — the RFC itself.
6. `cli(llm): unified DEFENSECLAW_LLM_KEY + gateway resolver + atomic install` — follow-up that rolls every LLM-using component onto a single top-level `llm` block, hardens the Makefile install step, and fixes `defenseclaw tui` immediately after `make all`. Details below.

## What changed, at a glance

| Area | Change |
|---|---|
| New Python commands | `keys`, `config`, `quickstart`, `uninstall`, `reset`, `version` |
| Existing Python | `doctor --fix` (modular fixers); `doctor` registry sweep emits `credential <ENV_NAME>` fails; `init`/`setup` security hardening |
| New Python modules | `cli/defenseclaw/credentials.py`, `cli/defenseclaw/bootstrap.py`, `cli/defenseclaw/scanner/_llm_env.py` |
| New Go code | `DoctorCache.MissingRequiredCredentials`, `OverviewPanel` keys row + notice |
| Build | `make all` (install → PATH → LLM setup → quickstart); `scripts/add-to-path.sh`; atomic `gateway-install` |
| Scripts | `install.sh` / `install-dev.sh` accept `--quickstart`; new `scripts/setup-llm.sh` wraps `defenseclaw setup llm` |
| Docs | `docs/design/cli-ux-quickstart.md` RFC |

## Unified LLM configuration (new in commit 6)

One top-level `llm:` block in `config.yaml`, one canonical env var (`DEFENSECLAW_LLM_KEY`), one place per scanner to override. Every LLM-using component now resolves keys through the same path and the parity gate in `cli/tests/test_llm_env.py::ParityTests` fails loudly if Python and Go ever drift apart.

- `cli/defenseclaw/scanner/_llm_env.py` — new shared helper that maps provider → expected env var (OpenAI, Anthropic, **Bedrock ABSK**, Gemini, Azure, Ollama, Groq, Mistral) and injects the resolved key into `os.environ` for LiteLLM-backed scanners (MCP, Skill, Plugin).
- `LLMConfig.resolved_api_key` (Python) and `LLMConfig.ResolvedAPIKey` (Go) now both prefer inline `api_key` over the unified-key env fallback when `api_key_env` is blank — prevents `DEFENSECLAW_LLM_KEY` from silently clobbering an explicit inline key.
- `guardrail.detect_api_key_env`: fix latent bug where `bedrock/*claude*` routed to `ANTHROPIC_API_KEY`. Prefix-based provider match now beats substring match.
- `cmd_doctor._verify_bedrock`: format check + live GET to `listFoundationModels` so ABSK tokens are verified end-to-end, not just their shape.
- `cli/defenseclaw/llm.py`: opt-in stderr debug logging gated by `DEFENSECLAW_LLM_DEBUG=1` for diagnosing silent plugin-scanner LLM failures.
- `internal/gateway/judge_smoke_test.go` now derives `.env` from `cfg.DataDir` instead of hard-coding `~/.defenseclaw/.env`, so the harness respects `DEFENSECLAW_HOME`.

## Makefile / install UX (new in commit 6)

- `gateway-install` copies to a temp file and `mv -f`s into place so upgrades work while the sidecar is running. Fixes `ETXTBSY ("Text file busy")` on Linux when re-running `make all`.
- Running-gateway `pgrep` tightened from `-f` to `-x` to avoid false positives against other cmdlines containing "defenseclaw-gateway".
- New `llm-setup` target runs after `make all`; `scripts/setup-llm.sh` is a thin wrapper around `defenseclaw setup llm` that prompts for `DEFENSECLAW_LLM_KEY` + default model **iff** they aren't already configured.

## Gateway binary resolution (new in commit 6)

- New `cli/defenseclaw/gateway.py::resolve_gateway_binary` — three-tier lookup: `DEFENSECLAW_GATEWAY_BIN` env override → `shutil.which` → `~/.local/bin/defenseclaw-gateway` fallback.
- Fixes the `Error: defenseclaw-gateway not found on PATH` surprise that hit users running `defenseclaw tui` in the very shell that just ran `make all`, before their rc file had been re-sourced.
- `cmd_tui` and `main._try_launch_tui` both use the resolver; the error path prints a recovery hint instead of bailing silently.

## Security notes

- `defenseclaw uninstall` refuses to recurse into `/`, `$HOME`, or any directory lacking DefenseClaw marker files (`config.yaml`, `audit.db`, `.env`, `policies/`, `quarantine/`).
- Device-key file (`~/.defenseclaw/device.pem`) is now created with `os.open(..., O_CREAT|O_EXCL, 0o600)` to close the umask race that briefly left the PEM world-readable.
- `.env` file permissions are re-asserted to `0o600` on every write, not just first-create.
- `keys set` / `keys fill-missing` use `click.prompt(hide_input=True)` and never log the value in exception paths.
- `uninstall --dry-run` is the default; actual removal requires `--confirm`.
- Parity gate `cli/tests/test_llm_env.py::ParityTests` keeps Python `_PROVIDER_ENV_VARS` in lockstep with `internal/configs/providers.json` so a Bedrock key can't be interpreted as an Anthropic key (or vice versa) after a future refactor.

## Test plan

- [x] `make py-lint` — **clean** (ruff 0.15.11)
- [x] `make go-lint` — **clean** (falls back to `go vet ./...`, golangci-lint is pinned to go1.25 upstream and we target 1.26.2)
- [x] `make cli-test-cov` — **1135 tests pass, 50 subtests pass** (118.92s)
- [x] `go test ./...` — all packages green
- [x] `gofmt -l .` — clean
- [x] New coverage: `cli/tests/test_llm_env.py` (parity gate + env injection), `cli/tests/test_gateway.py::ResolveGatewayBinaryTests` (three-tier lookup), `cli/tests/test_cmd_doctor.py::VerifyBedrockTests` + `BedrockRoutingTests`, `internal/tui/setup_guardrail_inherit_test.go`
- [ ] Manual smoke on macOS: `make clean && make all` → `defenseclaw doctor`
- [ ] Manual smoke on Linux CI: `install.sh --quickstart` in a fresh container
- [ ] `defenseclaw uninstall --dry-run` from an unrelated dir (verify refusal)
- [ ] `defenseclaw keys fill-missing` with partial `.env` (verify hidden input + chmod 0o600 preserved)
- [ ] `defenseclaw tui` in a fresh shell immediately after `make all` (verify canonical-path fallback)

## Rollout

No data migration required. Existing `~/.defenseclaw/config.yaml` files continue to load; new features are opt-in and gated by config. First `doctor` run after upgrading populates the credential registry surface automatically. The unified `llm:` block is additive — existing per-scanner `api_key`/`api_key_env` values still win.

## Deferred (future work)

- TUI-native key-entry wizard — the CLI `keys fill-missing` flow remains canonical. Revisit when telemetry shows user demand.
- OS-keychain secret storage — attractive on macOS but falls back to plaintext files on Linux CI, so not a clear win over `.env` with `0o600`.


cli/defenseclaw/commands/cmd_init.py
```diff
@@ -348,9 +348,23 @@ def _ensure_device_key(path: str) -> None:
...
-        os.chmod(path, 0o600)
+    # Create the file with 0o600 atomically so the key is never
+    # world-readable, even for the brief window between open() and
+    # the previous chmod(). ``O_EXCL`` ensures we don't overwrite a
+    # concurrently-created key (idempotent early-exit already covered
+    # the is-it-there case above).
+    flags = os.O_WRONLY | os.O_CREAT | os.O_EXCL
+    try:
+        fd = os.open(path, flags, 0o600)
+    except FileExistsError:
```
ihabler: Claude pointed on the following security issue:
The O_EXCL fix is correct — closes the umask window. But the FileExistsError branch then trusts whatever is currently at path. Combined with the symlink semantics of O_EXCL (refuses to follow), this is mostly safe, but an attacker who can create files in ~/.defenseclaw/ could pre-stage device.key and we'd silently use their key.

Defense-in-depth: after the early-exit, verify the existing file is mode 0o600, owned by the current user, and at least the right size for an Ed25519 PEM. If anything looks off, fail loudly rather than silently trust.

cli/defenseclaw/commands/cmd_setup.py
```diff
@@ -277,11 +277,24 @@ def _load_dotenv(path: str) -> dict[str, str]:
...
     lines = [f"{k}={v}\n" for k, v in sorted(entries.items())]
     fd = os.open(path, os.O_WRONLY | os.O_CREAT | os.O_TRUNC, 0o600)
     with os.fdopen(fd, "w") as f:
         f.writelines(lines)
+    try:
+        os.chmod(path, 0o600)
+    except OSError:
+        # Best-effort: on some filesystems chmod is a no-op. We've
+        # already written the data, so don't fail the caller here.
+        pass
```
ihabler: recommend to change into:
    try:
        os.fchmod(fd, 0o600)   
    except OSError:
        pass
    with os.fdopen(fd, "w") as f:
        f.writelines(lines)

to use  same inode as the open as openning and chmod op

cli/defenseclaw/commands/cmd_uninstall.py
```diff
@@ -0,0 +1,299 @@
...
+        return
+    try:
+        shutil.rmtree(data_dir)
+        click.echo(f"  ✓ removed {data_dir}")
+    except OSError as exc:
+        click.echo(f"  ⚠ failed to remove {data_dir}: {exc}")
+
+
+def _remove_binaries() -> None:
+    targets = [
```
ihabler: Nits
_remove_binaries doesn't verify the binary it's removing
if ~/.local/bin/defenseclaw is a symlink to a Homebrew install or a system package manager, we silently unlink the symlink. Probably fine (os.unlink only removes the link), but worth checking os.readlink if it's a symlink and warning if the target isn't inside our install tree.

vineethsai7 (dismissed):

PR #130, opened 2026-04-21
Update README.md

truedays:
Improved legibility of ASCII logo


nghodkicisco (changes requested):
what is the change required here?

nghodkicisco:
The change is not required , will come with updated README

PR #141, opened 2026-04-24
Feature/connector architecture v3

nghodkicisco:
# Connector Architecture v3 — Executive Summary

**Branch**: `feature/connector-architecture-v3`
**Date**: 2026-04-24

---

## Problem

DefenseClaw's guardrail proxy was hardcoded to work only with OpenClaw. Every authentication check, request signal extraction, hook script, and telemetry label assumed a single agent framework. Supporting additional frameworks (Claude Code, Codex, ZeptoClaw, Cursor, OpenCode) required duplicating proxy logic for each one, creating a maintenance and security burden that would grow linearly with each new integration.

Additionally, a code review of the existing connector package uncovered 18 findings — including two critical bugs that could silently break hook teardown and registration — that needed to be addressed before the architecture could be trusted in production.

---

### 1. Pluggable Connector Architecture

DefenseClaw now has a **Connector interface** — a contract that any agent framework adapter implements. Each connector owns five responsibilities for its agent: authentication, LLM traffic routing, tool call inspection mode, subprocess enforcement policy, and setup/teardown lifecycle.

Four built-in connectors ship with this release:
- **OpenClaw** — fetch interceptor plugin (existing, now behind the interface)
- **ZeptoClaw** — `api_base` redirect with `before_tool` hook
- **Claude Code** — environment variable override with PreToolUse/PostToolUse hook scripts
- **Codex** — environment variable override with hook scripts and response scanning

A **connector registry** manages built-in and external connectors. External connectors can be loaded as Go plugin `.so` files from a configurable directory, enabling third-party or enterprise-specific integrations without modifying DefenseClaw core.

### 2. Hook Event Handlers

Two new gateway-level hook handlers process agent lifecycle events:

- **Claude Code hook handler** — Receives events from Claude Code's hook system (PreToolUse, PostToolUse, UserPromptSubmit, SessionStart, Stop, etc.) and dispatches them through the inspection pipeline. In `mode=action`, dangerous tool calls and prompt injections are blocked before execution. In `mode=observe`, findings are logged without blocking.

- **Codex hook handler** — Same capability adapted for the Codex event model.

Both handlers support component scanning on session start, changed-file scanning on session stop, and structured responses that tell the agent whether to proceed or abort.

### 3. Inspection Pipeline Expansion

Three new inspection endpoints complement the existing `/api/v1/inspect/tool`:

- **Pre-request scan** (`/inspect/request`) — Scans user queries for prompt injection and data exfiltration patterns before they reach the LLM
[...]

PR #165, opened 2026-04-27
fix(hooks): fail-open when DefenseClaw is disabled or uninstalled (S7.1 / clean uninstall)

vineethsai:
## Summary

Today every generated hook script defaults to `FAIL_MODE=closed` and dials the gateway unconditionally. That is the right policy while DefenseClaw is healthy, but it silently bricks the agent the moment the operator runs `defenseclaw setup guardrail --disable`, runs `defenseclaw uninstall`, or just `rm -rf ~/.defenseclaw`. The hook keeps firing, curl errors out (the gateway is gone), the script exits 2, and Claude Code / Codex / OpenClaw / ZeptoClaw treats every tool call as denied. This is the \"can't actually turn the guardrail off\" behavior tracked under the *clean uninstall* requirement.

This PR bumps every hook template to `v2` and adds a fail-open guard at the very top of each script:

```bash
DEFENSECLAW_HOME=\"\${DEFENSECLAW_HOME:-\${HOME}/.defenseclaw}\"
if [ ! -d \"\${DEFENSECLAW_HOME}\" ] || [ -f \"\${DEFENSECLAW_HOME}/.disabled\" ]; then
  exit 0
fi
```

For the two token-bearing hooks (`claude-code-hook.sh`, `codex-hook.sh`) we additionally short-circuit when the `.token` sidecar is gone **and** no `DEFENSECLAW_GATEWAY_TOKEN` env override is set — otherwise a stale hook script left behind by an aborted uninstall turns into a permanent 401-driven block.

## Why version-agnostic marker

`hookMarker` in `internal/gateway/connector/claudecode.go` is now `\"# defenseclaw-managed-hook v\"` (prefix match) instead of `\"# defenseclaw-managed-hook v1\"`. Without this, `Teardown` would no longer recognise older hooks generated by previous installs, orphaning them in agent configs and silently re-bricking the agent the next time it ran. The trailing version digit is deliberately not part of the match.

## What this PR does **not** do

- It does *not* implement the Python plumbing for `defenseclaw setup guardrail --enable/--disable` (that's S7.4, in a follow-up PR — this PR is the runtime contract those subcommands target).
- It does *not* change `FAIL_MODE` semantics. When the guardrail is enabled and reachable, the existing fail-closed behavior is preserved bit-for-bit.

## Test plan

New unit tests in `internal/gateway/connector/connector_test.go`:

- [x] `TestHookScripts_FailOpen_OnDisabledSentinel` — all 6 hooks (`inspect-tool`, `inspect-request`, `inspect-response`, `inspect-tool-response`, `claude-code-hook`, `codex-hook`) exit 0 and never dial the stub curl when `~/.defenseclaw/.disabled` exists.
- [x] `TestHookScripts_FailOpen_OnMissingDefenseClawHome` — same, for the `rm -rf ~/.defenseclaw` uninstall case.
- [x] `TestHookScripts_TokenedHooks_FailOpen_OnMissingToken` — `claude-code-hook` and `codex-hook` short-circuit when `.token` is missing and `DEFENSECLAW_GATEWAY_TOKEN` is unset.

```
$ go test -count=1 ./internal/gateway/connector/...
ok      github.com/defenseclaw/defenseclaw/internal/gateway/connector  1.014s
```

All pre-existing connector tests (`TestWriteHookScript*`, `TestWriteHookScriptsWithToken_*`, `TestHookScripts_ReturnsList`) continue to pass — the new sentinel runs *before* the curl path so the existing happy-path assertions are unchanged.

## Stack

Stacked on top of \`feature/connector-architecture-v3\` (PR #141).

Refs: claw-agnostic-refactor S7.1.


vineethsai:
Superseded by #194 (release rollup PR for the entire connector-architecture-v3 wave, targets `main`). Each step is preserved as a discrete merge commit in the release branch. Closing to keep the PR list clean.

PR #190, opened 2026-04-28
S7.3: defenseclaw uninstall polymorphic per-connector teardown

vineethsai:
## Summary

Route ``defenseclaw uninstall`` through the S7.2 ``defenseclaw-gateway connector teardown`` sentinel so each connector adapter owns its own rollback path. Today the Python flow always calls ``restore_openclaw_config()`` regardless of which framework is active — which would silently corrupt a Codex / Claude Code / ZeptoClaw install.

## What changed

| Step | Before                                                                   | After                                                                                 |
| ---- | ------------------------------------------------------------------------ | ------------------------------------------------------------------------------------- |
| Plan | Always ``revert openclaw.json`` + ``remove plugin``                       | Resolves ``Config.active_connector()`` and surfaces it in the plan                    |
| Render | Single OpenClaw-shaped line                                            | Per-connector wording: ``revert openclaw.json`` only for openclaw; ``teardown <name>`` for everything else |
| Execute | Direct calls to OpenClaw Python helpers                                | Sentinel-first: ``defenseclaw-gateway connector teardown --connector <name>`` (S7.2). Fallback to Python helpers **only** for OpenClaw on legacy gateways |
| Plugin | Always called ``_remove_plugin``                                       | Only called for OpenClaw. Other connectors had their hooks/shims removed by the sentinel |

## Safety guarantees

- Non-OpenClaw connectors **never** silently call ``restore_openclaw_config``. If the gateway is too old to expose the sentinel, the flow logs an explicit remediation (\"upgrade gateway to v0.7+\") and aborts the destructive step.
- Sentinel availability is detected with a non-destructive ``connector --help`` probe checking for ``teardown`` and ``list-backups`` in the output; the probe is bounded by a 10s timeout.

## Test plan

- [x] ``pytest tests/test_cmd_uninstall.py`` (25 cases, +19 new)
- [x] Build plan resolves connector from ``Config.active_connector``, ``guardrail.connector``, falls back to ``openclaw``
- [x] Render output is connector-specific (``codex`` shows ``teardown codex``, ``openclaw`` shows ``revert openclaw.json``)
- [x] Sentinel dispatch covers: modern gateway, missing gateway, gateway help missing the subcommand, gateway help exiting non-zero
- [x] Execute plan: codex skips ``_remove_plugin``; openclaw still calls it
- [x] Codex hard-fails (no Python fallback) when sentinel unavailable

Stacked on top of #189 (S7.2).

PR #189, opened 2026-04-28
S4 + S6 + S7: connector polymorphism for CLI, scan UX, and lifecycle (rollup of #178-#189)

vineethsai:
## Summary

Single rollup PR consolidating the entire S4 + S6 + S7 stack — 15 discrete commits, each a reviewable step. Replaces the 12-PR linear stack #178 → #179 → #180 → #181 → #182 → #183 → #184 → #185 → #186 → #187 → #188 → #189 with one PR.

## Why one PR

The native [GitHub Stacked PRs](https://github.github.com/gh-stack/introduction/overview/) feature is in private preview and isn't enabled for this repo (`gh stack link` returns *Stacked PRs are not enabled for this repository*). To stop polluting the PR list, the entire 12-PR chain is folded into this single PR. Each commit stays individually reviewable in the diff view, organized in logical groups below.

## Wave S4 — Python CLI polymorphism

| Commit | Step | What it does |
|--------|------|--------------|
| `76e5a00` | **S4.1** | Extract connector path discovery into `connector_paths.py`. `cli/defenseclaw/config.py` now exposes polymorphic `skill_dirs()` / `plugin_dirs()` / `mcp_servers()` that route through the active connector. |
| `542df85` | **S4.2** | `defenseclaw mcp set` / `unset` become connector-aware: writes to `~/.codex/config.toml` for codex, `~/.claude/settings.json` for claudecode, `~/.openclaw/settings.json` for openclaw, refuses for zeptoclaw with a helpful error. |
| `78b8a42` | **S4.3 / F4** | `claw_inventory.py` filesystem adapter — AIBOM is generated by walking the active connector's directories instead of hard-coded OpenClaw paths. |
| `80d1b7e` | **S4.4** | Split `guardrail.py` into `llm_keys.py` (cross-connector) + `openclaw_guardrail.py` (OpenClaw-specific). Adds connector-aware `skill_list` adapter. |
| `19382f2` | **S4.5 / F23** | `cmd_setup_sandbox.py` gates sandbox setup on connector capability — refuses cleanly for connectors that don't ship a sandbox. |

## Wave S6 — scan UX overhaul

| Commit | Step | What it does |
|--------|------|--------------|
| `f367045` | **S6.1** | New `cli/defenseclaw/commands/_scan_ui.py` — `ScanContext` + helpers for preamble / per-target glyph / verdict / `--json` payload, all in one place. |
[...]

PR #191, opened 2026-04-28
S7.4: defenseclaw guardrail {enable,disable,status} connector-agnostic

vineethsai:
## Summary

Add a top-level \`defenseclaw guardrail\` command group with three quick toggle subcommands so day-to-day operators have a fast on/off switch instead of re-running the full \`defenseclaw setup guardrail\` flow every time.

\`\`\`
defenseclaw guardrail status      # is it on, which connector, which mode
defenseclaw guardrail enable      # turn on  + connector setup
defenseclaw guardrail disable     # turn off + connector teardown
\`\`\`

## Connector polymorphism (S7 wave)

Both \`enable\` and \`disable\` resolve the active connector from \`Config.active_connector()\` and propagate the resolved name into \`_restart_services(connector=...)\`. The Go sidecar's boot path runs \`Connector.Setup\` / \`Connector.Teardown\` for the right adapter — so this CLI never has to know how Codex / Claude Code / ZeptoClaw configure themselves.

## Safety guarantees

- \`disable\` aborts before restarting if config save fails — otherwise the sidecar would tear down a connector the operator never actually disabled.
- \`enable\` refuses to flip the flag when \`guardrail.model\` and \`llm.model\` are both empty — re-enabling without a configured upstream would silently route real traffic through nothing.
- \`--no-restart\` is supported with explicit messaging that the setup/teardown won't run until the operator restarts the gateway manually.
- Confirmation prompt + \`--yes\` flag for non-interactive flows.

## Test plan

- [x] \`pytest cli/tests/test_cmd_guardrail.py\` — 16 cases
- [x] Active-connector resolution: method → guardrail.connector → \"openclaw\"
- [x] Status output for enabled vs disabled, openclaw vs codex
- [x] Disable for codex / claudecode / zeptoclaw / openclaw, including save-failure abort and decline-confirmation abort
- [x] \`--no-restart\` flag does NOT call \`_restart_services\`
- [x] Restart call propagates the resolved connector (the gateway must teardown the right adapter)
- [x] Enable refuses when both \`guardrail.model\` and \`llm.model\` are empty
- [x] Enable accepts \`llm.model\` as fallback when \`guardrail.model\` is empty
- [x] CLI registration smoke test confirms exactly {enable, disable, status} are exposed

Stacked on top of #190 (S7.3).

### DIFF
--- a/cli/defenseclaw/commands/cmd_guardrail.py
+++ b/cli/defenseclaw/commands/cmd_guardrail.py
@@ -1,0 +1,247 @@
+# Copyright 2026 Cisco Systems, Inc. and its affiliates
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+"""``defenseclaw guardrail {enable,disable}`` — connector-agnostic toggle.
+
+Today operators have to use ``defenseclaw setup guardrail [--disable]``,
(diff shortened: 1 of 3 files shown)


vineethsai:
Superseded by #189, which rolls S7.2 + S7.3 + S7.4 + S7.5 into a single PR (each step still reviewable as a discrete commit). Closing to keep the PR list clean.

Issue #150, opened 2026-04-24
guardrail proxy ignores api_base, routes to api.openai.com instead of configured endpoint


kumar-aamit:
Bug: guardrail proxy ignores configured api_base

Description

The guardrail proxy currently uses NewProvider() which ignores the user-configured api_base URL. 
This breaks support for custom or self-hosted LLM endpoints (e.g., Nemotron) and leads to misleading authentication errors.

Example Configuration
guardrail:
  api_base: https://my-custom-gateway.example.com/v1

Steps to Reproduce

- Configure a custom api_base in ~/.defenseclaw/config.yaml
- Start the guardrail proxy
- Send a request through the proxy

Actual Behavior

Requests are routed to:
https://api.openai.com/v1

Resulting in errors such as:
{"error":{"message":"invalid API key","type":"authentication_error","code":"invalid_api_key"}}
Expected Behavior

Requests should be routed to the configured api_base:
https://my-custom-gateway.example.com/v1

Root Cause

In internal/gateway/proxy.go, the provider is instantiated with:
**provider, err := NewProvider(cfgModel, apiKey)**

This ignores the configured api_base.

Proposed Fix

Replace with:
**provider, err := NewProviderWithBase(cfgModel, apiKey, p.cfg.APIBase)**

Impact
Severity: Critical
Effect: Guardrail proxy is non-functional when using a custom api_base
Scope: All versions using the built-in guardrail proxy

highopes:
Hi @kumar-aamit, I ran into a closely related issue while testing the guardrail fetch-interceptor path with a self-hosted OpenAI-compatible vLLM backend.

I opened #197 to make the proxy honor `X-DC-Target-URL` as the upstream base URL for fetch-intercepted providers, so those requests do not fall back to the public OpenAI endpoint.

I also opened #200 as an umbrella issue for self-hosted OpenAI-compatible provider support in the guardrail fetch-interceptor path.

I am not sure whether #197 fully covers the `guardrail.api_base` path described here, or whether an additional fix is needed for the configured-provider fallback path. I wanted to link the work here because the failure mode is very similar.

heungwing:
I got same issue, My provider is vllm on openclaw, but the defenseclaw redirect to openai.com as fallback

kumar-aamit:
@highopes @heungwing 
I have submitted a fix for this issue — PR #201:

Changes:
- proxy.go: NewProvider() → NewProviderWithBase() to honor api_base
[...]

PR #197, opened 2026-04-29
fix(guardrail): honor target base URL for intercepted providers

highopes:
## Summary

This PR makes the guardrail proxy honor `X-DC-Target-URL` as the custom upstream base URL for fetch-intercepted providers.

The change keeps the existing upstream handling for Azure and Bedrock, while extending the same base URL preservation behavior to self-hosted OpenAI-compatible providers such as vLLM.

## Motivation

The fetch interceptor forwards the original upstream origin via `X-DC-Target-URL`.

For self-hosted OpenAI-compatible providers, the guardrail proxy must preserve that origin when constructing the upstream provider. Without this, intercepted requests can fall back to a provider's public default endpoint, such as `https://api.openai.com`, and use the self-hosted provider token as an OpenAI API key.

## Changes

- Use `req.TargetURL` as the default provider base URL for fetch-intercepted provider resolution.
- Preserve the existing Bedrock-specific base URL normalization.
- Keep the change intentionally small and focused on provider base URL routing.

## Relationship to related issues and PRs

- Part of #200.
- Complementary to #201.
- #201 covers the direct-provider/config fallback path for #150.
- This PR covers the fetch-interceptor path where `X-DC-Target-URL` is present.

## Validation

- Rebased onto the latest upstream `main`.
- PR is mergeable after the rebase.

### DIFF
--- a/internal/gateway/proxy.go
+++ b/internal/gateway/proxy.go
@@ -1592,14 +1592,16 @@ func (p *GuardrailProxy) resolveProviderFromHeaders(req *ChatRequest) LLMProvide
 		return nil
 	}
 
+	// Fetch-intercepted requests carry the original upstream origin in
+	// X-DC-Target-URL. Use it as the provider base URL so self-hosted
+	// OpenAI-compatible endpoints, such as vLLM, are not routed to the
+	// provider's public default endpoint.
+	//
 	// Azure requires the specific resource endpoint as baseURL.
 	// Bedrock (ZeptoClaw / OpenClaw) uses regional bedrock-runtime URLs with
-	// ABSK bearer keys — Bifrost handles that path; pin BaseURL to the snapshot
-	// origin so the tenant routes to the correct region.
-	baseURL := ""
-	if prefix == "azure" {
-		baseURL = req.TargetURL
-	}
+	// ABSK bearer keys; keep the baseURL pinned to the snapshot origin so the
+	// tenant routes to the correct region.
+	baseURL := req.TargetURL
 	if prefix == "bedrock" {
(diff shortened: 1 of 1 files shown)


highopes:
Part of #200.

This PR addresses the first compatibility gap described in the issue: fetch-intercepted OpenAI-compatible requests should honor `X-DC-Target-URL` as the upstream provider base URL, instead of falling back to a public default endpoint.

highopes:
Refreshed this PR against the latest upstream main and kept the change intentionally minimal.

The current version preserves the existing Azure and Bedrock handling, while making X-DC-Target-URL the default provider base URL for fetch-intercepted providers. This keeps the fix focused on the fetch-interceptor path.

This is complementary to #201: #201 covers the direct-provider/config fallback path for #150, while this PR covers the fetch-interceptor path tracked in #200.

vineethsai:
Closing as superseded.

#284 (the 0.7.0 release) reworks the guardrail proxy so the upstream base URL is resolved from `X-DC-Target-URL` (and the configured `llm.base_url`) as a first-class path, rather than falling back to a provider's public default endpoint.

See `internal/gateway/proxy.go`, which resolves the upstream origin from `X-DC-Target-URL` and fails explicitly when neither the header nor a configured base is present:

> `missing X-DC-Target-URL header and no llm.base_url configured`

The self-hosted / OpenAI-compatible case this PR fixed — preserving the intercepted origin so an intercepted request never falls back to `https://api.openai.com` and leaks the self-hosted token as an OpenAI key — is handled there.

Reopen if you spot a gap in the new resolution path.

PR #198, opened 2026-04-29
fix(guardrail): preserve raw OpenAI-compatible request bodies

highopes:
## Summary

This PR adds a raw-forward path for fetch-intercepted OpenAI-compatible Chat Completions requests.

The goal is to preserve provider-specific request fields while keeping DefenseClaw PRE-CALL and POST-CALL guardrail inspection active.

## Motivation

Some OpenAI-compatible providers require extension fields that are not part of the core OpenAI Chat Completions schema. Examples include `chat_template_kwargs`, `extra_body`, `parallel_tool_calls`, and provider-specific sampling or model-control options.

When these requests are translated through a structured provider abstraction, extension fields can be dropped. For self-hosted providers such as vLLM, this can change model behavior and break compatibility.

## Changes

- Add a raw-forward path for fetch-intercepted requests with `X-DC-Target-URL`.
- Preserve the original OpenAI-compatible request body when forwarding upstream.
- Keep PRE-CALL prompt inspection before forwarding.
- Extract assistant-visible text from non-streaming and streaming responses.
- Keep POST-CALL completion inspection on assistant output only.
- Avoid scanning request-side system prompts, tool schemas, or agent context as completion text.
- Add unit tests for raw-forward completion extraction.

## Validation

- `go test ./internal/gateway -run 'TestExtractRawForwardCompletion' -v`: passed
- `go test ./internal/gateway`: passed
- `go build ./cmd/defenseclaw`: passed
- `make test`: passed
- `make lint`: failed on existing upstream lint findings unrelated to this PR. I reproduced the same lint failures on `upstream/main`.

Manual validation with a self-hosted vLLM OpenAI-compatible endpoint:

- Direct proxy request preserved `chat_template_kwargs.enable_thinking=false`.
- Normal completion returned assistant-visible content only.
- Prompt injection test was blocked by PRE-CALL guardrail.

## Notes

This PR intentionally does not add provider-specific request overrides. That is left for a separate PR so this change remains focused on raw request-body preservation and response extraction.


highopes:
Part of #200.

This PR addresses the second compatibility gap described in the issue: fetch-intercepted OpenAI-compatible requests need a raw body preserving path so provider-specific fields, such as `chat_template_kwargs`, `extra_body`, and similar extension fields, are not dropped by the structured translation path.

highopes:
After re-checking the current guardrail architecture, I want to clarify the architectural trade-off in this PR.

DefenseClaw currently uses Bifrost SDK as the unified provider routing layer. This PR introduces a raw-forward fallback path only for fetch-intercepted OpenAI-compatible requests with X-DC-Target-URL, because some self-hosted providers such as vLLM require provider-specific request body fields like chat_template_kwargs / extra_body that may be lost during structured provider translation.

This path does not bypass DefenseClaw PRE-CALL or POST-CALL inspection, but it does bypass Bifrost's structured provider forwarding for this specific compatibility case.

If maintainers prefer to keep all upstream forwarding strictly inside Bifrost, I am happy to rework this as either:
1. a feature-flagged temporary compatibility fallback, or
2. a smaller change that waits for / depends on Bifrost-level support for provider-specific body passthrough.

vineethsai7 (approved):

PR #199, opened 2026-04-29
feat(guardrail): support provider request overrides

highopes:
## Summary

This PR adds configurable request overrides for fetch-intercepted providers.

The main use case is self-hosted OpenAI-compatible providers such as vLLM, where operators may need provider-specific request fields while preserving DefenseClaw PRE-CALL and POST-CALL guardrail inspection.

## Motivation

Some OpenAI-compatible providers require provider-specific fields that are not part of the core Chat Completions schema.

For example, vLLM-hosted Qwen models may require:

```json
{
  "chat_template_kwargs": {
    "enable_thinking": false
  }
}
```

Hard-coding model-specific behavior in the guardrail proxy would not be appropriate. This PR adds a generic `request_overrides` mechanism in `custom-providers.json` so operators can configure these fields per provider.

## Changes

- Extend provider registry entries with optional `request_overrides`.
- Merge `request_overrides` from `~/.defenseclaw/custom-providers.json`.
- Apply provider request overrides before raw-forwarding fetch-intercepted OpenAI-compatible requests upstream.
- Deep-merge nested JSON override fields.
- Add unit tests for overlay merging and request override application.
- Document self-hosted OpenAI-compatible providers and request overrides in `docs/GUARDRAIL.md`.

## Example

```json
{
  "providers": [
    {
      "name": "vllm",
      "domains": ["192.168.160.113"],
      "env_keys": ["VLLM_API_KEY"],
      "request_overrides": {
        "chat_template_kwargs": {
          "enable_thinking": false
        }
      }
    }
  ],
  "ollama_ports": []
}
```

## Validation

- `go test ./internal/configs -run TestApplyOverlayMergesRequestOverrides -v`: passed
- `go test ./internal/gateway -run 'TestApplyProviderRequestOverrides' -v`: passed
- `go test ./internal/configs ./internal/gateway`: passed
- `go build ./cmd/defenseclaw`: passed
- `make test`: passed
- `make lint`: failed on existing upstream lint findings unrelated to this PR. I reproduced the same lint failures on `upstream/main`.

## Dependency

Depends on #198 because the request overrides are applied in the raw-forward path introduced there.


highopes:
Part of #200.

This draft PR addresses the third compatibility gap described in the issue: self-hosted OpenAI-compatible providers may need configurable request overrides, instead of hard-coded model-specific behavior in the proxy.

PR #225, opened 2026-05-06
feat(doctor): improve Splunk HEC diagnostics with HEC reply code parsing

feraszah:
## Problem

`defenseclaw doctor` already probes the Splunk HEC endpoint, but only inspects the HTTP status code. This means operators get vague, non-actionable output like `warn HTTP 400` or `authentication failed  (HTTP 403)` with no indication of what's actually wrong or how to fix it.

Splunk's HEC always returns a JSON body with an internal reply code (e.g. `{"text":"Incorrect index","code":7}`) that precisely identifies the failure. The previous implementation threw this away entirely.

### HEC reply code parsing
Adds `_parse_hec_response()` which decodes the Splunk HEC JSON reply code from the response body and maps it to a human-readable, actionable message based on Splunk's documented reply codes.

Specific improvements:

- **Disabled token (code 1)** → "token is disabled — enable the HEC token in Splunk" instead of generic "authentication failed"
- **Invalid token (code 4)** → "invalid token — verify splunk_hec.token_env points to the correct value" instead of generic "authentication failed"  
- **Incorrect index (code 7)** → "incorrect index (configured index: 'my-index') — create the index in Splunk or update splunk_hec.index" instead of "warn HTTP 400"
- **Server busy (code 9)** → "Splunk HEC server busy — consider reducing flush frequency"
- **Queues full (code 18)** → "Splunk HEC queues are full — indexer may be overloaded"
- **TLS errors** → detected from network failure body text and surfaced as `fail` with guidance to check `verify_tls`, instead of a generic `warn unreachable`

### Token security posture check
Adds `_check_splunk_token_posture()` which warns during doctor when the HEC token is stored inline in `config.yaml` rather than referenced via `token_env`. This surfaces a security posture issue before it becomes a finding, Splunk's own best practices documentation explicitly recommends against storing HEC tokens in configuration files.

Reference: https://help.splunk.com/en/splunk-enterprise/get-started/get-data-in/10.2/get-data-with-http-event-collector/troubleshoot-http-event-collector

## Testing

26 new tests in `cli/tests/test_cmd_doctor_splunk_hec.py`, all passing:

- 8 tests for `_parse_hec_response` covering known HEC codes, unknown codes, non-JSON bodies, and empty responses
- 15 tests for `_probe_splunk_hec` covering all HTTP status code branches, specific HEC reply codes, TLS detection, and missing config
- 3 tests for `_check_splunk_token_posture` covering inline token detection, token_env usage, and load failure resilience

All 32 existing `test_cmd_doctor.py` tests continue to pass.

### DIFF
--- a/cli/defenseclaw/commands/cmd_doctor.py
+++ b/cli/defenseclaw/commands/cmd_doctor.py
@@ -1044,6 +1044,7 @@ def _check_splunk_token_posture(cfg, d, label: str, r: _DoctorResult) -> None:
     security finding.
     """
     import os
+
     from defenseclaw.observability.writer import CONFIG_FILE_NAME, _load_yaml
     try:
         doc = _load_yaml(os.path.join(cfg.data_dir, CONFIG_FILE_NAME))
--- a/cli/defenseclaw/commands/cmd_doctor.py
+++ b/cli/defenseclaw/commands/cmd_doctor.py
@@ -916,15 +916,66 @@ def _probe_otel_destination(cfg, d, r: _DoctorResult) -> None:
         _emit("warn", label, f"{host}:{port} not reachable: {exc}", r=r)
 
 
+# HEC internal reply codes from Splunk docs — used to surface
+# actionable diagnostics instead of the raw HTTP status code.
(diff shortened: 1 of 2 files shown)


feraszah:
Working on fixing the CI lint error

vineethsai7 (approved):

PR #230, opened 2026-05-08
fix(release): drop homebrew + tag-driven version stamping + workflow_dispatch trigger

vineethsai:
## Summary

Three connected fixes that together unbreak the release pipeline and harmonize it with `scripts/install.sh`, `scripts/upgrade.sh`, and `defenseclaw upgrade`.

### 1. Drop the broken Homebrew cask block

GoReleaser has been failing on every tag since `0.3.0` because `homebrew_casks.repository.name: homebrew-tap` points at a repo that does not exist (`cisco-ai-defense/homebrew-tap` returns 404). With `skip_upload: auto`, GoReleaser still resolves the tap to detect its default branch and 404s, aborting the run before assets get uploaded — that's why `0.3.1` ended up as an empty release shell and `0.3.2` has no GitHub release at all.

We're not shipping Homebrew today, so the cask block is removed entirely.

### 2. Tag-driven version stamping (single source of truth)

Before this PR there were **four** independent version sources:
- `Makefile` `VERSION := 0.2.0`
- `extensions/defenseclaw/package.json` `0.2.0`
- `pyproject.toml` `0.5.0`
- `cli/defenseclaw/__init__.py` `__version__ = "0.5.0"`

…all of which fed into release artifact filenames (gateway tarball, wheel, plugin tarball) without coordination. A tag of `0.3.3` would have produced `defenseclaw_0.3.3_*.tar.gz` + `defenseclaw-0.5.0-py3-none-any.whl` + `defenseclaw-plugin-0.2.0.tar.gz` — and `defenseclaw upgrade` would 404 on the wheel.

The release workflow now runs `scripts/stamp-version.sh "$TAG"` before any build step, writing the tag value into all four files atomically. Wheel, plugin tarball, gateway binary, and `defenseclaw --version` all converge on the same value, which is what `install.sh` and `defenseclaw upgrade` look up at GitHub Releases.

In-repo defaults are bumped to `0.4.0` (the planned next release) so local `make dist` builds also produce correctly-named artifacts.

### 3. `workflow_dispatch` trigger so releases can be cut from the Actions UI

You can now go to **Actions → Release → Run workflow**, enter `0.4.0`, and the job will:

1. Validate the input is strict X.Y.Z (no `v` prefix, no pre-release suffix).
2. Refuse to run if the tag already exists locally or on origin.
3. Refuse to run if a release already exists for that version (Immutable Releases trap).
4. Stamp the version into all four files.
5. Build everything (GoReleaser → tarballs/sboms/checksums/sigs, wheel, plugin tarball).
6. **Atomically** create the tag and the GitHub release with all assets attached, via `gh release create --target $GITHUB_SHA`. A build failure leaves neither a tag nor a release behind.

The legacy `git push origin X.Y.Z` path still works — both triggers converge into the same job after the validation step.

### 4. Version-sync CI gate (defense in depth)

`.github/workflows/ci.yml` gets a new `version-sync` job that runs `make check-version-sync`. PRs that drift the four in-repo version sources fail CI before merge, so the next release is always clean.

## Files changed

- `.goreleaser.yaml` — drop the entire `homebrew_casks:` block.
- `.github/workflows/release.yaml` — add `workflow_dispatch`, single tag-resolution step, version stamping, `--target $GITHUB_SHA` for atomic tag+release.
- `.github/workflows/ci.yml` — add `version-sync` job.
- `scripts/stamp-version.sh` — new; updates Makefile / pyproject.toml / `__init__.py` / package.json in lockstep.
- `Makefile` — add `set-version` and `check-version-sync` targets; bump `VERSION` to `0.4.0`.
- `pyproject.toml`, `cli/defenseclaw/__init__.py`, `extensions/defenseclaw/package.json` — bump to `0.4.0`.

## How to release after this PR merges

**Option A (preferred):** Actions → Release → Run workflow → enter `0.4.0` → click Run.

**Option B:** `git tag 0.4.0 && git push origin 0.4.0`.

Either way you'll get a release with the same eight artifact families that `0.2.0` shipped: gateway tarballs (linux/darwin × amd64/arm64), SBOMs for each, `checksums.txt` + `.sig` + `.pem`, the Python wheel, and the OpenClaw plugin tarball.

WARNING: Do NOT click "Draft a new release" in the GitHub UI before triggering the workflow. Immutable Releases will freeze the empty release on creation, the workflow's pre-flight will fail, and you'll have to delete it manually before retrying — that's exactly what stranded `0.3.1` with zero assets.

## Test plan

- [x] Local: `scripts/stamp-version.sh 0.4.0` is idempotent; rejects `0.4.0-rc1`; updates all four files.
- [x] Local: `make set-version VERSION=0.4.0` works; `make set-version` (no arg) errors with usage.
- [x] Local: `make check-version-sync` reports OK with all four files at `0.4.0`.
- [x] CI: workflow YAML parses cleanly with both triggers, `version` input wired to `inputs.version`.
- [ ] After merge: trigger workflow_dispatch with `0.4.0` and confirm the release has the full asset set.

## Follow-up (already done out of band)

- Empty `0.3.1` GitHub release shell deleted; git tag preserved.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,6 +14,19 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  version-sync:
+    name: Version Sync Check
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      # Catches drift between the four in-repo version sources (Makefile,
+      # pyproject.toml, cli/__init__.py, package.json) before it can ship
+      # in a release. The release workflow restamps from the git tag, but
+      # PRs still need to keep the in-repo defaults in sync so local dev
+      # builds (`make dist`) produce artifacts whose names match what
+      # install.sh / `defenseclaw upgrade` will look up at the next tag.
+      - run: make check-version-sync
+
   go-lint:
     name: Go Lint
     runs-on: ubuntu-latest
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -18,7 +18,7 @@ name: Release
 # cli/defenseclaw/__init__.py, and extensions/defenseclaw/package.json so
 # every published artifact (gateway tarball, wheel, plugin tarball) is
 # named with the same X.Y.Z that scripts/install.sh and `defenseclaw
-# upgrade` will look up at GitHub Releases.
+# upgrade` look up at GitHub Releases.
 
 on:
   push:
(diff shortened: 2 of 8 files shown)


vineethsai:
Superseded by #231, which keeps the three release-workflow commits from this branch (`139dfc8`, `2895e6c`, `18874bc`) and adds the registries + OS-notifications work. Closing in favour of that PR.

PR #256, opened 2026-05-10
security: close DeepSec audit (75 findings → 0) [Do not merge yet]

vineethsai:
## Summary

Closes every actionable finding from the DeepSec security audit
(`.deepsec/data/defenseclaw/reports/report.md`): **1 CRITICAL, 26 HIGH,
24 MEDIUM, 9 HIGH_BUG, 15 BUG = 75 total**. The two reviewer-tagged
"FP" items are also remediated because their underlying code paths
needed changes anyway. **No protections were removed**; the audit
strictly tightens existing guards.

| Severity   | Count | Status |
|---|---|---|
| CRITICAL   |  1 | Fixed |
| HIGH       | 26 | Fixed |
| MEDIUM     | 24 | Fixed |
| HIGH_BUG   |  9 | Fixed |
| BUG        | 15 | Fixed |
| **Total**  | **75** | **All resolved** |

### CRITICAL
- Codex git-scan paths sanitized through `internal/gitsafe`; hostile
  repo config + external diff + `core.fsmonitor` injection vectors
  closed.

### HIGH
- `.github/workflows/{ci,docs-site,e2e,release}.yml`: pin every
  third-party action by SHA, drop unused `permissions`, quote /
  sanitize input expansions.
- `internal/gateway/proxy.go`: SSRF refusal across known/unknown
  branches with `ResolveAndPin`, exact provider matching, IP/CIDR-pinned
  dialer, userinfo / control-byte rejection, structured URL scrubbing
  on every log + metric surface.
- Connector subprocess + plugin loader: token-aware hook templates,
  plugin-loader TOCTOU closed (open-then-stat pinned to same fd),
  embedded plugin canonical token derivation.
- Scanners (mcp/plugin/skill) fail-closed on non-zero subprocess exit;
  `policy.ScanResultInput` extended for the full verdict shape.
- Watcher routes rescan + baseline through admission and installs
  recursive watches on admitted dirs.

### HIGH_BUG
- Scoped sandbox cleanup (no destructive global path); surgical 0.3
  migration; `http_jsonl` synchronous retry; watchdog flock + binary
  fingerprint; `killStaleProcesses` exe+env verification;
  `RepairPairing` fail-closed + atomic.

### MEDIUM
- Webhook URLs hashed/scrubbed on every log path; AI-discovery path
  fingerprints HMAC-keyed; MCP scan target URL validation; endpoint
  ignore-list switched from prefix matching to exact host / loopback
  IP literals.
[...]

PR #258, opened 2026-05-10
security: close all Critical/High findings from pre-launch security audits (consolidated)

vineethsai:
## Summary

Pre-launch security hardening pass that closes every Critical and High
issue found by two independent security audits, plus the bug-class
issues that materially weaken the guardrail or audit surfaces. This is
the single PR for manual acceptance testing — it supersedes #256.

| Severity        | Count | Status |
|-----------------|------:|--------|
| Critical        | 8+    | Fixed  |
| High            | 60+   | Fixed  |
| Medium          | 30+   | Fixed  |
| Bug / Bug-High  | 24    | Fixed  |

No protections were removed; every change tightens an existing guard or
flips a default that was silently fail-open.

---

## What was actually broken (and how)

Every bullet below describes a real bypass or wrong-default that
shipped in `main`. Each one is now closed in this branch with a
regression test pinning the new behavior.

### 1. Guardrails were fail-open by default

* **`hook_fail_mode` defaulted to `open`.** When the gateway returned
  4xx, malformed JSON, or a missing `action` field, the shell hooks
  silently permitted the tool/prompt to execute. Flipped to `closed` for
  fresh installs; legacy installs explicitly seed `open` via migration
  so upgrades don't surprise existing operators.
* **`inspectTool` (TypeScript plugin) returned `action=allow` on every
  sidecar failure** — non-2xx, malformed JSON, `fetch` reject. A
  compromised or partitioned sidecar effectively disabled tool
  inspection. Now fail-closed; legacy opt-out via
  `DEFENSECLAW_TOOL_INSPECT_FAIL_OPEN=1`.
* **Inspect hooks called the gateway without an `Authorization`
  header.** With a configured gateway token the sidecar returned 401;
  the Bash `.action // "allow"` default treated the auth failure as a
  permit decision. Combined with the request hook also being unauthed,
  a malicious agent had a complete unauthenticated path past every
  documented inspection layer. Hooks + PATH shims (curl, wget, nc, ssh,
  pip, npm) now load `${HOOK_DIR}/.token`, send `Authorization: Bearer
  ...`, and treat 401/403 as fail-closed (overrides `FAIL_MODE`).
* **`handleApprovalRequest` auto-approved empty-context approvals.**
  A frame with no `SystemRunPlan`, nested request, raw command, or argv
  went through the `autoApprove` branch and emitted `allow-once` without
  inspecting any command, so a compromised connector could shape an
  empty approval and skip every dangerous-pattern check.
* **`resolveOpenClawInspectConfirm` downgraded HILT confirm → alert**
  when the connector lacked a native approval surface. Hooks that only
  block on `action==block` therefore forwarded dangerous tool calls as
  audit-only telemetry. Now upgrades to `block` with
  `would_block=true`.
* **Substring-based provenance matching.** `_matches_provenance` ran
  `constraint in normalised_path` against the full path string, so an
  attacker-controlled `.defenseclaw-evil/` parent directory satisfied
  the `.defenseclaw` first-party constraint. Now requires constraints
  to align as a contiguous run of full path components.
* **`evaluateGuardrailPolicy` silently allowed on policy outage.**
  Configured bundle's constructor or evaluation could fail and the
  gateway converted the outage into a silent allow. Now returns
  fail-closed (block in action mode, alert in observe mode).
* **`policy validate` passed when `opa` was absent**; `policy
  activate` printed success despite malformed `data.json`. Both now
  surface non-zero exits.

### 2. SSRF / DNS-rebinding chain

The previous code had layered SSRF guards in three places that didn't
agree with each other. The whole stack is now wired to a single set of
primitives in `internal/gateway/provider.go` (`isUnsafeIP`,
`ssrfSafeDialContext`, `ssrfSafeCheckRedirect`).

* **IPv6 zone literals bypassed `isPrivateHost`.** `::1%lo0` and
  `fe80::1%lo0` are valid dial targets, but `net.ParseIP` returns nil
  on the zone-qualified form. The function fell through to DNS
  resolution (which fails for IP literals), returned `false`, and let
  shape-passthrough forward LLM-shaped requests to loopback /
  link-local. Fixed by stripping the zone before `net.ParseIP`.
* **DNS rebinding between preflight `LookupHost` and dial.** Multi-A
  rebinding (`203.0.113.10, 127.0.0.1`) bypassed the IP-literal guard.
  Dialer now re-resolves and rejects on the first private hit, then
  pins the dial to the chosen public IP.
* **Webhook dispatcher followed Go's default redirect policy.** A
  compromised webhook endpoint could 302 the dispatcher to
  `http://127.0.0.1:6379/` (Redis) or `http://169.254.169.254/` (cloud
  IMDS) and Go would follow it. `CheckRedirect` now re-runs the SSRF
  validator on every hop, caps at 5, and rejects non-http(s) schemes.
* **Webhook config-time validator and dispatch-time dialer used
  different deny-lists.** `defenseclaw setup webhook add` accepted URLs
  that the dialer would later reject with an opaque "private host"
  error. Config-time validator now delegates to `isUnsafeIP`.
* **`handleChatCompletion` ran SSRF guards before connector signal
  hydration.** Native-binary connectors (codex, zeptoclaw) have no
  fetch interceptor, so `X-DC-Target-URL` was empty on entry — the
  guards saw `""` and skipped, then `hydrateConnectorSignals` populated
  `TargetURL` with the connector snapshot's upstream. Reordered to
  match `handlePassthrough`.
* **`isAlreadyProxied` substring-matched the proxy hostport** against
  the whole URL string, so a normal provider URL like
  `https://api.openai.com/v1/chat?proxy=127.0.0.1:43099` spoofed proxy
  detection. Replaced with parsed-URL hostname/port comparison.
* **Plain `http.request` calls bypassed every interceptor rail.**
  Only `globalThis.fetch` and `https.request` were monkey-patched, so
  local Ollama clients hitting `http://127.0.0.1:11434/api/chat`
  escaped classification. `http.request` is now patched too.
* **`buildUrlStringFromArgs` ignored Node's options-bag overrides.**
  A caller could pass a benign URL alongside `{ hostname:
  "api.openai.com", path: "/v1/messages" }` and the classifier only
  saw the benign URL. Overlay is now folded into the URL up front so
  every rail sees the actual post-merge target.
* **Bedrock allowlist entry `bedrock-runtime.` was a raw prefix.**
  `matchProviderDomain` treated any "ends-with-dot" pattern as a
  hostname prefix, so an authenticated caller could set
  `X-DC-Target-URL` to `https://bedrock-runtime.attacker.example` and
  have it admitted as a known provider. Schema now uses a label-pinned
  wildcard (`bedrock-runtime.*.amazonaws.com`); the legacy
  bare-prefix syntax is rejected.
* **Passthrough completion extractor missed Ollama native and Bedrock
  Converse shapes.** Inspector short-circuited on empty text → response
  forwarded to client unscanned. Same gap in `extractSSEChunkText` for
  OpenAI Responses, Bedrock Converse, and Ollama streams. Fixed in
  both branches.
* **Tool-call inspection only ran on OpenAI Chat completions.**
  Anthropic `tool_use`, Gemini `functionCall`, Bedrock `toolUse`, and
  OpenAI Responses `function_call` outputs were forwarded
  uninspected, so a prompt-injection that emitted a native tool call
  executed without the configured tool-call rules.
* **`isUnsafeIP` missed CGNAT, ECS task metadata, IPv6 ULA.** Now
  blocks `100.64.0.0/10`, `169.254.170.2/32`, and `fd00::/8`. Tailscale
  / Ollama-on-NAS operators can opt CGNAT back in with
  `DEFENSECLAW_ALLOW_CGNAT=1` (loopback / IMDS / ECS / ULA stay
  blocked unconditionally).
* **`http.get`, Bedrock InvokeModel paths, and query-string secrets in
  interceptor logs.** Three remaining interceptor gaps: (a) `http.get`
  was unpatched — only `fetch`, `https.request`, and `http.request`
  were wrapped — so a streaming client using `http.get` against a local
  Ollama endpoint slipped past the proxy; it is now patched (preserving
  the implicit `end()`). (b) The LLM path-suffix detector did not
  recognize the Bedrock InvokeModel REST shapes (`/invoke`,
  `/invoke-with-response-stream`), so those calls were not classified as
  LLM traffic; both suffixes are now matched. (c) The interceptor logged
  the raw request URL, leaking provider credentials carried in the query
  string (Gemini `?key=`, AWS SigV4 `?X-Amz-Signature=`) into plaintext
  logs; every query-parameter value is now redacted before any log or
  egress-telemetry hop.

### 3. Hostile-repo Git RCE during plugin scanning

`internal/gateway/codex_hook.go` shelled out to `git` in the inherited
process environment. A hostile repository config could weaponize the
scan:

* `core.hooksPath` → run an attacker script
* `core.editor` / `core.sshCommand` → command substitution
* `core.fsmonitor` → silent watcher exec
* `protocol.<x>.allow` → smuggle ext:// transports

New `internal/gitsafe` package wraps `exec.Command("git", ...)` with
`GIT_CONFIG_NOSYSTEM=1`, `GIT_CONFIG_GLOBAL=/dev/null`, scrubbed
`HOME`, and a layered `GIT_CONFIG_KEY_<n>/VALUE_<n>` override that
neutralizes the dangerous keys at the repo-local layer too. Every git
operation in the gateway now goes through `gitsafe.Command`.

### 4. Default rule pack had six trivial bypasses

* `C2-METADATA-GCP` was case-sensitive; `METADATA.GOOGLE.INTERNAL`
  reached the GCP IMDS without firing the Critical SSRF rule.
* PATH-`<credential>` rules home prefix `(?:~|$HOME|/home/\w+|/root)`
  did not match `${HOME}/.aws/credentials` (the braced shell form), so
  any tool call referencing the canonical shell expansion slipped
  through.
* `PATH-ENV-FILE` suffix delimiter class lacked `;`, so `cat .env;
  true` was not flagged.
* `CMD-REVSHELL-NC` only matched `nc <host> <port> -e /bin/sh`; the
  equally common `nc -e /bin/sh <host> <port>` flag-first ordering was
  unmatched.
* `CMD-REVSHELL-DEVTCP` only matched dotted-quad IPv4; `/dev/tcp/<host
  name>/<port>` (which bash supports natively) was unmatched.
* `CMD-PIPE-{CURL,WGET,BASE64}` only matched bare `sh`/`bash` after
  the pipe; `curl … | /bin/sh` (absolute path) was unmatched.

All six are patched in lockstep across the YAML rule packs, the
runtime scanner defaults, and the in-process scanner fallback.

### 5. Audit-sink token leak (TLS verify defaulted OFF)

Splunk HEC and HTTP-JSONL sinks defaulted to `InsecureSkipVerify=true`
and required operators to opt **into** certificate validation via
`verify_tls=true`. Operators who omitted the flag silently shipped HEC
tokens (`Authorization: Splunk ...`) and bearer-token webhook payloads
in plaintext-equivalent form to any on-path peer.

Contract is now flipped: TLS verify **on** by default. The legacy
`verify_tls=false` field is parsed but silently ignored — the secure
default wins. Operators that genuinely need the insecure path (dev
environments with self-signed HEC) must set `insecure_skip_verify:
true` explicitly, and a boot warning is emitted whenever the opt-out
is honoured against an HTTPS endpoint.

### 6. Daemon PID-reuse / spawn-race bypass

* `Daemon.Start()` left a stale `gateway.pid` behind when the spawned
  child crashed during startup. `processExists()` returned true for
  unreaped zombies (`FindProcess + Signal(0)` is not a real liveness
  check on Linux).
* After PID reuse, `verifyProcess()` accepted ANY live process at the
  recorded PID whose executable matched, because the `StartTime` field
  was a wall-clock-from-the-parent value and was never compared to the
  kernel's view of the process.

New `processStartIdentity(pid)` reads the kernel-side starttime
(`/proc/<pid>/stat` field 22 on Linux, `ps -o lstart=` on Darwin); the
identity is captured at spawn, persisted to `gateway.pid`, and
re-checked on every `Stop()` / `IsRunning()`. Combined with an
exit-watcher goroutine that detects children dying inside the 100ms
grace window, stale PID files cannot pin the daemon to an unrelated
process.

### 7. MCP scanner fail-open on malformed configs

* `parseConfigFile` / `loadMCPConfigs` swallowed every JSON / YAML /
  read failure and returned an empty server list. The scanner then
  emitted an INFO-only "No MCP server configurations found" finding,
  and `PolicyEnforcer`'s default `warnOnSeverity=MEDIUM` treated that
  INFO finding as `clean` — so a hostile or corrupted MCP config could
  skip every server check by being syntactically malformed. Loader now
  returns `{configs, errors}` and the scanner emits a Critical
  "Malformed MCP config rejected" finding per parse failure.
* Dangerous env-var **references** (e.g. `OPENAI_API_KEY:
  "${OPENAI_API_KEY}"`) only emitted INFO; the enforcer collapsed them
  to `clean` and admitted MCP servers that asked for sensitive host
  credentials. Bumped to High.
* `DANGEROUS_ENV_KEYS` only listed OpenAI / Anthropic / AWS keys;
  Google, Gemini, Azure, Mistral, Cohere, DeepSeek, Together, Groq,
  Perplexity, HuggingFace, and Replicate were not flagged. Set now
  mirrors the cross-runtime provider table.

### 8. Plugin scanner gaps

* Scanned only `.ts/.js/.mjs`. `.cjs/.cts/.jsx/.tsx` were never
  scanned, and `dist/` (the npm runtime entrypoint) was unconditionally
  skipped.
* Followed symlinked `openclaw.plugin.json` without `O_NOFOLLOW` →
  trivial TOCTOU against the manifest read.
* `SRC-CHILD-PROC` was Medium and `SRC-EXEC` was High but neither was
  enabled in the default profile — direct RCE primitives weren't
  blocking install.
* `defenseclaw plugin install --force` warn-and-installed even when
  the scanner reported High/Critical findings. Now refuses unless the
  operator records explicit risk acceptance via `defenseclaw plugin
  allow <name>` first.

### 9. Sandbox / pre-pair on a planted `device.key`

`cmd_setup_sandbox.py` would pre-pair from any `device.key` it
discovered, so another local user could plant a key and silently
hijack the pairing record. The gateway now writes a sibling
`device.key.provenance` sentinel; pre-pair refuses any key without
the sentinel (or with loose mode/owner). `chown -R` /
`setfacl -R` operations are pinned to `openclaw_home_original` so
they cannot follow attacker-pointed symlinks, and `chown` over system
roots is rejected outright. Operator opt-out:
`DEFENSECLAW_PREPAIR_TRUST_DEVICE_KEY=1`.

### 10. Codex connector loopback was always trusted

`internal/gateway/connector/codex.go` accepted any 127.0.0.1 caller
without authentication, on the theory that loopback was a hardware
trust boundary. On a shared host that's not true. With
`DEFENSECLAW_GATEWAY_TOKEN` configured the codex connector now
requires `X-DC-Auth`, the master key, or an `Authorization: Bearer`
that matches a recorded codex provider api_key. Native `codex` CLI
continues to authenticate via its provider key. Operator opt-out:
`DEFENSECLAW_CODEX_LOOPBACK_TRUST=1`.

### 11. Upgrade / install lacked integrity verification

* `defenseclaw upgrade` and `scripts/upgrade.sh` downloaded gateway
  tarballs and wheels with no checksum / signature step. A MITM or a
  compromised GitHub release tag could swap the artifact silently.
  Now verifies every download against an operator-pinned `sha256` OR a
  published `<artifact>.sha256` sidecar before installing. Operator
  opt-out: `DEFENSECLAW_UPGRADE_ALLOW_UNVERIFIED=1`.
* `scripts/install-openshell-sandbox.sh` installed from a mutable OCI
  tag with no digest pin → vulnerable to tag mutation. Now requires
  `DEFENSECLAW_OPENSHELL_ARCH_DIGEST` or
  `DEFENSECLAW_OPENSHELL_BINARY_SHA256` (or explicit
  `DEFENSECLAW_OPENSHELL_ALLOW_UNPINNED=1`).

### 12. Firewall bypass on dual-stack hosts

* `internal/firewall/iptables/iptables.go` applied rules as IPv4 only.
  On any dual-stack host, IPv6 egress went completely unfiltered.
  Rules are now split into v4 / v6 sections and piped to
  `iptables-restore` and `ip6tables-restore`.
* `internal/firewall/pfctl/pfctl.go` shipped with a blanket
  `pass out quick inet proto udp keep state` rule that let arbitrary
  outbound UDP/QUIC bypass default-deny. Removed.

### 13. Scoped-tool blocks were stored but never enforced

`defenseclaw tool block <name> --source <agent>` wrote a scoped audit
entry but the runtime enforcer only consulted the global block list,
so per-agent blocks did nothing in production. Scoped blocks now
additionally write a global block; the audit record preserves the
operator's scoping intent for later reporting. Tracked for a future
runtime-scope enforcement pass.

### 14. CI: fork PRs ran on self-hosted runner with full secrets

`.github/workflows/e2e.yml` ran the `core` E2E job on a long-lived
self-hosted ARM64 runner with `OPENCLAW_GATEWAY_TOKEN`,
`OPENAI_API_KEY`, `ANTHROPIC_API_KEY`, `BIFROST_API_KEY`, and
`SPLUNK_ACCESS_TOKEN` exposed via its env block. The `if:` clause
only filtered scheduled runs, so any fork PR (including drive-by) ran
attacker-controlled code on the self-hosted runner with full secret
access AND persistent shell foothold. Now both self-hosted jobs share
the same fork-PR exclusion the `full-live` job already had; external
contributors get the GitHub-hosted shadow job.

### 15. Scan results leaked secrets back through API and JSON sinks

* `internal/cli/scan_v7.go` (`defenseclaw scan code --json`) did not
  run `redaction.ForSinkString` on Description / Location /
  Remediation. Secrets matched by scanners were echoed back verbatim
  in the JSON output.
* Gateway REST `POST /api/v1/scan/code` had the same gap, serializing
  the raw `scanner.ScanResult` to the response body.
* AI-discovery events emitted plain `sha256:` path digests, which are
  trivially reversible by dictionary attack against well-known paths
  (`~/.aws/credentials`, etc.). Now derived as
  `HMAC-SHA256(api_token, "ai-discovery/path-hash/v1")` and installed
  synchronously during sidecar boot.

### 16. Environment variable surface had no inventory

`~70` `DEFENSECLAW_*` env vars accreted across Go, Python, TypeScript,
and shell, with no central inventory. Several materially weakened
security defaults (`DEFENSECLAW_DISABLE_REDACTION`,
`DEFENSECLAW_OTEL_TLS_INSECURE`,
`DEFENSECLAW_LOCAL_USERNAME/PASSWORD`, the AWS HTTP/1 shim escape
hatches, `DEFENSECLAW_HOOK_MAX_BODY`, the per-signal OTel surface,
etc.) and were undocumented. There was also no mechanism to prevent
silent drift.

* `internal/envvars/registry.json` is now the single source of truth
  for every env var, with category, purpose, default, accepted values,
  security impact, consumers, since-version, and a public
  `security_note` for opt-outs. Both Go and Python load the same blob.
* `defenseclaw doctor` surfaces every active security opt-out with the
  matching `security_note` and the recommended replacement.
* `docs/ENV-VARS.md` and `docs-site/content/docs/reference/env-vars.mdx`
  are regenerated from the registry via `scripts/gen_envvars_docs.py`.
* `cli/tests/test_envvars_codebase_coverage.py` is a CI gate: any new
  `DEFENSECLAW_*` reference that isn't declared in the registry fails
  the build.
* `internal/envvars/sync_test.go` pins Go and Python to the same
  registry data.

---

## Operator-visible behavior changes (read before upgrading)

These are intentional fail-closed flips; each has an env opt-out for
operators who need the old behavior temporarily during migration.

| Change | Default before | Default after | Opt-out env var |
|---|---|---|---|
| Guardrail hook fail mode | `open` | `closed` (new installs only) | `hook_fail_mode: open` in config |
| Splunk HEC / HTTP-JSONL TLS verify | OFF | ON | `insecure_skip_verify: true` |
| Webhook redirects | followed default policy | re-validated each hop, http(s) only, 5 max | `DEFENSECLAW_WEBHOOK_ALLOW_LOCALHOST=1` |
| `defenseclaw upgrade` integrity | not verified | sha256 required | `DEFENSECLAW_UPGRADE_ALLOW_UNVERIFIED=1` |
| `install-openshell-sandbox.sh` | mutable tag accepted | digest/sha256 required | `DEFENSECLAW_OPENSHELL_ALLOW_UNPINNED=1` |
| Codex connector loopback trust | implicit | requires auth when token configured | `DEFENSECLAW_CODEX_LOOPBACK_TRUST=1` |
| Agent binary discovery trust roots | included user-writable dirs (e.g. the Codex CLI standalone root under `~/.codex/packages`) | system dirs only; user-writable roots require explicit opt-in | `DEFENSECLAW_TRUSTED_BIN_PREFIXES=<colon-separated paths>` |
| Sandbox pre-pair `device.key` | accepted any | requires gateway-written provenance | `DEFENSECLAW_PREPAIR_TRUST_DEVICE_KEY=1` |
| CGNAT / ECS metadata / IPv6 ULA dial | allowed | blocked | `DEFENSECLAW_ALLOW_CGNAT=1` (CGNAT only) |
| Plugin `tsx/jsx/cjs/cts` + `dist/` | skipped | scanned | — |
| Plugin install with High/Critical | warn-and-install on `--force` | refuse | `defenseclaw plugin allow <name>` first |

**Migration: trailing-dot provider entries.** Operator
`~/custom-providers.json` entries that use the legacy trailing-dot
prefix syntax (e.g. `"bedrock-runtime."`) **must** migrate to a
wildcard pattern (e.g. `"bedrock-runtime.*.amazonaws.com"`). The
trailing-dot form is no longer honoured because it accepted spoofed
hosts like `bedrock-runtime.evil.example`.

**Migration: Codex CLI standalone discovery.** Agent binary discovery
no longer trusts user-writable directories by default, so a Codex CLI
installed under its standalone root (`~/.codex/packages/.../bin`) is not
auto-discovered. Operators who rely on that layout opt the path back in
with `DEFENSECLAW_TRUSTED_BIN_PREFIXES=/abs/path[:/another]`; system
locations (`/usr/bin`, `/usr/local/bin`, etc.) remain trusted by
default.

**Doctor surfaces every active opt-out.** `defenseclaw doctor` now
prints a "Security overrides" row for each opt-out env var that is
currently set, with a prose description of what's being weakened and
the recommended replacement. Audit operators can see at a glance which
security defaults are relaxed on a given host.

---

## Verification

Re-verified on the branch rebased onto `main`; all CI jobs are green.

| Step | Result |
|---|---|
| `go build ./...` | clean |
| `go vet ./...` | clean |
| `go test ./... -count=1 -timeout=300s` | 32/32 packages with tests pass (0 failures) |
| `make ts-test` (extensions/defenseclaw) | 253/253 pass (20 files) |
| `npx tsc --noEmit --strict` | clean |
| `make cli-test` (`pytest cli/tests`) | 3,407 passed, 3 skipped, 243 subtests |
| `make rego-test` | 90/90 pass |
| `make check-schemas` | clean (Go ↔ embed.FS mirror byte-identical) |
| `pytest cli/tests/test_envvars_codebase_coverage.py` | 3/3 pass |
| `pytest cli/tests/test_envvars.py` | 28/28 pass |
| `pytest cli/tests/test_cmd_doctor.py` | 55/55 pass |

**Manual acceptance test** (this branch, against live OpenClaw on a
real workstation): 22 scenarios across guardrail content inspection,
tool-call inspection (rm -rf, /etc/shadow, AWS IMDS exfil, secret
leakage in tool output), CLI flows (`tool block/allow/unblock`,
`setup webhook add` with private/IMDS/CGNAT URLs), doctor output for
all active opt-outs, gateway auth (Bearer + `X-DefenseClaw-Client`),
edge cases (invalid JSON 400, empty content 200, oversized 504,
wrong-method 405), rate limiting (loopback exempt), and OpenClaw
plugin handshake / heartbeat / agent discovery. All 22 passed.

## Test plan

- [x] All language-level test suites green (above).
- [x] Manual acceptance against a live OpenClaw + DefenseClaw on a
      real workstation.
- [x] CI: `Go Test`, `TypeScript Build & Test`, `Rego Policy Tests`,
      `Python Lint & Test`, `make test (unified)`, `Go Lint`,
      `Windows Hook Path`, `Schema & Parity Checks`, and the docs
      `Build static export` jobs green.
- [x] CI: `e2e` connector matrix (openclaw / zeptoclaw / claudecode /
      codex) green.
- [ ] Reviewer spot-checks of new `*_hardening_test.go` and
      `*_failclosed_test.go` regression files added in this PR.

## Documentation

* `CHANGELOG.md` — new pre-launch hardening section grouped by public
  severity (Critical / High / Medium), no internal codenames.
* `docs/ENV-VARS.md` and `docs-site/content/docs/reference/env-vars.mdx`
  — auto-generated from the registry, list every `DEFENSECLAW_*` var
  with its security note and replacement hint.
* `SECURITY.md` — unchanged (procedural disclosure process; not a CVE
  log).


vineethsai:
## Update: env-var registry now landed (commit `1fc960be`)

Following the audit you posted listing ~70 `DEFENSECLAW_*` variables across the repo, I've added a **single source of truth** for all of them instead of patching the docs file ad-hoc.

### What landed

| Component | File | Purpose |
|---|---|---|
| Registry JSON | `internal/envvars/registry.json` | All ~70 vars with category, purpose, default, security_impact, consumers (file:line), security_note (prose description of the defense the var gates), replacement hints |
| Go loader | `internal/envvars/registry.go` | `//go:embed`'d at build time; typed API + validation |
| Python loader | `cli/defenseclaw/envvars.py` | Mirror loader; same semantics |
| Doctor wiring | `cli/defenseclaw/commands/cmd_doctor.py` | New **Security overrides** section — surfaces every active opt-out as a warn row with a one-line description of what the bypass weakens |
| Docs generator | `scripts/gen_envvars_docs.py` | Renders `docs/ENV-VARS.md` + `docs-site/.../env-vars.mdx` from the registry |
| Plain-MD docs | `docs/ENV-VARS.md` | New, auto-generated |
| Docs-site update | `docs-site/content/docs/reference/env-vars.mdx` | Body tables now auto-generated; previously missing rows (LOCAL_USERNAME/PASSWORD, OTEL_TLS_INSECURE, AWS HTTP/1 shim, HOOK_MAX_BODY, per-signal OTel, etc.) now present |
| Coverage gate | `cli/tests/test_envvars_codebase_coverage.py` | Walks the repo, fails CI on any `DEFENSECLAW_*` reference not in the registry |
| Cross-lang sync | `internal/envvars/registry_test.go#TestCrossLanguageSync` | Go and Python loaders **must** agree on the name set — drift in either language fails CI in either language |

### How this answers "too many env vars"

The variables themselves aren't going away — most are legitimate (debug flags, runtime-path overrides, hook-internal identity propagation, credential indirection). The actual problem was **invisibility**:

1. **`defenseclaw doctor` now reports active overrides**, so a forgotten `DEFENSECLAW_DISABLE_REDACTION=1` from a debugging session is impossible to miss
2. **Docs are auto-generated from the registry**, so drift between code and docs is now mechanically impossible
3. **A CI gate fails the build** if anyone adds a new `DEFENSECLAW_*` reference without registering it
4. **A cross-language sync test** ensures Go and Python never disagree about what's registered

### Re: the 5 "not yet wired" opt-outs

After your audit message, I re-grep'd this PR branch (not main) and confirmed all 5 *are* wired in this PR:
- `DEFENSECLAW_PREPAIR_TRUST_DEVICE_KEY` → `cli/defenseclaw/commands/cmd_setup_sandbox.py:1759`
- `DEFENSECLAW_CODEX_LOOPBACK_TRUST` → `internal/gateway/connector/codex.go` (recent commit)
- `DEFENSECLAW_UPGRADE_ALLOW_UNVERIFIED` → `cli/defenseclaw/commands/cmd_upgrade.py` + `scripts/upgrade.sh`
- `DEFENSECLAW_OPENSHELL_ALLOW_UNPINNED` → `scripts/install-openshell-sandbox.sh`
- `DEFENSECLAW_DUMP_RAW_SECRETS` → `scripts/test-e2e-full-stack.sh`

(Audit was likely run against `main`; these were added in earlier commits on this branch.) The new coverage test will keep us honest going forward.

### Migration impact

**Zero.** No env var was renamed, removed, or had its semantics changed. Pure additive: registry + docs + tests + one new doctor section.

### Pre-existing test failures

3 modules in `cli/tests/` import `pytest` directly which isn't in the venv (`test_ai_signatures`, `test_connector_mcp_writers`, `test_connector_paths`) — those fail on `main` too, not from this PR. The remaining **2101/2104** Python tests pass, plus the new **65** envvars+doctor tests, plus the full Go envvars suite including cross-lang sync.

PR #282, opened 2026-05-18
Fix OpenClaw restart overwriting Discord allowance

barryqy:
## Summary

This fixes a restart regression where the managed OpenClaw extension could be redeployed on gateway startup and overwrite a local Discord posting hotfix.

The Discord channel message endpoint looks like an LLM `/messages` API by path suffix, so it could be routed through DefenseClaw enforcement and blocked after the generated extension bundle reverted. The fix adds explicit known-non-LLM handling for Discord channel message endpoints in both the OpenClaw extension and gateway classifier, and adds corpus/test coverage for the overlap.

The connector installer is also now idempotent for unchanged managed bundles. It writes a managed bundle hash marker and skips redeploying the extension on same-version restarts, while still redeploying if required files are missing or the embedded bundle changes. `openclaw.json` writes are skipped when the content is already byte-identical.

## Validation

- `npm test -- --run src/__tests__/shape-detection.test.ts src/__tests__/provider-coverage.test.ts`
- `make sync-openclaw-extension && go test ./internal/gateway -run 'Test(IsLLMPathSuffix|IsKnownNonLLMEndpoint|ProviderCoverageCorpus)' -count=1`
- `make sync-openclaw-extension && go test ./internal/gateway/connector -run 'TestOpenClawExtensionCurrent|TestInstallOpenClaw_SymlinkedExtDir' -count=1`
- `npm run build`
- `go test ./internal/gateway ./internal/gateway/connector -count=1`
- `npm test`
- `git diff --check`

### DIFF
--- a/docs-site/content/docs/connectors/openclaw.mdx
+++ b/docs-site/content/docs/connectors/openclaw.mdx
@@ -149,6 +149,10 @@ Not sure what to pick? Run `defenseclaw setup guardrail` (no flags) — the [int
 
 A hash-checked backup of `openclaw.json` is stored before edits; teardown restores or surgically removes only DefenseClaw-owned entries.
 
+<Callout>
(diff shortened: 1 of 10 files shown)


barryqy:
Closing this for now while we tighten the scope around the global fetch false-positive behavior.

PR #304, opened 2026-05-25
feat(config): add in-memory key store for secret hydration

nghodkicisco:
Secrets fetched from OpenBao at startup can now be stored in memory via SetKey/GetKey rather than injected into the process environment. ResolvedAPIKey() on both LLMConfig and CiscoAIDefenseConfig checks the key store before os.Getenv, keeping secrets out of /proc/environ.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -329,12 +329,18 @@ type LLMConfig struct {
 // these stay in lock-step.
 func (l LLMConfig) ResolvedAPIKey() string {
 	if l.APIKeyEnv != "" {
+		if v, ok := GetKey(l.APIKeyEnv); ok && strings.TrimSpace(v) != "" {
+			return strings.TrimSpace(v)
+		}
 		if v := strings.TrimSpace(os.Getenv(l.APIKeyEnv)); v != "" {
 			return v
 		}
 	}
 	if l.APIKey != "" {
 		return l.APIKey
+	}
+	if v, ok := GetKey(DefenseClawLLMKeyEnv); ok && strings.TrimSpace(v) != "" {
+		return strings.TrimSpace(v)
 	}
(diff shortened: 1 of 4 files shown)


rucpande (commented):
Should resolveConfiguredProvider in proxy.go also use the key store, or is the assumption that the fetch interceptor will always be used, so that use case doesn't need to be accounted for?

nghodkicisco:
> Should resolveConfiguredProvider in proxy.go also use the key store, or is the assumption that the fetch interceptor will always be used, so that use case doesn't need to be accounted for?

No, resolveConfiguredProvider doesn't need the key store. The two paths serve different use cases with different key-resolution mechanisms:

Token resolver path (line 1552): In enterprise, SetTokenResolver is registered by the token hydrator in main.go. This is the gateway's proxy key — it resolves per-provider keys for the agent's upstream LLM calls. It's already checked first in resolveConfiguredProvider.

Key store path (config.GetKey): This is for the judge and Cisco AI Defense keys — used by LLMConfig.ResolvedAPIKey() and CiscoAIDefenseConfig.ResolvedAPIKey(). Those are internal DefenseClaw calls, not the agent's proxied traffic.

The assumption is correct: in the direct-provider path (no fetch interceptor), the gateway proxy key comes from either:

The external tokenResolver (enterprise: sidecar-hydrated), OR
req.TargetAPIKey (connector-supplied), OR
p.cfg.APIKeyEnv via ResolveAPIKey (local/OSS: env var or dotenv)
The key store is deliberately scoped to config.*ResolvedAPIKey() consumers (judge LLM, AID inspect). Mixing it into the gateway proxy path would conflate two distinct credential domains — the agent's upstream provider key vs. DefenseClaw's internal service keys.

rucpande (approved):

PR #317, opened 2026-05-29
[REM-95] add reveal toggle to masked secret fields

Rmontalbano:
## Summary

Secret-like fields (API keys, tokens) are already masked on this branch — `SecureField` + secret-name detection in the Config editor, and `.password` kind in Setup — but there was no way to verify what was typed.

This adds a small reusable **`RevealableSecureField`** (show/hide eye toggle that stays `.privacySensitive()` while masked) and wires it into the three existing masked render sites:

- Setup `.password` control (`SetupView`)
- Setup workflow `.password` field (`SetupView`)
- Config editor secret-like scalar (`StructuredDocumentEditorView`)

The component encapsulates the styling the call sites previously applied inline, so the diff is net-negative in the edited files.

## Scope

Reveal affordance only. Secret **storage** routing (Keychain / managed `.env` instead of writing the raw value into `config.yaml`) is the bigger P0 and is tracked separately — it touches the secret model shared with the gateway, so worth aligning on approach first.

## Validation

- `swift build --package-path apps/shared`
- `swift build --package-path apps/appkit-app`
- `./script/build_and_run.sh --qa setup llm` — toggle works on the DefenseClaw LLM API Key field

Closes REM-95.

> Note: opened as a draft against `codex/reuse-pr62-macos-app` — happy to fold into PR #161 however you prefer, @vineethsai7.

### DIFF
--- a/apps/appkit-app/DefenseClawAppKit/Views/Setup/SetupView.swift
+++ b/apps/appkit-app/DefenseClawAppKit/Views/Setup/SetupView.swift
@@ -523,11 +523,10 @@ private struct SetupFieldCard: View {
             .labelsHidden()
             .pickerStyle(.menu)
         case .password:
-            SecureField(field.placeholder, text: Binding(
+            RevealableSecureField(placeholder: field.placeholder, text: Binding(
                 get: { model.textBindingValue(for: field) },
                 set: { model.setText($0, for: field) }
             ))
-            .textFieldStyle(.roundedBorder)
         default:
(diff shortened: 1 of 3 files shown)


Rmontalbano:
Closing as superseded. Per the plan to consolidate into fewer/larger PRs and rebase the macOS work on top of #284 (`feat/hook-collector-unification`), this reveal-toggle change (REM-95) is carried forward in the rebased macOS branch and will land via the larger PR. Thanks @vineethsai7!

PR #334, opened 2026-06-02
fix(gateway): advertise protocol v4 in connect handshake (support v3 and v4)

gthermaenius:
## Problem

The gateway client advertises only protocol **v3** in the connect handshake (`internal/gateway/client.go`, `sendConnect`: `minProtocol: 3, maxProtocol: 3`). OpenClaw gateways running protocol **v4** (e.g. `2026.5.28`) reject the connect request and close the socket:

```
[gateway] connect error: code=INVALID_REQUEST message=protocol mismatch
[gateway] read error: websocket: close 1002 (protocol error): protocol mismatch
```

The WebSocket upgrade (HTTP 101) and challenge/response auth succeed; only protocol negotiation fails, because the advertised range does not include `4`. There is no config/flag/env override for the range.

## Fix

Widen the advertised range to `[3, 4]` (`minProtocol: 3, maxProtocol: 4`). A gateway accepts any client range that includes its own version, so this negotiates with **both** v3 and v4 gateways — backward compatible, not a breaking change (per the SemVer note in CONTRIBUTING). The now-stale comment and the connect log line are updated to match.

## Testing

- New unit test `TestConnectAdvertisesProtocolRange3To4` stands up a mock gateway, captures the connect frame, and asserts `[minProtocol, maxProtocol] = [3, 4]`.
- Full package suite passes: `go test ./internal/gateway/`.
- Verified end-to-end against a live OpenClaw `2026.5.28` gateway: a build advertising v4 completes the handshake (`[gateway] handshake complete`) where v3 was rejected with close `1002`. `[3, 4]` is accepted by that same gateway because the range includes `4`.

## Notes

Protocol v4 may introduce payload/framing changes beyond the version number; this PR addresses only the negotiation gate. Follow-ups welcome if v4 requires additional connect fields.

### DIFF
--- a/internal/gateway/client.go
+++ b/internal/gateway/client.go
@@ -95,7 +95,7 @@ func (c *Client) wsURL() string {
 	return u.String()
 }
 
-// Connect establishes the WebSocket connection and completes the protocol v3
+// Connect establishes the WebSocket connection and completes the protocol v3/v4
 // handshake including device challenge-response authentication.
 func (c *Client) Connect(ctx context.Context) error {
 	target := c.wsURL()
@@ -135,7 +135,7 @@ func (c *Client) Connect(ctx context.Context) error {
 	c.startHandshakeEventBuffer()
 	go c.readLoop()
 
-	fmt.Fprintf(os.Stderr, "[gateway] sending connect (protocol=3, role=operator, device=%s) ...\n",
+	fmt.Fprintf(os.Stderr, "[gateway] sending connect (protocol=3-4, role=operator, device=%s) ...\n",
 		c.device.DeviceID)
 	hello, err := c.sendConnect(ctx, nonce)
 	buf := c.stopHandshakeEventBuffer()
@@ -262,7 +262,7 @@ func (c *Client) sendConnect(ctx context.Context, nonce string) (*HelloOK, error
 
 	params := map[string]interface{}{
 		"minProtocol": 3,
-		"maxProtocol": 3,
+		"maxProtocol": 4,
 		"client": map[string]interface{}{
 			"id":       clientID,
 			"version":  "1.0.0",
--- a/internal/gateway/client_protocol_test.go
+++ b/internal/gateway/client_protocol_test.go
@@ -1,0 +1,92 @@
+// Copyright 2026 Cisco Systems, Inc. and its affiliates
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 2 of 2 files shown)


gthermaenius:
## Protocol v4 compatibility analysis

To confirm this isn't just a version-number flip that ignores other v4 changes, I traced what protocol v4 actually changed in `openclaw/openclaw` and checked it against this client's behavior.

**What v4 changed (vs v3):**
- The connect handshake gate is **purely numeric** — the gateway accepts any client whose `[minProtocol, maxProtocol]` range includes `4` (`src/gateway/server/ws-connection/message-handler.ts`). No new *required* connect fields, **no framing changes**, no new required caps/scopes. Required connect fields remain `minProtocol`, `maxProtocol`, and `client{...}`, all already sent here.
- The only client-visible breaking change is the **chat delta event** schema (openclaw/openclaw@`150bebcd0c`, "require v4 chat deltas"): `deltaText` became required and a `replace` flag was added. Everything else new in v4 (`tasks.*`, `session.operation`, …) is additive / opt-in.

**Why that one change does not affect this client:**
- `EventRouter` does consume `chat` events, but `handleChatEvent` (`internal/gateway/router.go`) only reads `state` and `errorMessage`, and acts solely on `state == "error"`. It never reads `deltaText` / `text` / `replace`, and Go's JSON decoder ignores the changed/added fields — so the v4 delta restructuring is a no-op here.
- The client is otherwise a generic event pump: unknown frame types and unhandled event names are logged, not errored. Its enforcement path is `tool_call` / `tool_result` (advertised via `caps:["tool-events"]`), which v4 left unchanged.

**Verdict:** advertising `[3, 4]` is sufficient — no additional connect fields, framing, or event-handling changes are required for v4. Consistent with end-to-end verification against a live OpenClaw `2026.5.28` gateway (handshake completes; tool events flow).

*Residual (disclosed for completeness): if a future change also renamed the chat **error** event's `state`/`errorMessage` fields, this client could silently skip logging some chat errors — graceful degradation, not a broken connection. No evidence v4 did this; only the delta variant changed.*

---
<sub>This protocol-v4 compatibility analysis was performed by **Claude Opus 4.8** to independently review and verify the change.</sub>

gthermaenius:
Closing as a duplicate of #303, which implements the same OpenClaw protocol v3→v4 handshake fix — the same backward-compatible `[3, 4]` range — and is both earlier (opened May 25) and more complete (it also updates the docs and the existing test fixtures). Deferring to #303; thanks @barryqy.

For reference, the thread above includes an independent protocol-v4 compatibility analysis confirming that advertising `[3, 4]` is sufficient (no other client-side changes — connect fields, framing, or event handling — are required for v4). That may be useful context for #303's review.

PR #337, opened 2026-06-03
feat(gateway): auto-restore connector hook config when removed at runtime

vineethsai:
> Stacked on top of #284 (`feat/hook-collector-unification`). Review/merge that PR first; this PR's base is the stack parent, not `main`.

## Problem

DefenseClaw enforcement for hook-native agents depends entirely on a hook
block living inside each agent's own config file (e.g. `~/.cursor/hooks.json`,
`~/.claude/settings.json`, `~/.codex/config.toml`). If a user edits that file
and deletes or strips the DefenseClaw hook entries **while the gateway is
running**, enforcement silently lapses: the agent keeps running with no
pre/post-tool inspection, and nothing restores the hook until the next
`setup`/restart or a connector switch. There is no detection and no recovery
of this tamper case at runtime.

## Current Situation

- Connectors install their hook block during `Connector.Setup()` and remove it
  during `Teardown()`. Once Setup has run, nothing watches the file.
- `defenseclaw doctor` can detect drift on demand, but it is not running
  continuously and does not auto-repair connector config files.
- Hook-native connectors (codex, claudecode, cursor, windsurf, ...) run in
  observability-only mode and never reach `proxy.Run`, so any recovery must be
  wired on a path both proxy-bound and observability-only connectors hit.
- Net effect: a single manual edit disables enforcement indefinitely with no
  audit signal.

## Solution

Add an event-driven self-heal guard that watches the active connector's config
file(s) and re-installs the hook block the moment it goes missing.

- `HookConfigGuard` (`internal/gateway/hook_config_guard.go`) watches the
  **parent directory** of each resolved config path via `fsnotify` (so atomic
  editor rename/replace saves are caught), debounces bursts, and on a debounced
  event runs a **presence check**. It only re-runs `conn.Setup()` when the
  owned hook command is actually gone, so legitimate unrelated edits cause no
  churn. It suppresses its own re-install writes to avoid loops.
- Presence detection (`internal/gateway/connector/hook_config_paths.go`) reuses
  the existing `ResolvedConnectorLocations` path contract and the
  `HookScriptOwner` interface, then does a format-agnostic substring match for
  the exact hook command the patchers write (works for JSON, TOML, YAML).
  Proxy/plugin connectors (openclaw, zeptoclaw) implement no `HookScriptOwner`
  and are treated as inert.
- `GuardrailProxy` owns one guard, starts it in `runGuardrail` **before** the
  observability short-circuit (covering both connector classes), `Repoint`s it
  on runtime connector hot-swap, and `SuppressHealing` across the swap so the
  connector being torn down is not re-installed.

**Why this approach over alternatives:** polling would add latency and wasted
work; inverting `VerifyClean` would over-trigger on benign edits. Event-driven
+ presence-gated re-Setup is precise, low-overhead, and reuses the connectors'
existing idempotent Setup so there is no second code path to keep in sync.

Gated by `guardrail.hook_self_heal` (default **true**); set false to restore
the prior behavior (hooks stay removed until next setup/restart).

## Architecture Diagram

```
Before:
  ┌──────────┐  Setup() writes hook  ┌────────────────────┐
  │Connector │──────────────────────▶│ agent config file  │
  └──────────┘                       │ (~/.cursor/...json) │
                                     └────────────────────┘
        (user deletes hook block) ─────────▶  enforcement lapses, no recovery

After:
  ┌──────────────┐ owns ┌───────────────┐ fsnotify ┌────────────────────┐
  │GuardrailProxy│─────▶│HookConfigGuard │◀─────────│ parent dir of cfg  │
  └──────┬───────┘      └───────┬───────┘  events    └─────────┬──────────┘
         │ Repoint/Suppress     │ presence-gated re-Setup       │
         │ on connector switch  ▼                               ▼
         │              ┌──────────────┐ rewrites    ┌────────────────────┐
         └─────────────▶│ Connector    │────────────▶│ agent config file  │
                        │ .Setup()     │             └────────────────────┘
                        └──────────────┘
                               │ audit + OTel + webhook on heal
                               ▼
                     connector-hook-tampered / connector-hook-repaired
```

## Flow Diagram

```
User           fsnotify        HookConfigGuard          Connector        Audit/OTel/Webhook
 │ delete hook  │                │                        │                  │
 │─────────────▶│ write/remove   │                        │                  │
 │              │───────────────▶│ record pending         │                  │
 │              │                │ (debounce tick)        │                  │
 │              │                │ OwnedHooksPresent? no  │                  │
 │              │                │ log tampered ──────────┼─────────────────▶│
 │              │                │ Setup(ctx) ───────────▶│ re-patch hook    │
 │              │                │ suppress self-write    │                  │
 │              │                │ re-sync watch set      │                  │
 │              │                │ log repaired + metric ─┼─────────────────▶│
 │              │                │ onHealed ──────────────┼──▶ webhook       │
```

## What Changed (Where & How)

| File | Change Type | Summary |
|---|---|---|
| `internal/gateway/connector/hook_config_paths.go` | added | `HookConfigPathsForConnector` + `OwnedHooksPresent`: resolve watched config files and detect the owned hook command (format-agnostic); inert for proxy/plugin connectors |
| `internal/gateway/hook_config_guard.go` | added | `HookConfigGuard`: fsnotify watch on parent dirs, debounce, self-write suppression, presence-gated re-Setup; `Start`/`Repoint`/`SuppressHealing`/`Stop` |
| `internal/gateway/proxy.go` | modified | Proxy owns the guard; `StartHookConfigGuard`, `notifyHookHealed` webhook fan-out, `Repoint` + `SuppressHealing` on runtime connector switch |
| `internal/gateway/sidecar.go` | modified | Start the guard in `runGuardrail` before the observability short-circuit, gated by `Enabled && HookSelfHeal` |
| `internal/config/config.go` | modified | Add `HookSelfHeal` (default true) + `HookSelfHealDebounceMs` to `GuardrailConfig`; viper defaults |
| `internal/config/defaults.go` | modified | Set defaults in `DefaultConfig()` |
| `internal/config/config_test.go` | test | Assert the new defaults |
| `internal/audit/actions.go` | modified | Add `ActionConnectorHookTampered` / `ActionConnectorHookRepaired` (+ `AllActions`) |
| `cli/defenseclaw/audit_actions.py` | modified | Mirror the two new action constants for Go/Python parity |
| `schemas/audit-event.json` | config | Add the two actions to the audit-event enum |
| `internal/cli/embed/audit-event.json` | config | Embedded schema parity for the two new actions |
| `internal/gateway/connector/hook_config_paths_test.go` | test | Unit tests for path resolution + presence detection (incl. inert proxy connectors, missing file) |
| `internal/gateway/hook_config_guard_test.go` | test | Integration tests: delete→restore, full-file recreate, unrelated-edit no-op, disabled no-op, repoint follows switch, notifier fires, suppress-then-resume |

## Open Issues / Follow-ups

- Transient `Setup` failure during a heal is logged as `guardrail-degraded`
  but not retried on a timer; recovery waits for the next filesystem event or a
  restart. Deferred — connector Setup is idempotent and failures are rare; a
  backoff-retry can be added if telemetry shows real occurrences.
- `SuppressHealing` uses a fixed 5s window across a connector hot-swap. A
  connector `Setup` that exceeds 5s would re-open a narrow race; abnormal and
  acceptable for now. TODO(@vineeth): revisit if slow-Setup connectors appear.
- Webhook fan-out reuses the existing dispatcher, which (correctly) rejects
  loopback endpoints via its SSRF allow-list; verified the callback path
  directly in tests rather than through a loopback receiver.

## Test Plan

- `gofmt -l` on changed files → clean
- `go build ./...` → OK
- `go vet ./internal/gateway/...` → OK
- `go test -race -run 'HookConfigGuard|OwnedHooksPresent|HookConfigPathsForConnector' ./internal/gateway/...` → PASS (delete→restore, recreate, unrelated-edit no-op, disabled no-op, repoint, notifier-fires, suppress-then-resume)
- `go test ./internal/config/...` → PASS (default-config assertions)
- `go test ./internal/audit/...` → PASS (audit action / schema parity)
- Manual end-to-end driver against the real `GuardrailProxy.StartHookConfigGuard` wiring + real audit SQLite store + real `WebhookDispatcher`: verified delete→restore, full-file recreate, unrelated-edit no-op, TOML (codex) restore, real `connector-hook-tampered`/`connector-hook-repaired` rows written, and the heal notifier firing with the correct connector name + path.
- Not tested: live multi-day soak; Windows path (native `defenseclaw-gateway hook` invocation) exercised via unit logic only, not on a Windows host.

### DIFF
--- a/cli/defenseclaw/audit_actions.py
+++ b/cli/defenseclaw/audit_actions.py
@@ -102,6 +102,13 @@ ACTION_CONNECTOR_HOOK: Final[str]              = "connector-hook"
 ACTION_CONNECTOR_HOOK_SYNTHETIC: Final[str]    = "connector-hook-synthetic"
 ACTION_ASSET_POLICY: Final[str]                = "asset-policy"
 
+# Connector hook self-heal. Mirrors
+# internal/audit/actions.go::ActionConnectorHook{Tampered,Repaired}.
+# The hook config guard re-installs a connector's hook block after a
+# user manually removes it while the gateway is running.
+ACTION_CONNECTOR_HOOK_TAMPERED: Final[str]     = "connector-hook-tampered"
+ACTION_CONNECTOR_HOOK_REPAIRED: Final[str]     = "connector-hook-repaired"
+
 # Codex notify webhook (agent-turn-complete et al.). The notify
 # bridge POSTs codex's JSON arg to /api/v1/codex/notify; the
 # gateway derives the action key from the payload's `type` field.
@@ -306,6 +313,8 @@ ALL_ACTIONS: Final[tuple[str, ...]] = (
     ACTION_CONNECTOR_HOOK,
     ACTION_CONNECTOR_HOOK_SYNTHETIC,
     ACTION_ASSET_POLICY,
+    ACTION_CONNECTOR_HOOK_TAMPERED,
+    ACTION_CONNECTOR_HOOK_REPAIRED,
     ACTION_CODEX_NOTIFY,
     ACTION_CODEX_NOTIFY_AGENT_TURN_COMPLETE,
     ACTION_CODEX_NOTIFY_MALFORMED,
--- a/internal/audit/actions.go
+++ b/internal/audit/actions.go
@@ -112,6 +112,16 @@ const (
 	ActionConnectorHookSynthetic Action = "connector-hook-synthetic"
 	ActionAssetPolicy            Action = "asset-policy"
 
+	// Connector hook self-heal. The hook config guard
+	// (internal/gateway/hook_config_guard.go) watches each active
+	// connector's agent config file (e.g. ~/.cursor/hooks.json,
+	// ~/.claude/settings.json, ~/.codex/config.toml) and re-installs
+	// the DefenseClaw hook block when a user removes it while the
+	// gateway is running. Tampered records the detection; Repaired
+	// records the successful re-install.
+	ActionConnectorHookTampered Action = "connector-hook-tampered"
+	ActionConnectorHookRepaired Action = "connector-hook-repaired"
+
 	// Codex notify webhook (agent-turn-complete et al.). The
 	// notify-bridge.sh shim installed by the codex connector POSTs
 	// codex's raw JSON arg to /api/v1/codex/notify after every
@@ -341,6 +351,8 @@ func AllActions() []Action {
 		ActionConnectorHook,
 		ActionConnectorHookSynthetic,
 		ActionAssetPolicy,
+		ActionConnectorHookTampered,
+		ActionConnectorHookRepaired,
 		ActionCodexNotify,
 		ActionCodexNotifyAgentTurnComplete,
 		ActionCodexNotifyMalformed,
--- a/internal/cli/embed/audit-event.json
+++ b/internal/cli/embed/audit-event.json
@@ -58,6 +58,8 @@
             "connector-hook",
             "connector-hook-synthetic",
             "asset-policy",
+            "connector-hook-tampered",
+            "connector-hook-repaired",
             "codex.notify",
             "codex.notify.agent-turn-complete",
             "codex.notify.malformed",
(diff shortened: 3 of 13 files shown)


rucpande (commented):
approach lgtm -haven't reviewed the code itself

PR #347, opened 2026-06-09
fix(scanner): route gateway MCP scans through the defenseclaw mcp scan CLI

ihabler:
## Summary

The Go gateway's MCP scanner subprocess has been broken since the initial commit — it invoked `mcp-scanner scan ...` but the `cisco-ai-mcp-scanner` CLI has never had a usable `scan` subcommand. Per @vineethsai's review, this PR **drops the standalone `mcp-scanner` binary entirely and routes through the SDK-backed Python CLI** (`defenseclaw mcp scan --json`) — the same pattern the plugin scanner already uses (`defenseclaw plugin scan --json`).

This replaces the earlier `config --config-path` attempt, which still couldn't work: `--output` is a file destination (not a JSON selector), the gateway passes a URL or server name (never a config file), and the SDK's per-item array never matched the Go parser.

## What changed

**Reroute to the Python CLI (`internal/scanner/mcp.go`)**
- `buildArgs` now emits `mcp scan --json [--analyzers <a>] [--scan-prompts] [--scan-resources] [--scan-instructions] <target>` — flags follow the `scan` subcommand, target is positional (matches click parsing).
- Coerce the empty default and the legacy `mcp-scanner` binary (incl. pathed variants) to `defenseclaw`, so existing configs self-heal on upgrade.
- Parse the `ScanResult.to_json()` shape (mirrors `parsePluginOutput`), dropping suppressed findings. The Python CLI resolves URL-vs-name targets (`_resolve_scan_target`, incl. local stdio servers) and LLM / Cisco AI Defense credentials from its own config, so the Go-side `MCP_SCANNER_*` env injection is removed (single SDK code path, no drift-prone second parser).

**Companion controls that must ship with a working dial path**

Before this PR, gateway-initiated MCP scans *silently failed* (no dial). The reroute is what finally makes them actually dial the target, so it carries:
- **SSRF guard** (`validateMCPScanTargetURL`): rejects loopback / private / link-local / cloud-metadata / credential-bearing URLs before dialing. Opt-out via `DEFENSECLAW_ALLOW_LOCAL_MCP_TARGETS=1` (local dev only). Required because neither the Python CLI nor the SDK validates the target, and `/v1/mcp/scan` accepts an arbitrary target from the request body.
- **Fail closed** on any non-zero scanner exit, even when stdout parsed cleanly.

**Tests**
- New unit tests: `buildArgs` (default + all knobs), binary coercion table, and `parseMCPOutput` against the `ScanResult.to_json()` schema (suppressed-drop / line / tags / empty).
- Updated the default-binary test (`defenseclaw` + legacy coercion) and removed the obsolete `scanEnv` tests.
- SSRF-guard tests included.

## Verification
- `go build ./...` clean; `go test ./internal/scanner/` passes; `go vet` clean; `gofmt` clean.
- Live: `defenseclaw mcp scan --json --analyzers yara <server>` → exit 0, clean `ScanResult.to_json()` on stdout, empty stderr; the Go parser deserializes it 1:1 (`{"scanner","target","timestamp","findings":[],"duration_ms"}`).

## Notes
- `cisco-ai-mcp-scanner` stays required — it provides the `mcpscanner` SDK that `defenseclaw mcp scan` imports — so `scripts/setup-scanners.sh` is unchanged; only the standalone `mcp-scanner` binary invocation is dropped.
- Operational: if a deployment sets `scanners.mcp.analyzers` to include `llm`, the LLM key must be present in `~/.defenseclaw/config.yaml` (or a standard provider env var) — same config source as before, now resolved by the Python CLI. The default `yara` analyzer needs no LLM.

### DIFF
--- a/internal/scanner/mcp.go
+++ b/internal/scanner/mcp.go
@@ -22,18 +22,112 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"net"
+	"net/url"
 	"os"
 	"os/exec"
+	"path/filepath"
+	"strings"
 	"time"
 
 	"github.com/defenseclaw/defenseclaw/internal/config"
 )
 
-// MCPScanner shells out to the Python “cisco-ai-mcp-scanner“ CLI.
-// Mirrors “SkillScanner“: the LLM-facing surface is driven by the
-// unified “config.LLMConfig“ resolved at “scanners.mcp“, with the
-// legacy “InspectLLMConfig“ kept only to preserve the old
-// “NewMCPScanner“ signature for existing callers/tests.
+// mcpScanTargetLooksLikeURL returns true when “target“ parses as a
+// remote URL (http/https/ws/wss/sse). Local paths and stdio targets
+// fall through unchanged so the scanner still accepts them.
+func mcpScanTargetLooksLikeURL(target string) bool {
+	if target == "" {
+		return false
+	}
+	lower := strings.ToLower(target)
+	for _, scheme := range []string{"http://", "https://", "ws://", "wss://", "sse://"} {
+		if strings.HasPrefix(lower, scheme) {
+			return true
+		}
+	}
+	return false
+}
+
+// validateMCPScanTargetURL refuses MCP scan URLs that point at
+// loopback / private / link-local / cloud metadata destinations, or
+// that embed inline credentials. The check is opt-out via
+// DEFENSECLAW_ALLOW_LOCAL_MCP_TARGETS=1 for local development only.
+func validateMCPScanTargetURL(target string) error {
+	if os.Getenv("DEFENSECLAW_ALLOW_LOCAL_MCP_TARGETS") == "1" {
+		return nil
+	}
+	u, err := url.Parse(target)
+	if err != nil {
+		return fmt.Errorf("invalid MCP scan target URL: %w", err)
+	}
+	if u.User != nil {
+		return errors.New("MCP scan target URL must not contain inline credentials")
+	}
+	host := u.Hostname()
+	if host == "" {
+		return errors.New("MCP scan target URL has empty host")
+	}
+	lowerHost := strings.ToLower(host)
+	if lowerHost == "localhost" ||
+		lowerHost == "ip6-localhost" ||
+		lowerHost == "ip6-loopback" ||
+		strings.HasSuffix(lowerHost, ".localhost") {
+		return fmt.Errorf("MCP scan target %q points at loopback host", host)
+	}
+	if lowerHost == "metadata.google.internal" {
+		return fmt.Errorf("MCP scan target %q points at cloud metadata host", host)
+	}
+	if literal := net.ParseIP(host); literal != nil {
+		if literal.IsLoopback() ||
+			literal.IsPrivate() ||
+			literal.IsLinkLocalUnicast() ||
+			literal.IsLinkLocalMulticast() ||
+			literal.IsUnspecified() ||
+			literal.IsMulticast() {
+			return fmt.Errorf("MCP scan target IP %s is loopback/private/link-local/multicast", literal)
+		}
+		// AWS / Oracle / DO IMDS literals.
+		if literal.String() == "169.254.169.254" || literal.String() == "fd00:ec2::254" {
+			return fmt.Errorf("MCP scan target %s is a cloud metadata endpoint", literal)
+		}
+		return nil
+	}
+	addrs, err := net.LookupIP(host)
+	if err != nil {
+		// DNS failure is treated as fatal here: an unresolvable host
+		// would be revalidated at dial time and we cannot rely on the
+		// downstream scanner to do that.
+		return fmt.Errorf("MCP scan target %q does not resolve: %w", host, err)
+	}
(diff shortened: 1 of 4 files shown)


vineethsai:
## PR Review, Need to fix

Good catch that `mcp-scanner scan` is invalid (valid subcommands are `static, remote, prompts, resources, instructions, behavioral, stdio, config, known-configs`). But the replacement has three issues that mean gateway-initiated MCP scans will still fail. I verified against the installed `cisco-ai-mcp-scanner` CLI and the actual gateway call paths.

### 1. `--output json` does not select JSON output
`--output`/`-o` is a **file destination** ("Save scan results to a file"), not a format selector. Tested offline (YARA only, no network):

```
mcp-scanner --output json --analyzers yara static --tools tools.json
```

stdout was the human-readable summary, not JSON, and no file named `json` was written:

```
=== MCP Scanner Results Summary ===
Scan Target: ...
Total tools scanned: 1
Safe items: 1
...
```

The Go scanner parses **stdout** as JSON (`cmd.Stdout` → `parseMCPOutput`), so this fails with `failed to parse mcp-scanner output`. To get JSON on stdout you need `--raw`/`-r`. (The original `--format json` was also invalid — `json` isn't one of `--format`'s choices: `raw,summary,detailed,by_tool,by_analyzer,by_severity,table`.)

### 2. `config --config-path <target>` is the wrong subcommand for what the gateway passes
The gateway never passes a config-file path. In the rescan path, `scanTargetFor` returns a **URL** (remote server) or a **server name** (local server):

```go
// internal/watcher/rescan.go
if entry.URL != "" {
    return entry.URL
}
return entry.Name
```

and `/v1/mcp/scan` passes an arbitrary `target` from the request body (typically a URL). So in practice this produces `config --config-path https://.../mcp` or `config --config-path github`, both of which fail because the argument isn't a config file.

### 3. Output schema mismatch (pre-existing, not addressed)
Even with the correct `--raw` flag, the real output is a top-level **array** of per-item results with `findings` keyed by analyzer:

```json
[
  {
    "status": "completed",
    "is_safe": true,
    "findings": { "yara_analyzer": { "severity": "SAFE", "total_findings": 0 } },
    "tool_name": "echo",
    "item_type": "tool"
  }
]
```

but `parseMCPOutput` expects a top-level object `{"findings":[{id,severity,title,...}]}`. So findings get dropped regardless of the subcommand fix.

### Tests / verification don't cover the change
`go test ./internal/scanner/ -run MCP` only exercises `scanEnv` and the default binary name — there's no test for `buildArgs` or output parsing, and the PR adds none. The manual command uses `config --config-path ~/.zeptoclaw/config.json`, which is a config file that exists locally but is not what the gateway passes. "Runs successfully" = exit 0, which hides that stdout contains no parseable JSON.

---

## Recommendation: drop the `mcp-scanner` CLI entirely and go through the SDK

There's already a working SDK-backed path: the Python CLI (`cli/defenseclaw/scanner/mcp.py`) uses the `mcpscanner` SDK directly and handles remote/local/prompts/resources/instructions plus correct schema mapping. The cleanest fix is to have the Go gateway shell out to `defenseclaw mcp scan <target> --json` and parse `ScanResult.to_json()` — exactly the pattern the plugin scanner already uses (`defenseclaw plugin scan --json`, see `internal/scanner/plugin.go`).

Why this is better:
- `defenseclaw mcp scan` already does URL-vs-name target resolution (`_resolve_scan_target`), so the gateway's URL and server-name targets work as-is — including local stdio servers, which the `config` subcommand can't handle from just a name.
- `--json` output is pure `ScanResult.to_json()`:
  ```json
  {"scanner","target","timestamp","findings":[{"id","severity","title","description","location","remediation","scanner","tags","rule_id?","line_number?"}],"duration_ms"}
  ```
  which Go can parse with the same struct shape as `parsePluginOutput`.
- Preserves the `--analyzers` / `--scan-prompts` / `--scan-resources` / `--scan-instructions` knobs (the Python `mcp scan` accepts all of them) instead of dropping them.
- One SDK code path — no second, drift-prone JSON parser to keep in sync.

Concretely:
- Default the MCP scanner binary to `defenseclaw` (and coerce a legacy `mcp-scanner` value, which can never work as `mcp-scanner mcp scan`).
- Build args: `mcp scan --json [--analyzers <a>] [--scan-prompts] [--scan-resources] [--scan-instructions] <target>`.
- Replace `parseMCPOutput` with a `ScanResult.to_json()`-shaped parser (mirror `parsePluginOutput`), and add `buildArgs` + parse unit tests.

Note that `cisco-ai-mcp-scanner` is still required — it provides the `mcpscanner` SDK module that `defenseclaw mcp scan` imports — so `scripts/setup-scanners.sh` stays as-is; we just stop invoking the standalone `mcp-scanner` binary.

vineethsai7:
Tracking the known SSRF-guard limitations (`validateMCPScanTargetURL`) as follow-up work in #350 — TOCTOU/DNS-rebinding, unfollowed redirects, process-global opt-out, and dedup with `webhook.go`. The guard in this PR is intentionally advisory (it runs in the Go parent while the dial/re-resolution happens in the Python subprocess); #350 covers making it a hard boundary.

PR #348, opened 2026-06-09
fix(setup): prompt to trust binary prefix + add Caskroom to defaults

ihabler:
## Summary

- When a connector binary resolves outside trusted prefixes (e.g. Homebrew Caskroom), interactive action-mode setup now **prompts the operator** to trust the directory instead of failing with a cryptic error — and after trusting, **re-runs the full version → hook-contract gate**, so an unsupported version is still refused
- Adds `/opt/homebrew/Caskroom` to the default trusted prefix list (same ownership/trust model as the already-trusted `/opt/homebrew/Cellar`)
- Adds first-class trusted-locations management: a `defenseclaw setup trusted-paths` CLI group and a native TUI panel/editor (review follow-up)

## Problem

Claude Code installed via `brew install --cask claude-code` resolves to `/opt/homebrew/Caskroom/claude-code/<version>/claude`. This path isn't in the default trusted prefix list, so `defenseclaw setup claude-code --mode action` fails with:

```
✗ Claude Code: version not available (binary path is not in a trusted install prefix)
  Refusing action-mode hook setup...
  Set DEFENSECLAW_ALLOW_HOOK_CONTRACT_DRIFT=1 only for exploratory testing.
```

The error gives no actionable remediation. (The Go gateway independently applies the same gate at boot and **skips** unverified action connectors — it does not crash; an earlier revision of this description overstated that.)

## Fix

**Interactive CLI:** prompts (default **No**, with a one-line warning of what trusting means) to add the resolved directory to `DEFENSECLAW_TRUSTED_BIN_PREFIXES`, persists to `~/.defenseclaw/.env` (0600), then re-enters the gate non-interactively so the contract check still applies and no re-prompt loop is possible.

**Non-interactive (CI):** prints the resolved path and points at `defenseclaw setup trusted-paths add <dir>`.

**TUI:** never prompts. A read-only "Trusted Paths" setup section plus a `TrustedPathsEditorScreen` list prefixes, highlight connectors resolving into untrusted directories, and dispatch add/remove through the standard command-preview pipeline. Untrusted-binary connector setups are routed into the editor instead of reaching `click.confirm`.

**CLI group:** `defenseclaw setup trusted-paths list|add|remove` with `--json` and `--force`; `add` refuses world-writable or non-absolute directories unless forced; `remove` never touches built-in defaults. The CLI, the inline prompt, and the discovery gate share the same helpers and a single `agent_discovery.validate_trusted_prefix` guard so they cannot drift.

**`agent discover`:** prints the same trusted-paths remediation hint for skipped untrusted binaries.

**Hardening from review:** shared `UNTRUSTED_PREFIX_ERROR` constant instead of a duplicated string literal; prefix lists join/split on `os.pathsep` so Windows drive-qualified paths survive; the prompt warns that version-pinned paths need re-trusting after upgrades.

## Test plan

- [x] `cli/tests/test_cmd_setup_trusted_paths.py` — gate re-run after trusting (incl. version-present-but-unsupported refusal), decline path, `_add_trusted_bin_prefix` append/dedupe/0600/environ, Caskroom default, trusted-paths CLI (list/add/remove, `--json`, world-writable refusal, default protection), `agent discover` hint
- [x] `cli/tests/tui/test_trusted_paths_editor.py` + `test_app_shell.py` — TUI editor add/remove/default-protection and untrusted-binary routing
- [x] Full TUI suite: 657 passed; `test_agent_discovery.py`: pass
- [x] Manual: `defenseclaw setup codex --mode action` on an untrusted install shows the default-No prompt, persists trust on `y`, re-runs the gate, and proceeds on a supported version; TUI panel/editor and routing verified interactively
- [ ] CI green

### DIFF
--- a/cli/defenseclaw/commands/cmd_agent.py
+++ b/cli/defenseclaw/commands/cmd_agent.py
@@ -40,6 +40,41 @@ def agent() -> None:
     """Inspect locally installed agent surfaces."""
 
 
+def _emit_untrusted_prefix_hint(disc: agent_discovery.AgentDiscovery) -> None:
+    """Point operators at the generic remediation when a connector binary
+    was skipped for living outside a trusted install prefix.
+
+    Without this, the only signal is the per-row "binary path is not in a
+    trusted install prefix" error — accurate but not actionable. The hint
+    is connector-agnostic and recommends the same ``trusted-paths`` command
+    the setup gate now points at, so discovery and setup stay consistent.
+    """
+    from defenseclaw import ux
+
+    dirs: list[str] = []
+    seen: set[str] = set()
+    for sig in disc.agents.values():
+        if sig.error == agent_discovery.UNTRUSTED_PREFIX_ERROR and sig.binary_path:
+            parent = os.path.dirname(os.path.realpath(sig.binary_path))
+            if parent and parent not in seen:
+                seen.add(parent)
+                dirs.append(parent)
+    if not dirs:
+        return
+    n = len(dirs)
+    click.echo()
+    ux.warn(
+        f"{n} director{'y' if n == 1 else 'ies'} hold connector binaries outside a "
+        "trusted install prefix; they were skipped during version discovery."
+    )
+    for d in dirs:
+        ux.subhead(f"  Trust it with: defenseclaw setup trusted-paths add {d}")
+    ux.subhead(
+        "  Only trust directories you control — a trusted directory lets "
+        "DefenseClaw execute any binary placed there during discovery."
+    )
+
+
 @agent.command("discover")
 @click.option("--refresh", is_flag=True, help="Refresh cached discovery before rendering.")
 @click.option("--no-cache", is_flag=True, help="Bypass the discovery cache for this run.")
@@ -75,6 +110,15 @@ def discover(
     gateway_token_env: str | None,
 ) -> None:
     """Run local agent discovery and optionally emit OTel telemetry."""
+    # `agent` is in main.SKIP_LOAD_COMMANDS, so the root group never loads
+    # config — and thus never hydrates ~/.defenseclaw/.env. Without this, a
+    # prefix persisted via `setup trusted-paths add` would be ignored here, so
+    # the untrusted-binary hint below would point at a fix this very command
+    # then disregards. Hydrate just the .env (cheap; no full config load or
+    # validation, keeping `agent` fast) so discovery honours persisted trust.
+    from defenseclaw.config import _load_dotenv_into_os, default_data_path  # noqa: PLC0415
+
+    _load_dotenv_into_os(str(default_data_path()))
     started = time.monotonic()
     disc = agent_discovery.discover_agents(use_cache=not no_cache, refresh=refresh)
     duration_ms = int((time.monotonic() - started) * 1000)
@@ -99,6 +143,7 @@ def discover(
         return
 
     click.echo(agent_discovery.render_discovery_table(disc).rstrip())
+    _emit_untrusted_prefix_hint(disc)
     if emit_otel:
         if otel_result["emitted"]:
             click.echo("  OTel: emitted agent discovery telemetry")
--- a/cli/defenseclaw/commands/cmd_setup.py
+++ b/cli/defenseclaw/commands/cmd_setup.py
@@ -26,6 +26,7 @@ import os
 import shutil
 import socket
 import subprocess
+import sys
 from pathlib import Path
 from typing import Any
 
@@ -1583,6 +1584,20 @@ def _write_dotenv(path: str, entries: dict[str, str]) -> None:
         pass
 
 
+def _add_trusted_bin_prefix(prefix: str, data_dir: str) -> None:
+    """Add a directory to DEFENSECLAW_TRUSTED_BIN_PREFIXES in .env and os.environ."""
+    dotenv_path = os.path.join(data_dir, ".env")
+    existing = _load_dotenv(dotenv_path)
+    current = existing.get("DEFENSECLAW_TRUSTED_BIN_PREFIXES", os.environ.get("DEFENSECLAW_TRUSTED_BIN_PREFIXES", ""))
+    parts = [p.strip() for p in current.split(":") if p.strip()]
+    if prefix not in parts:
+        parts.append(prefix)
+    new_val = ":".join(parts)
+    existing["DEFENSECLAW_TRUSTED_BIN_PREFIXES"] = new_val
+    _write_dotenv(dotenv_path, existing)
+    os.environ["DEFENSECLAW_TRUSTED_BIN_PREFIXES"] = new_val
+
+
 def _print_summary(sc, llm, aid) -> None:
     click.echo()
     click.echo("  Saved to ~/.defenseclaw/config.yaml")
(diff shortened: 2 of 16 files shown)


vineethsai (changes requested):
Thanks @ihabler — the direction is right: turn the cryptic action-mode failure into something actionable, and trust Homebrew Caskroom by default. The Caskroom default is reasonable — it has the same ownership/trust model as the already-trusted `/opt/homebrew/Cellar`, and `_is_trusted_binary_path` still enforces the world-writable-parent guard on top of it.

Before merge there's **one must-fix correctness/security bug (#1)** and **one must-fix default (#2)**. I'd also like us to land a first-class way to manage trusted locations — both an interactive TUI flow and a clean non-interactive CLI — instead of relying only on the inline confirm prompt. Details below, findings ordered by severity.

---

## Feature request: first-class "trusted locations" management (TUI + CLI)

Today the only ways to add a trusted prefix are (a) the new inline confirm prompt, (b) hand-editing `~/.defenseclaw/.env`, or (c) exporting `DEFENSECLAW_TRUSTED_BIN_PREFIXES`. None of these are discoverable or scriptable. Let's make this a real surface, and let the inline prompt become a thin caller of the same code path.

### Clean CLI (mirror the `defenseclaw keys` group UX in `cli/defenseclaw/commands/cmd_keys.py`)

```
defenseclaw setup trusted-paths list                 # defaults + operator-added, with source (default | .env | env) and a per-path ✓/✗
defenseclaw setup trusted-paths add <dir> [--force]  # validate, then persist to ~/.defenseclaw/.env
defenseclaw setup trusted-paths remove <dir>         # remove an operator-added entry (never a built-in default)
```

- Reuse the existing helpers (`_add_trusted_bin_prefix`, `_load_dotenv`, `_write_dotenv`) and the `agent_discovery._is_trusted_binary_path` validation so the CLI, the prompt, and the gate can never drift.
- `add` should **warn and refuse (unless `--force`)** when the directory is world-writable or not absolute — that is precisely what the allow-list defends against.
- Add `--json` output for automation/CI, consistent with the rest of the CLI.
- `remove` must only touch operator-added entries, never the built-in `_TRUSTED_BIN_PREFIXES_DEFAULT`.

### Interactive TUI (under `cli/defenseclaw/tui/panels/setup.py` / screens)

- A "Trusted binary locations" panel that lists current prefixes, highlights which connector binaries currently resolve into an *untrusted* directory, and lets the operator add/remove with a confirmation that surfaces the security implication.
- When setup detects an untrusted connector binary, route the **TUI** user into this panel instead of calling `click.confirm` (which can corrupt a full-screen TUI — see #5).

Net result: a clean non-interactive CLI for CI/automation, an interactive TUI for first-run, and the inline prompt reduced to a thin wrapper over the shared logic.

---

## Findings (ordered by severity)

### 1. [High — security/correctness] The prompt path bypasses the action-mode contract gate

The new success path returns `True` as soon as a non-empty version string comes back from re-discovery:

```python
if click.confirm(f"  Add '{parent}' to trusted binary prefixes?", default=True):
    _add_trusted_bin_prefix(parent, data_dir or os.path.expanduser("~/.defenseclaw"))
    disc2 = agent_discovery.discover_agents(use_cache=False, refresh=True, data_dir=data_dir)
    signal2 = disc2.agents.get(connector)
    if signal2 and signal2.version:
        ux.ok(f"{label}: version {signal2.version} verified after trusting {parent}.")
        return True
    ux.err("  Version still could not be verified after adding trusted prefix.")
```

Everywhere else, this function only returns `True` in action mode **after `resolve_connector_contract(...)` confirms the version maps to a known hook contract**. A version that is *present but unsupported* (too new/old) is deliberately refused — see `cli/defenseclaw/commands/cmd_setup.py` lines ~2562–2575:

```python
detail = (
    f"{label}: installed version {version_display} is not covered by a "
    f"DefenseClaw hook contract ({compatibility.reason})."
)
...
if action_mode and not allow_drift:
    ...
    return False
```

By short-circuiting on `signal2.version` truthiness alone, the prompt path lets an operator trust the prefix and proceed into action mode with a version the gate is specifically designed to reject. The success message also overstates it ("version … verified") — only the *path* was trusted; the contract was never re-checked.

**Fix:** after adding the prefix and re-discovering, re-run the full compatibility evaluation rather than checking `version` truthiness. Cleanest is to recurse with the prompt suppressed (which also avoids an infinite re-prompt loop when the parent is world-writable and trusting it can't help):

```python
if click.confirm(f"  Add '{parent}' to trusted binary prefixes?", default=False):
    _add_trusted_bin_prefix(parent, data_dir or os.path.expanduser("~/.defenseclaw"))
    # Re-run the full gate (now non-interactive) so the contract check still applies.
    return _check_connector_version_supported_for_setup(
        connector, mode=mode, emit=emit, data_dir=data_dir, _allow_prompt=False
    )
```

(Requires threading a small `_allow_prompt`/`interactive` parameter that defaults to today's behavior.)

### 2. [Medium — security] `default=True` on a trust-broadening prompt

`click.confirm(..., default=True)` means a bare Enter extends the binary-execution allow-list. That allow-list exists precisely to stop DefenseClaw exec'ing an attacker-planted binary picked up via `$PATH` (see the `M-4` rationale comment at the top of `cli/defenseclaw/inventory/agent_discovery.py`).

The world-writable guard (`parent_st.st_mode & 0o002`) is a strong mitigation, so the worst case (a world-writable drop dir) is still rejected. But a user-owned-but-unexpected location would be one keystroke from being permanently trusted. For a security-relevant decision, **default to `False`** and add a one-line warning describing what trusting the directory means. This also aligns with not making the risky option the path of least resistance.

### 3. [Medium — robustness] Fragile exact-string coupling

```python
is_untrusted_path = probe_error == "binary path is not in a trusted install prefix" and signal and signal.binary_path
```

This literal is duplicated from another module — `agent_discovery._version_for_binary` returns the same string when `_is_trusted_binary_path` fails. If that wording ever changes, the prompt silently stops firing and users are back to the cryptic error, with no test to catch it. **Extract a shared constant** (e.g. `agent_discovery.UNTRUSTED_PREFIX_ERROR`) and reference it from both sides.

(Minor: `is_untrusted_path` evaluates to the `binary_path` string rather than a bool. It works via truthiness, but `bool(...)` reads more clearly.)

### 4. [Medium — UX/robustness] Trusting the version-specific parent dir is brittle

`parent = os.path.dirname(os.path.realpath(signal.binary_path))`. For a Caskroom install that resolves to `/opt/homebrew/Caskroom/<app>/<version>/...`, so the persisted prefix is version-pinned. On the next upgrade the version directory changes and trust breaks again, re-triggering the prompt. The Caskroom case is now covered by the new default, but for **any other** untrusted location the prompt hits, persisting a version-scoped path means recurring prompts. Since the prefix match is a true path-boundary prefix (`resolved.startswith(prefix + os.sep)`), trusting a higher-level directory would be durable. Consider offering the higher-level prefix (or at least document that re-trust is needed after upgrades). The proposed `trusted-paths` UX above is a good place to make this an explicit operator choice.

### 5. [Medium] Interactivity is gated only on `isatty()` — TUI re-entrancy risk

```python
interactive = sys.stdin.isatty() and sys.stdout.isatty()
```

The PR description says the TUI path is treated as non-interactive, but nothing here distinguishes "real terminal" from "terminal owned by a live TUI." If the Go TUI ever drives `defenseclaw setup mode <connector>` / `setup guardrail` over a PTY, `click.confirm` will fire mid-render and corrupt the display (or block on input the TUI can't provide). The robust gate is an explicit flag threaded from the caller — the multi-connector path already passes `emit=False`, so the precedent exists. At minimum, confirm the TUI never reaches the call sites (`cmd_setup.py` ~3118 / ~3480 / ~4771 / ~4828) with a TTY attached. The TUI panel proposed above removes the need to prompt from this function at all in the TUI case.

### 6. [Low] Non-interactive hint is circular and uses the internal connector name

```python
ux.subhead(f"  To fix, run: defenseclaw setup {connector} --mode action")
```

In the non-interactive branch this suggests re-running the *same* command that just failed (still non-interactive → same outcome). Also, `connector` is the normalized internal id (e.g. `claudecode`), which isn't necessarily the user-facing CLI form (`claude-code` / `setup mode claudecode`). The genuinely useful instruction is the next line (manually add to `.env`) — or, better, point at the new `defenseclaw setup trusted-paths add <dir>` command.

### 7. [Low] Missing tests

The test plan cites `cli/tests/test_agent_discovery.py` (16/16), but that suite doesn't exercise any of the new code — it all lives in `cmd_setup.py`. There are no tests for `_add_trusted_bin_prefix`, the confirm/decline branches, the re-discovery success/failure paths, or the new `/opt/homebrew/Caskroom` default. `test_agent_discovery.py` already monkeypatches `_TRUSTED_BIN_PREFIXES_DEFAULT`, so please add at least:

- `_add_trusted_bin_prefix` appends-and-dedupes, persists to `.env` at `0600`, and updates `os.environ`;
- the prompt path returns the gate's verdict (covering #1) using a mocked `discover_agents` — including the "version present but unsupported" case;
- `"/opt/homebrew/Caskroom" in agent_discovery._TRUSTED_BIN_PREFIXES_DEFAULT`;
- tests for the new `trusted-paths` CLI (list/add/remove, world-writable refusal, `--json`).

### 8. [Low] Windows path separator

`_add_trusted_bin_prefix` joins with `:` and `_trusted_bin_prefixes()` splits on `:`. On Windows, `os.path.dirname(os.path.realpath(...))` yields `C:\...`, which a `:`-split mangles. This is a pre-existing limitation of the mechanism (all defaults are POSIX), but the new prompt increases the chance of it firing on Windows. Worth a guard or at least a documented known-limitation, especially given the recent Windows-installer work on `main`.

### 9. [Doc accuracy] "Crash for ALL connectors at gateway boot" isn't substantiated by the diff

The PR is CLI-only (no Go changes), yet the description claims a gateway-boot crash. There's no trusted-prefix concept in the Go gateway, and the in-repo comment in `cli/defenseclaw/commands/cmd_init.py` (~line 831) states the gateway *skips* unverified action connectors rather than crashing:

> The Go gateway applies the same gate at boot (skipping unverified action connectors) …

Either the crash is a real, separate symptom (please link the trace and explain how a CLI-only change fixes it), or the description overstates it. Per the repo's PR-description standard, this should be precise.

---

## Suggested priority / checklist

- [ ] #1 — gate bypass (must-fix)
- [ ] #2 — default-to-No + security warning (must-fix)
- [ ] New — `defenseclaw setup trusted-paths` CLI group (list/add/remove/`--json`) + interactive TUI panel
- [ ] #3 — shared `UNTRUSTED_PREFIX_ERROR` constant
- [ ] #7 — tests covering #1, the new CLI, and the Caskroom default
- [ ] #4 / #5 / #6 / #8 — polish + cross-platform/TUI hardening
- [ ] #9 — tighten the PR description

PR #352, opened 2026-06-10
Add Microsoft Scout surface connector support

barryqy:
## Summary
- add Microsoft Scout as a surface-only connector for local skill discovery
- keep Scout guardrail setup blocked until Microsoft publishes a hook or proxy contract we can safely integrate
- refresh schemas, connector metadata, docs, and parity tests against current main
- address review feedback around surface-only resolver scope, Scout MCP manifest rejection, OpenCode hook-only routing, dashboard options, and Scout read-only bundled skill handling

## Validation
- `uv run pytest cli/tests/test_multi_connector_config.py cli/tests/test_cmd_skill.py cli/tests/test_codeguard_opt_in.py cli/tests/test_registry_manifest.py cli/tests/test_check_schemas.py cli/tests/test_ai_signatures.py cli/tests/test_agent_discovery.py -q`
- `uv run ruff check cli/defenseclaw/commands/__init__.py cli/defenseclaw/commands/cmd_codeguard.py cli/defenseclaw/commands/cmd_setup.py cli/defenseclaw/commands/cmd_skill.py cli/defenseclaw/registries/manifest.py cli/tests/test_cmd_skill.py cli/tests/test_multi_connector_config.py cli/tests/test_registry_manifest.py`
- `make check-schemas`
- `make sync-openclaw-extension; go test ./internal/config ./internal/daemon ./internal/gateway ./internal/gateway/connector ./internal/inventory`
- `npm run build` in `docs-site`
- `git diff --check`


vineethsai7 (commented):
Deep review notes from the isolated worktree. I focused on Scout discovery semantics, inventory signal quality, and whether the documented Scout skill tiers are covered.

internal/inventory/ai_signatures.json
```diff
@@ -389,6 +389,45 @@
...
+      ],
+      "process_names": [
+        "Scout",
+        "Microsoft Scout",
+        "Clawpilot"
+      ],
+      "config_paths": [
+        "~/.copilot/skills",
+        "~/.copilot/m-skills",
+        "~/.copilot"
```
vineethsai7: [P2] This config-path signature will produce Scout false positives on hosts that only have GitHub Copilot state. Both `~/.copilot/skills` and especially bare `~/.copilot` are shared/ambiguous; the CLI mirror also marks Scout installed for any existing config candidate. I reproduced an empty `~/.copilot` directory returning `installed=True` with no Scout binary. Consider requiring a Scout-specific binary/app/process signal, removing the bare root, or downgrading shared paths to weak corroborating evidence instead of a primary `supported_connector` signal.

internal/inventory/ai_signatures.json
```diff
@@ -389,6 +389,45 @@
...
+      ],
+      "extension_ids": [],
+      "mcp_paths": [],
+      "package_names": [],
+      "env_var_names": [],
+      "domain_patterns": [
+        "microsoft.com",
```
vineethsai7: [P2] `microsoft.com` and `microsoft365.com` are too broad for Scout identity evidence. Domain matches from shell history are emitted without the weak `Quality=0.5` treatment used for history substrings, so ordinary Microsoft URLs can become high-confidence Scout inventory signals under this `confidence: 0.92` signature. Please use Scout-specific domains/commands if available, or remove/downgrade these patterns.

internal/gateway/connector/scout.go
```diff
@@ -0,0 +1,153 @@
...
+	return map[string][]string{
+		"skill": scoutSkillPaths(SetupOpts{}),
+	}
+}
+
+func scoutSkillPaths(opts SetupOpts) []string {
+	_ = opts
+	return uniqueNonEmptyStrings([]string{
+		homePath(".copilot", "skills"),
+		homePath(".copilot", "m-skills"),
```
vineethsai7, line 151: [P3] This omits Scout bundled skills. Microsoft documents three active skill tiers: bundled `~/.copilot/bundled-skills/`, workspace `~/.copilot/m-skills/`, and global user `~/.copilot/skills/` (see https://learn.microsoft.com/en-us/microsoft-scout/use-microsoft-scout and https://learn.microsoft.com/en-us/microsoft-scout/faq). Since `skill list/scan --all --connector scout` and sidecar component targets use this list, bundled Scout skills will be invisible unless the omission is intentional and documented.

vineethsai7 (commented):
Follow-up review after commit 61952c7c. The previous Scout identity false positives and missing bundled-skill scan coverage look addressed; I found one remaining mutation-safety issue around treating bundled skills as a normal skill root.

cli/defenseclaw/connector_paths.py
```diff
@@ -564,6 +581,17 @@ def _copilot_skill_dirs(workspace_dir: str | None = None) -> list[str]:
     )
 
 
+def _scout_skill_dirs() -> list[str]:
+    home = str(Path.home())
+    return _dedup(
+        [
+            os.path.join(home, ".copilot", "bundled-skills"),
```
vineethsai7, line 687: [P2] `bundled-skills` should not be in the same mutable path set used by skill quarantine and `skill scan --action`. Microsoft documents this tier as shipped with the app and not user-editable, but `connector_paths.skill_dirs("scout")` feeds the generic skill commands. I reproduced `defenseclaw skill quarantine word --connector scout` moving `~/.copilot/bundled-skills/word` into DefenseClaw quarantine (`bundled_exists False`). `_apply_scan_enforcement()` would use the same path on `skill scan --all --connector scout --action` when policy chooses `file=quarantine`. Please separate read-only scan/list roots from mutable/enforcement roots, or explicitly reject quarantine/restore/scan-action for paths under `~/.copilot/bundled-skills` while still allowing read-only scan/list.

barryqy:
One note on the Scout discovery comments: I left Scout config-path discovery empty on purpose. Earlier review found that shared ~/.copilot skill directories create false positives on Copilot-only machines, and the branch now has tests pinning that behavior in cli/tests/test_ai_signatures.py, cli/tests/test_agent_discovery.py, and internal/inventory/ai_discovery_test.go. The Scout skill directories remain scan/list surfaces, but they are not install identity evidence by themselves.

PR #362, opened 2026-06-14
fix(correlator): wire data-axis + tool-capability labeling end-to-end

ihabler:
The sliding-window correlator could never fire its capability-aware or judge-driven patterns because findings were persisted without data_axis / tool_capability_class labels:

- AxesForJudgeCategory and ClassifyToolName were only exercised in tests, leaving DESTRUCTIVE-FLOW inert and judge findings unlabeled.
- Several real producer rule prefixes (CS-PII-*, JSON-SEC-*, COG-*, JUDGE-*) had no axis mapping.

Changes:
- Add AxesForFinding unified entrypoint (rule id -> judge category -> tags) and wire it into the finding enricher.
- Add CapabilityForRuleID + SetCapabilityEnricher so content-matched CMD-*/SRC-* findings get a tool_capability_class; stamp capability from the tool name in proxy tool-call inspection.
- Extend axis prefix coverage and sync DumpRuleAxesSnapshot + rule-axes.json.
- Fix misleading axes_test assertions to use real producer rule ids and add producer-coverage tests.


ihabler:
Also added oredering patterns. Ordering is keyed on the DB timestamp column (and not turn_id which was missing from several connectors)

vineethsai7 (approved):

PR #372, opened 2026-06-22
[feat] unified enterprise hardening with config.yaml source of truth

vineethsai7:
> **Unified scope note:** this PR now directly contains the work previously reviewed in [#373](https://github.com/cisco-ai-defense/defenseclaw/pull/373) and [#379](https://github.com/cisco-ai-defense/defenseclaw/pull/379). Review it as one enterprise-hardening change against `main`, not as a remaining stack: `config.yaml` convergence, automatic application protection, and managed-enterprise enforcement all ship together. The final reviewed and remediated range is `94dd46c689fefbcf85b8f478249e74f3925eca49..b1cc27512`.

## Problem

DefenseClaw's runtime and enterprise controls were split across several partially independent mechanisms:

- Guardrail startup configuration came from `config.yaml`, while live guardrail overrides could come from `${data_dir}/guardrail_runtime.json`.
- The guardrail proxy had its own polling/reload path instead of one validated gateway configuration lifecycle.
- AI discovery could identify supported applications, but hook-native protection still required manual onboarding and automatic connectors could inherit overly broad global action posture.
- Enterprise deployments needed a durable administrator boundary around configuration, hook authorization, per-user hook files, and connector credentials across Linux, macOS, and Windows.
- Existing CLI commands, the local API, background reload, root-assisted guardian workflows, and packaging all needed to agree on the same effective policy and managed-mode authorization model.

Without a unified design, disk state, status output, and live enforcement could diverge; a local user or lower-privilege service could also gain influence through unsafe ownership, ACL, link, token, or migration paths.

## Current Situation

Before this unified change:

- `/v1/guardrail/config` wrote a secondary runtime JSON file, and the proxy reloaded it independently.
- Most gateway subsystems were startup-bound; there was no central validated snapshot/diff/reconciliation manager.
- Upgrade and startup behavior did not consistently fold legacy guardrail state into the canonical YAML configuration.
- Supported hook-native applications were discovered but not policy-driven through setup, verification, health, and status.
- Automatic connectors could inherit global action settings rather than requiring an explicit automatic-protection opt-in.
- Managed deployments did not have one cross-platform trust model for root-owned config, hook targets, scoped tokens, symlinks/reparse points, ACLs/DACLs, repair authorization, or truthful readiness.
- Root-assisted enterprise hook installation and repair were documentation-heavy rather than a first-class CLI and packaged guardian workflow.

This PR supplies that integrated architecture. The final security review found three P2 gaps in the original reviewed head; all three were remediated in `b1cc27512` and revalidated through their original runtime or realistic interface paths.

### 1. Make `config.yaml` the runtime source of truth

- Remove `guardrail_runtime.json` as a live configuration source.
- Patch `/v1/guardrail/config` changes into `config.yaml` with validation and rollback.
- Add Python upgrade migration and a Go startup compatibility migration for supported legacy fields.
- Keep normal reloads YAML-only so a deleted legacy file cannot reappear as an active overlay.
- Preserve YAML comments/order for dotted scalar updates.

Managed/root startup now refuses to consume service-owned legacy state; only unmanaged compatibility migration remains automatic, while managed migration requires an explicit administrator-controlled configuration change.

### 2. Centralize validation, snapshots, and reconciliation

- Watch and debounce `config.yaml` changes in one `ConfigManager`.
- Validate the complete candidate config and retain the last-good immutable snapshot on failure.
- Diff changed sections and reconcile OTel, audit sinks, webhooks, notifications, discovery, API, watcher, proxy, and related controllers.
- Support explicit hot/restart behavior and reject storage identity changes that cannot be safely applied in process.
- Audit successful reloads and publish reload health.

The mutation path now enforces that transactional invariant: eligible guardrail changes synchronously reach the shared snapshot and live proxy before HTTP 200, restart-required hot changes return 409 without changing disk, and explicit restart mode returns 202 with `live:false`.

### 3. Add policy-driven automatic application protection (#373)

- Add `application_protection` configuration with global enablement, confidence thresholds, include/exclude filters, disappearance policy, and per-connector overrides.
- Keep automatic connectors in `observe` by default. Automatic `action` or asset-policy enforcement requires explicit `application_protection` opt-in and does not inherit the broad global action posture.
- Feed `supported_connector` discovery signals into a gateway controller that filters to hook-native connectors, preflights action-mode hook safety, and reuses existing setup/verify/contract-lock paths.
- Persist controller state and expose connector source, skipped reasons, errors, discovered apps, and effective automatic policy through the existing `defenseclaw status` and `status --json` commands.
- Keep proxy connectors setup-only; this change intentionally adds no new Python `agent protection` command family.
- Move newly trusted binary prefixes into `config.yaml`; legacy environment entries remain readable for compatibility when strict trusted-path checking is enabled.

### 4. Establish the managed-enterprise trust boundary (#379)

- Pin `DEFENSECLAW_DEPLOYMENT_MODE=managed_enterprise` in packaged services and reject non-administrator config mutations, including authenticated API downgrade attempts.
- Add root-assisted `defenseclaw-gateway enterprise hooks install|reconcile|watch` workflows backed by a strict target manifest and authorization ledger.
- Generate canonical per-user hooks and connector-scoped `.hook-<connector>.token` sidecars; keep the older shared `.token` only as a compatibility fallback.
- Validate protected path ancestry, ownership, modes, regular-file/link status, Linux ACLs, macOS ACL intent, and Windows DACL/reparse semantics before trusting config or credentials.
- Make first-time installation fail closed; after explicit root authorization, repair only the declared target-owned hook footprint and preserve unchanged metadata to avoid self-triggering watcher loops.
- Preflight every configured connector token before activation, roll back partial startup, and keep managed readiness non-enforcing until all declared targets are authorized and healthy.
- Bound Linux guardian privileges to the documented UID/GID/chown capability set and keep the gateway service itself at `NoNewPrivileges=true` with an empty capability set.
- Ship executable systemd/launchd packaging, operator architecture/runbooks, destructive lab validation, incident recovery, and privileged Ubuntu/macOS connector matrix coverage.

The macOS runtime validator and launchd installer now inspect every relevant path element for write-capable extended ACL entries and fail closed before trusting or mutating the managed layout.

### 5. Preserve and clarify CLI behavior

- **New gateway commands:** `enterprise hooks install`, `enterprise hooks reconcile`, and `enterprise hooks watch`.
- **Existing Python commands remain the operator surface:** setup, status, doctor, config, settings, migrations, observability, guardrail, connector, and uninstall/upgrade flows remain available.
- **Automatic protection uses existing status UX:** no new Python command family is introduced by #373.
- **Managed-mode writes are authorization-gated:** standard users may inspect permitted status but cannot mutate root-managed policy; the local API rejects managed configuration mutation.
- **Restart semantics are explicit:** commands that change constructor-time or security-critical state must report/require restart rather than imply a partial live apply.
- **Sandbox parsing is unambiguous:** child-command flags follow `--`, and `sandbox exec --help` remains side-effect free.

The complete current-head command trees were recursively exercised during the final review: 212 Python CLI paths and 40 gateway CLI paths rendered help successfully. The three new enterprise-hook subcommands also rendered help on the authorized Linux ARM64 host.

### 6. Package and verify the integrated posture

- Align Python/Go config models, defaults, validation, migrations, status, and setup persistence.
- Add native Linux, macOS, and Windows path/token controls and platform build/test coverage.
- Add launchd/systemd installers and guardian services, release archive entries, CI hardening harnesses, and enterprise runbooks.
- Keep exact-head unit, race, CLI, migration, packaging, syntax, dependency, and live-host validation in the test plan below.

```mermaid
flowchart TD
    Admin["Enterprise administrator"] --> Config["Root-owned config.yaml\ncanonical policy"]
    Legacy["Legacy guardrail_runtime.json\none-time compatibility input"] --> Migration["Explicit trusted migration"]
    Migration --> Config

    Config --> Manager["ConfigManager\nvalidate + last-good snapshot + diff"]
    Manager --> Runtime["Gateway sidecar controllers"]
    Runtime --> Proxy["Guardrail proxy\nlive enforcement"]
    Runtime --> Discovery["AI discovery"]
    Discovery --> Auto["Application-protection policy"]
    Auto --> Setup["Existing connector setup + verify"]

    Admin --> Manifest["Root-owned guardian manifest\nand authorization ledger"]
    Manifest --> Guardian["enterprise hooks\ninstall / reconcile / watch"]
    Guardian --> Hooks["Canonical per-user hooks"]
    Guardian --> Tokens["Connector-scoped protected tokens"]
    Hooks --> HookAPI["Loopback authenticated hook API"]
    Tokens --> HookAPI
    HookAPI --> Proxy

    Status["Existing status / status --json"] --> Runtime
    Status --> Auto
    Status --> Guardian
```

## What Changed (Where & How)

| Area | Paths | Change |
| --- | --- | --- |
| Canonical configuration | `internal/config/config.go`, `guardrail_runtime_migration.go`, `yaml_patch.go`, `cli/defenseclaw/{config.py,migrations.py}` | Makes `config.yaml` authoritative, adds legacy migration, YAML patching, validation, defaults, and Python/Go parity. |
| Dynamic runtime config | `internal/gateway/{config_manager.go,sidecar.go,api.go,proxy.go,otel_runtime.go}` | Adds watch/diff/last-good snapshots, API YAML persistence, live controller reconciliation, and restart-required classification. |
| Runtime providers and sinks | `internal/{telemetry,audit}/...` | Builds inactive OTel providers, swaps sink/provider state safely, and keeps runtime consumers bound to the current snapshot. |
| Automatic application protection (#373) | `internal/config/application_protection.go`, `internal/gateway/application_protection.go`, `internal/inventory/ai_discovery.go` | Adds config-driven hook-native activation, confidence/filter policy, source-aware authorization, persisted controller state, and setup preflight. |
| Discovery/status/trusted paths (#373) | `cli/defenseclaw/inventory/agent_discovery.py`, `commands/{cmd_setup.py,cmd_status.py}`, TUI trusted-path editor | Makes binary trust config-backed and extends existing status text/JSON with automatic-protection state and connector source. |
| Enterprise hook CLI (#379) | `internal/cli/enterprise_hooks.go`, `internal/enterprisehooks/...` | Adds install/reconcile/watch, strict manifest handling, authorization, event-driven repair, and no-op-stable reconciliation. |
| Scoped hook credentials (#379) | `internal/gateway/connector/hook_api_*`, `hookexec/`, hook templates | Adds per-connector raw token sidecars, protected-state revalidation, platform ownership/ACL/DACL checks, and legacy fallback precedence. |
| Managed runtime semantics (#379) | `internal/managed/...`, `internal/gateway/sidecar.go`, `internal/cli/status.go` | Pins managed mode, rejects unauthorized writes, preflights all connectors, publishes truthful health, and protects runtime/config paths. |
| Privileged setup and sandbox (#379) | `cli/defenseclaw/commands/{cmd_init_sandbox.py,cmd_setup.py}`, `internal/cli/sandbox.go` | Validates privileged helpers and targets, makes config rollback transactional, waits for verified restart state, and clarifies `--` parsing. |
| Packaging and services (#379) | `packaging/systemd/...`, `packaging/launchd/...`, `.goreleaser.yaml` | Adds bounded guardian services, managed ownership/mode layout, shipped launchd installer, and release artifacts. |
| Docs and CI (#379) | `docs-site/content/docs/setup/...`, `.github/workflows/ci.yml`, `scripts/test-enterprise-hook-hardening.sh`, `scripts/test-macos-enterprise-packaging.sh` | Adds architecture, provisioning, operations, incident/tamper runbooks, four-cell privileged hook CI, and native macOS packaging validation. |
| Security remediation | `internal/config/config.go`, `internal/gateway/{api.go,config_manager.go,sidecar.go}`, `internal/managed/trust_acl_*`, `packaging/launchd/install-enterprise.sh` | Closes privileged legacy-state promotion, disk/live guardrail divergence, and macOS extended-ACL trust gaps with rollback and exploit-focused tests. |
| Regression coverage | `internal/**/*_test.go`, `cli/tests/**`, workflow/script tests | Covers migration, reload, auto-protection, managed auth, hook repair, scoped tokens, platform trust, packaging, status, and legacy CLI behavior. |

## Breaking Changes

- `guardrail_runtime.json` is no longer a live configuration source. Supported values are migrated once and the file is removed; scripts that continue editing it have no runtime effect.
- `/v1/guardrail/config` now mutates and validates `config.yaml` rather than writing a sidecar overlay.
- Failed startup migration or invalid config reload fails closed instead of running with ambiguous split-brain policy.
- Security-critical and storage-identity changes may require process restart rather than partial hot application.
- Automatic connectors do not inherit global `guardrail.mode: action` or global asset-policy action; automatic enforcement requires explicit `application_protection` policy.
- `ai_discovery.require_trusted_binary_paths` defaults to `false`; when enabled, config-backed prefixes and legacy environment prefixes are both honored.
- Managed packages pin `managed_enterprise`; local config and authenticated non-admin API calls cannot downgrade that deployment mode.
- Managed per-user credentials use `.hook-<connector>.token`; the shared `.token` remains compatibility-only.
- Root-authorized guardian reconciliation may replace target-owned symlinks and normalize unsafe target-owned modes inside the declared hook footprint; foreign-owned objects remain hard failures.
- Linux guardian units set `NoNewPrivileges=false` only for the bounded privilege-regain workflow; the gateway keeps `NoNewPrivileges=true`.
- `defenseclaw-gateway sandbox exec` requires command-specific flags after `--`.
- macOS managed installation requires the shipped installer or equivalent MDM policy to establish and verify the service identity and protected filesystem layout.

## Security Review Status

The final Codex Security diff review covered the unified #372 scope, including all work merged from #373 and #379. The original reviewed head produced three high-confidence **Medium / P2** findings. Commit `b1cc27512` fixes all three, adds exploit-focused regression coverage, and was validated against the original broken controls.

| Finding | Remediation | Closure evidence |
| --- | --- | --- |
| Root guardian migrated service-owned legacy state into administrator config | Automatic legacy overlay migration is disabled whenever the authoritative config or immutable environment selects `managed_enterprise`; the overlay is preserved for an explicit administrator-controlled change. | A real root `enterprise hooks reconcile` run on `openclaw-vineeth` left root-owned `config.yaml` byte-for-byte unchanged and preserved the service-owned overlay. |
| Guardrail config could report success while live enforcement remained stale | The API now serializes writes, validates the complete candidate, synchronously reloads through `ConfigManager`, verifies the live snapshot, and rolls disk/runtime back on failure. Restart-required hot changes return 409; restart mode reports 202 and `live:false`. | The authenticated API-to-manager-to-sidecar-to-proxy integration test verifies disk, manager, sidecar, proxy, and response all reach `action` together. Race and rollback tests pass. |
| macOS managed-path trust ignored write-capable extended ACLs | Darwin runtime validation, CLI diagnostics, and the launchd installer reject write/create/delete/chown-like allow ACL entries on every relevant path element before trust or mutation. | Native macOS tests create real extended ACLs and prove config/runtime rejection; the installer function accepts an ACL-free path and rejects a write ACL. |

A nearby TOCTOU path was also closed: guardrail PATCH re-snapshots and re-authorizes after entering its write transaction, so an administrator transition to managed mode or token rotation cannot be overwritten by an already-parsed request.

**Merge readiness:** the three security merge blockers are resolved in the pushed PR head. Normal CI and reviewer approval still apply.

The readable scan and remediation evidence remain attached to Codex Security scan `ae173d1f-28ca-49fa-80c9-76cc788a7aa4`.

## Open Issues / Follow-ups

None.

## Test Plan

All commands below were run against the complete remediated working tree immediately before commit `b1cc27512`.

### Complete Go validation

```bash
go test ./...
go vet ./...
```

Observed result: every Go package passed, including gateway, connector, enterprise hooks, managed trust, unit, and end-to-end packages; vet passed with no output.

```bash
go test -race ./internal/gateway ./internal/gateway/connector ./internal/enterprisehooks -count=1
```

Observed result: passed (`internal/gateway` 403.104s, connector 10.394s, enterprise hooks 2.935s).

The final post-lock authorization hardening was also race-tested:

```bash
go test -race ./internal/gateway -run 'GuardrailConfig|GuardrailAPIPatch|ConfigManager' -count=1
```

Observed result: passed.

### Complete Python CLI validation

```bash
make _bundle-data
uv run pytest -q cli/tests
```

Observed result: `4679 passed, 3 skipped, 11 warnings, 335 subtests passed in 443.01s`.

```bash
uv run ruff check \
  cli/defenseclaw/config.py \
  cli/defenseclaw/migrations.py \
  cli/defenseclaw/commands/cmd_setup_observability.py \
  cli/tests/test_config.py \
  cli/tests/test_migrations.py \
  cli/tests/test_enterprise_packaging.py
python3 -m compileall -q cli/defenseclaw cli/tests
```

Observed result: passed.

### Platform and packaging validation

```bash
GOOS=windows GOARCH=amd64 go test -c ./internal/managed
GOOS=linux GOARCH=arm64 go test -c ./internal/managed
go test -v ./internal/managed -run 'DarwinACL' -count=1
git diff --check
git diff --name-only -- '*.sh' | xargs -n 1 bash -n
```

Observed result: Windows/amd64 and Linux/arm64 managed-trust builds compiled; native Darwin ACL tests passed; whitespace and every changed shell script passed syntax validation.

The launchd ACL helper was directly exercised with clean and write-ACL fixtures. Observed result: `installer_acl_clean=accepted installer_acl_write=rejected`.

### Original security reproductions after remediation

- **Linux root-guardian migration:** built the current Linux ARM64 gateway and copied it to the authorized `openclaw-vineeth` host. A disposable managed layout used root-owned mode-`action` YAML and a service-owned mode-`observe` legacy overlay. Real root `enterprise hooks reconcile` returned successfully with `config_unchanged=true`, `config_meta=root:defenseclaw 640`, `runtime_preserved=true`, and `runtime_meta=defenseclaw:defenseclaw 600`. The fixture and binary were removed.
- **Guardrail live-state split:** the real authenticated PATCH handler, YAML loader, `ConfigManager.Reload`, `Sidecar.applyConfigReload`, and proxy now complete as one tested transaction. HTTP 200 is emitted only after all layers contain `action`; restart-required hot mutation leaves the original YAML bytes and live scanner mode unchanged.
- **macOS ACL trust:** real write-capable ACLs on a config file and managed runtime directory are rejected; an ACL-free path remains accepted. The installer rejects the unsafe ACL before mutation and preserves the existing config hash.

### Dependency and static review

The exact-head security review also ran:

```bash
go mod verify
govulncheck ./...
```

Observed result: all modules verified and `govulncheck` reported `No vulnerabilities found`. Broad Gosec, Bandit, and Ruff results were triaged; no additional reportable finding survived validation.


vineethsai7:
Implemented the PR comment fixes, rebased onto latest `origin/main`, committed, and force-pushed the branch.

What changed:
- `openshell` is now included in config diffs, so `OpenShell`-only edits no longer get ignored.
- `llm` changes now restart LLM consumers that cache resolved LLM settings.
- `guardrail.judge` reload now rebuilds the shared judge and reapplies it to the API server and event router, not just the proxy path.
- Added focused regression tests for all three gaps.

PATCH behavior:
- `/v1/guardrail/config` does **not** accept raw `config.yaml`.
- It accepts a JSON partial body for supported guardrail fields: `mode`, `scanner_mode`, `block_message`, `connector`, `hilt_enabled`, `hilt_min_severity`.
- The handler patches `config.yaml`, validates the full config, updates the API server’s live guardrail view, then the config watcher/reconciler reloads and applies the latest config.
- It does **not** fully restart everything. It hot-applies simple settings and restarts only affected in-process components. Restart-required storage identity paths remain rejected.

Hot reload flow:
1. Admin edits `config.yaml`, or PATCH writes supported guardrail keys into `config.yaml`.
2. YAML patch preserves comments/order and validates the full file.
3. Watcher detects the change and debounces events.
4. Config manager reloads with `config.LoadFromFile`.
5. Reconciler diffs old vs new config.
6. Invalid or restart-required changes are rejected and last-good config stays active.
7. Valid changes are prepared, committed, and applied to affected subsystems.
8. Proxy/API/router/judge/watcher/OTel/audit/AI discovery are hot-applied or restarted only as needed.

DefenseClaw knows through a filesystem watcher, not by polling.

At sidecar startup it resolves the active config path, defaulting to `config.yaml`, then starts a `ConfigManager` goroutine:

- Default config file is `config.yaml`: [internal/config/defaults.go](/Users/vnarajal/Desktop/defenseclaw-1/internal/config/defaults.go:66)
- Sidecar creates the manager for that exact path: [internal/gateway/sidecar.go](/Users/vnarajal/Desktop/defenseclaw-1/internal/gateway/sidecar.go:550)
- The manager uses `fsnotify` to watch the directory containing the config file: [internal/gateway/config_manager.go](/Users/vnarajal/Desktop/defenseclaw-1/internal/gateway/config_manager.go:93)
- It only reacts when the changed path matches the active config path and the event is `Write`, `Create`, or `Rename`: [internal/gateway/config_manager.go](/Users/vnarajal/Desktop/defenseclaw-1/internal/gateway/config_manager.go:122)
- After a 500ms debounce, it calls `Reload(ctx, "fsnotify")`: [internal/gateway/config_manager.go](/Users/vnarajal/Desktop/defenseclaw-1/internal/gateway/config_manager.go:140)

Then reload does the actual decision-making:

1. Reads the full config with `config.LoadFromFile`.
2. Diffs old vs new config.
3. If nothing changed, exits.
4. If changed, calls `s.applyConfigReload`.
5. If apply succeeds, publishes the new immutable snapshot and bumps generation.

That path is here: [internal/gateway/config_manager.go](/Users/vnarajal/Desktop/defenseclaw-1/internal/gateway/config_manager.go:147)

The default file is `config.yaml`, not `config.yml`. If someone edits a different file name than the one the gateway was started with, DefenseClaw will not reload it.

Also, this is currently event-driven. If a filesystem does not deliver `fsnotify` events reliably, there is no periodic checksum fallback yet. For local disks and normal atomic editor saves, the current watcher handles direct writes and rename-over-file saves.

vineethsai7:
Security review finding (High): `guardrailNeedsRestart` does not include the singular `guardrail.connector` field when deciding whether to restart the guardrail subsystem.

Repro/context: start with `guardrail.connector: openclaw`, so `runGuardrail` starts the proxy listener. Then PATCH or edit config to `guardrail.connector: codex`, where `proxyShouldBindForConnector` should take the hook-only/observability path and close the proxy port. Because the diff does not mark `guardrail.connector` as a guardrail restart trigger, the old listener can remain bound. The reverse transition can also leave a needed proxy listener stopped.

Affected area: `internal/gateway/sidecar.go` in `guardrailNeedsRestart` and the config reload path. Recommended fix: treat `oldCfg.Guardrail.Connector != newCfg.Guardrail.Connector` as a guardrail restart condition and add a regression test for proxy-bound ↔ hook-only connector reload.

ihabler:
**Possible gap: `setup observability migrate-splunk --apply` bypasses the managed-mode config write-gate**

While manually testing the admin-separation hardening on macOS (as a non-admin user, euid 501), I found that `migrate-splunk` is the one config-writing command that doesn't honor the managed write-gate.

**Repro** (non-admin, `deployment_mode: managed_enterprise` config with a legacy `splunk:` block):
```
DEFENSECLAW_CONFIG=<managed>.yaml defenseclaw setup observability migrate-splunk --apply
```
→ exits 0 and **rewrites the managed `config.yaml`** (removes `splunk:`, writes a new `audit_sinks:` entry). Reproduced twice.

**Root cause:** `cmd_setup_observability.py::migrate_splunk_cmd` writes via the module-local `_write_atomically()` (lines 522 & 540), which does a plain `tempfile + os.replace` and never calls `_assert_config_write_allowed`. Its siblings in the same file (`observability add` / `disable` / `remove`, lines 1023/1084/1114) correctly use `write_config_yaml_secure` and stay gated — I confirmed `remove` is rejected with `PermissionError` on a managed config. So this is the only config writer that escapes the gate.

**Impact:** Lower severity than the headline levers — it can't flip `guardrail.enabled`, `hook_fail_mode`, or `deployment_mode`. But `audit_sinks` controls where audit/telemetry is forwarded, so a non-admin on a managed host can alter the SIEM/HEC forwarding on a config the boundary is supposed to freeze.

**Suggested fix:** route `migrate_splunk_cmd`'s writes through `write_config_yaml_secure` (or call `_assert_config_write_allowed` inside `_write_atomically`), and add a managed-mode rejection test in `test_observability.py` (which currently only covers `migrate-splunk` idempotency, not gating).

PR #373, opened 2026-06-22
[codex] Add automatic application protection

vineethsai7:
Stack/base note: stacked on PR #372 (`codex/dynamic-config-config-yaml`). This change is scoped to hook-native automatic protection only. Proxy connectors remain setup-only, and this PR intentionally does not add any new `defenseclaw agent protection ...` CLI surface; visibility is through existing `defenseclaw status` and `defenseclaw status --json`. Application-protection policy rides on PR #372's `config.yaml` source-of-truth path.

## Problem

DefenseClaw already has AI discovery and can auto-enable observe behavior, but automatic application protection needed to be fully policy-driven. Operators should not have to manually onboard every supported hook-native application after DefenseClaw discovers it, but automatic action/enforce behavior must be an explicit config decision.

## Current Situation

Before this change:

- AI discovery could report local AI tooling and supported connector signals.
- Hook-native connectors still depended on manual setup paths for activation.
- Automatic connectors could inherit broad/global action posture too easily.
- Trusted binary path enforcement was treated as the default for version probing.
- `defenseclaw status` did not show automatic-vs-manual connector source, active auto-protected connectors, skipped automatic-protection reasons, activation errors, or automatic protection policy posture.

### Configurable Automatic Protection

Adds `application_protection` to both Go and Python config models, with global controls and per-connector overrides:

- `enabled`, `min_confidence`, `remove_when_gone`, `gone_after_min`
- `include_connectors`, `exclude_connectors`
- top-level automatic `guardrail.mode` and `asset_policy.mode`
- per-connector `enabled`, `min_confidence`, guardrail behavior, and asset-policy mode

Fresh defaults keep automatic discovery and hook installation enabled, but keep automatic connectors in observe mode:

```yaml
application_protection:
  enabled: true
  guardrail:
    mode: observe
  asset_policy:
    mode: observe
```

Automatic guardrail mode precedence is:

1. `application_protection.connectors.<name>.guardrail.mode`
2. `application_protection.guardrail.mode`
3. `observe`

Automatic asset-policy mode precedence is:

1. `asset_policy.connectors.<name>.mode`
2. `application_protection.connectors.<name>.asset_policy.mode`
3. `application_protection.asset_policy.mode`
4. `observe`

Global `guardrail.mode: action` and global `asset_policy.mode: action` do not promote auto-discovered connectors into action mode. Manual connectors keep existing behavior.

### Discovery-Driven Controller

[...]

PR #379, opened 2026-06-24
[feat] Add enterprise managed hardening

vineethsai7:
> **Stacked PR:** Review and merge after #372 and #373. This branch is restacked on their latest pushed heads. Commit `42ccc4a0b` contains the broad review/security fixes; the platform-hardening series includes the Windows descriptor/root-cause work through `ddf85338a`, Linux runtime fixes in `712653212`, OpenClaw E2E cleanup in `721f8cb01`, multi-connector Linux guardian hardening in `f7812b54e`, the normal-install scoped-token compatibility fix in `9a6f6f8cb`, the enterprise operator runbooks in `a682af8aa`, and the macOS/Linux Codex/Claude Code hardening matrix in `85c5c2723`, the trusted-platform-root/port-reservation portability fix in `20dc5aa81`, hosted listener diagnostics/timeout stabilization in `2ed9f43b1`, executable macOS packaging enforcement in `d94f3e9e7`, protected-path smoke correction in `8f4e5c5c2`, and test-only environment namespace correction in `fe804720b` at the current head.

## Problem

Enterprise-managed deployments need a trust boundary that local users cannot silently downgrade, bypass, or misreport. The original stack added managed mode and automatic application protection, but full code review, Codex Security review, and repeated CodeRabbit passes found several paths where mutable process state, unsafe filesystem metadata, incomplete rollback, inherited environment, or partial connector startup could weaken fail-closed enforcement.

## Current Situation

The full stack is now restacked and reviewed. The final fixes cover:

- managed-mode write protection and restart-required enforcement boundaries;
- privileged command/script trust, option termination, sudo failure propagation, and OpenClaw path pinning;
- connector-scoped hook-token validation across macOS ACLs, Linux POSIX/NFSv4 ACLs, and Windows DACL/reparse-point semantics;
- Windows `OWNER RIGHTS` handling after trusted-owner validation and effective-versus-inherit-only shared-ancestor ACL semantics, without broadening protected-directory or direct `CREATOR OWNER` access;
- correct `GetNamedSecurityInfo` ownership handling: the x/sys wrapper returns a Go-heap copy and already frees the native allocation, so callers must not `LocalFree` the returned descriptor;
- bounded Windows gateway restart recovery by retrying address-in-use failures for up to 30 seconds while still aborting immediately on cancellation or any other listen error;
- bounded API startup by skipping full Windows trust-chain walks for absent hook-token scopes while preserving full validation for every present token;
- bounded fresh hook-token creation by validating the parent chain before mkdir and then validating only the newly created directory element plus canonical containment;
- truthful connector setup completion by waiting for a fresh sidecar state generation written only after verified hook installation;
- canonical hook entrypoint resolution so symlink wrappers cannot redirect sibling resource loads;
- transactional `.env` rollback and atomic config-write behavior;
- strict manifest, TOML hook-command, shell JSON-fallback, and path-containment parsing;
- fatal scoped-token preflight before any multi-connector state mutation;
- truthful single-connector health by publishing scoped-token startup failures as `StateError` while the installed hook remains fail closed;
- connector-scoped per-user token sidecars so Codex and Claude Code cannot overwrite each other's hook credentials;
- a canonical raw scoped-token file format shared by service validation and generated hooks, preventing normal single-user installs from rewriting the service credential as shell syntax;
- authorization-gated repair of target-owned hook modes and symlinks, including removal of SUID/SGID/sticky bits;
- no-op-safe atomic reconciliation so the filesystem watcher does not trigger itself continuously;
- a Linux guardian capability model verified to regain only its bounded UID/GID/chown capabilities across repeated target-user credential transitions;
- side-effect-free `sandbox exec --help` handling through Cobra instead of treating `--help` as a privileged command argument;
- accurate local/remote guardian status reporting and deterministic test cleanup;
- detailed enterprise hardening, provisioning, security-boundary, tamper-validation, incident-response, and day-2 operations documentation;
- executable macOS packaging that atomically installs and verifies the managed config as `root:defenseclaw` mode `0640` before launchd startup; and
- privileged end-to-end CI coverage for Codex and Claude Code on both Ubuntu and macOS, including root-boundary assertions, scoped-token isolation, regular-file and symlink tamper repair, no-op stability, and a live generated-hook request.

### Immutable Managed Trust Boundary

- Pin `DEFENSECLAW_DEPLOYMENT_MODE=managed_enterprise` in managed services and reject non-admin config writes even when managed mode is environment-selected.
- Treat connector, scanner, policy, topology, deployment-mode, and managed-enforcement changes as restart-required.
- Keep managed readiness non-enforcing until every configured connector is covered by trusted guardian authorization.

### Privileged Guardian and Filesystem Trust

- Validate privileged binaries and installers through root-owned, non-writable path chains while permitting trusted alternatives-managed command symlinks.
- Pin and post-validate OpenClaw ownership-transfer targets before recursive privileged operations.
- Validate hook-token directories, ancestors, files, ACLs, DACLs, owners, modes, symlinks, junctions, and reparse points on each supported platform.
- On Windows, distinguish effective shared-ancestor rights from inherit-only templates while rejecting descendant-compromising delete-child, delete, DACL, owner, or full-control rights on every protected path element.
- Keep Windows security-descriptor lifetime under x/sys ownership; do not double-free the Go-managed descriptor returned by `GetNamedSecurityInfo`.
- Retry only Windows restart-time address-in-use failures for a bounded 30-second window so the replacement gateway can reclaim its fixed loopback API port.
- Run Linux guardian jobs with `NoNewPrivileges=false` because the long-lived watcher must regain bounded `CAP_SETUID`/`CAP_SETGID` after temporarily adopting each allow-listed target user; retain a five-capability bounding set, syscall filter, SUID/SGID restriction, and narrow writable paths.
- Fail closed when ACL semantics cannot be verified.

### Hook Runtime and Credential Hardening

- Resolve the hook file itself before deriving `HOOK_DIR`, preventing symlink wrappers from redirecting `_hardening.sh` or sibling resources.
- Issue connector-scoped protected tokens and revalidate protected disk state on every authentication.
- During bulk token discovery, omit absent scope files before expensive trust validation; any present file still requires the complete path, owner, ACL/DACL, and reparse-point checks.
- During fresh token creation, validate the data directory before mkdir; for a newly created hooks directory, validate its own owner/ACL/reparse state and canonical containment without rewalking every trusted ancestor.
- Write and consume raw `.hook-<connector>.token` credentials in shared per-user hook directories, making the scoped sidecar authoritative over inherited generic credentials while retaining shell-formatted `.token` only as a legacy fallback.
- Parse fallback JSON decisions only from top-level fields and preserve jq-compatible `empty`, `null`, and explicit non-string behavior.
- Match managed commands structurally in JSON/YAML/TOML rather than accepting decoy or wrapper text.

### Linux Watcher Repair Semantics

- Refuse unsafe first-time installs, foreign-owned paths, and unsafe parent directories.
- After root authorization exists for the exact user/home/connector target, normalize target-owned writable or special-bit modes and replace target-owned symlinks with canonical files.
- Preserve mtimes and ownership when content and metadata already match, preventing self-generated `fsnotify` repair loops.
- Treat repair as detection and bounded recovery, not instantaneous prevention: a target user still owns the native agent configuration and can create a short bypass window before the watcher restores it. Operators requiring prevention must pair this with endpoint controls that deny user writes.

### Transactional Configuration and Startup

- Snapshot exact dotenv bytes/mode, reject special files, and restore through same-directory atomic replacement.
- Keep config writes atomic and serialize read/merge/write cycles.
- Preflight every connector-scoped token before connector setup or credential mutation; publish `StateError` on any preflight failure.
- Publish every non-cancellation guardrail startup error through the central guardrail health state, including single-connector scoped-token failures.
- Roll back partially started hook guards and failed connector setup deterministically.
- Make `setup <connector> --restart` wait for a fresh, matching `active_connector.json` generation and fail closed on timeout; validate the state descriptor before parsing to prevent symlink/special-file swap races.
- Parse `sandbox exec` flags normally so help is available before config/audit initialization, while requiring `--` before command-specific flags.

### Packaging, Docs, and CI

- Align systemd/launchd ownership, capability, runtime-directory, and guardian-manifest guidance.
- Document the complete enterprise security architecture, explicit threat model and non-goals, administrator provisioning and rollout, managed token and authorization flows, Linux and macOS service differences, day-2 changes, incident recovery, and the bounded-recovery reason a non-privileged AI agent cannot permanently remove the system control plane.
- Add a lab-safe hardening validation runbook with root-boundary, API downgrade, hook/config/token tamper, symlink-decoy, mode-normalization, watcher, and systemd capability tests.
- Run the real privileged guardian CLI in a required four-cell Ubuntu/macOS × Codex/Claude Code matrix using isolated root-owned `/var/lib` (Ubuntu) or `/Library` (macOS) control planes and disposable user-owned agent homes.
- Ship `packaging/launchd/install-enterprise.sh` in release archives so macOS ownership and mode requirements are executable packaging behavior rather than documentation-only commands; refuse symlink destinations, verify trusted system parents, atomically install files, and validate every owner/mode before loading launchd.
- Run a native macOS packaging smoke with a disposable non-login service identity, metadata drift repair, standard-user write denial, and symlink-decoy refusal.
- Assert root ownership and non-writability, raw connector-scoped token format, canonical repair after replacement and symlink attacks, unchanged decoys, stable mtimes on no-op reconcile, and live hook authentication despite an attacker-controlled inherited generic token.
- Document why only guardian jobs opt out of `NoNewPrivileges`, while the gateway retains `NoNewPrivileges=true` and an empty capability set.
- Harden generated policy paths and shell values against control characters and whitespace ambiguity.
- Terminate the exact OpenClaw PID started by core E2E, with bounded `SIGTERM` grace and a tracked-PID-only `SIGKILL` fallback.
- Preserve required live E2E and cross-platform packaging coverage.

```mermaid
flowchart LR
    Admin["Enterprise administrator"] --> Config["Root-owned config and pinned managed mode"]
    Config --> Gateway["DefenseClaw gateway"]
    Gateway --> Desired["Configured connector targets"]
    Desired --> Guardian["Privileged hook guardian"]
    Guardian --> Hooks["Canonical managed hooks"]
    Guardian --> Ledger["Root-owned authorization ledger"]
    Ledger --> Health["Managed readiness and enforcement health"]
    Hooks --> Token["Connector-scoped protected token"]
    Token --> ACL["Mode + owner + ACL/DACL validation"]
    ACL --> API["Loopback hook inspection API"]
    API --> Gateway
```

## What Changed (Where & How)

| Area | Paths | Change |
| --- | --- | --- |
| Privileged sandbox setup | `cli/defenseclaw/commands/cmd_init_sandbox.py`, `cli/tests/test_cmd_setup_sandbox.py` | Resolves trusted helpers, validates root-owned chains, propagates sudo/chmod/remove failures, pins OpenClaw targets, and preserves version overrides across sudo. |
| Transactional config | `cli/defenseclaw/commands/cmd_setup.py`, `cli/defenseclaw/config.py`, trusted-path tests | Adds exact snapshots, regular-file checks, atomic restore, managed env-pin protection, and dry-run-safe writers. |
| Hook token trust | `internal/gateway/connector/hook_api_*` | Adds canonical ancestor validation and platform-specific macOS ACL, Linux POSIX/NFSv4 ACL, and Windows DACL/reparse-point checks; permits `OWNER RIGHTS` only after trusted-owner validation, ignores inherit-only templates on shared ancestors whose actual descendants are independently checked, rejects direct `CREATOR OWNER` or effective delete-child access, bounds trust walks, and avoids double-freeing Go-owned descriptors. |
| Hook runtime | `internal/gateway/connector/hooks/`, `hook_config_paths.go` | Resolves symlinked entrypoints, hardens JSON fallback semantics, and parses managed hook commands structurally. |
| Multi-connector user credentials | `internal/gateway/connector/hook_api_token.go`, `subprocess.go`, `hookexec/`, hook templates | Uses canonical raw connector-scoped user token sidecars, keeps native scoped-over-legacy precedence, clears inherited generic credentials when scoped files are absent/unreadable, and prevents shared `.token` collisions. |
| Multi-connector startup | `internal/gateway/sidecar.go`, `sidecar_multi_boot_test.go` | Preflights all scoped tokens, prevents partial-success activation on trust failure, and cleans up partially started guards. |
| Linux startup truthfulness | `internal/gateway/sidecar.go`, `sidecar_multi_boot_test.go` | Publishes single-connector scoped-token startup failures as guardrail errors instead of leaving health at disabled while hooks reject requests. |
| Sandbox command parsing | `internal/cli/sandbox.go`, `internal/cli/sandbox_test.go` | Makes `sandbox exec --help` side-effect free, preserves command flags after `--`, and retains sudo option termination. |
| Setup readiness | `cli/defenseclaw/commands/cmd_setup.py`, `cli/tests/test_guardrail.py`, `internal/gateway/api.go` | Waits for verified connector runtime state after restart, rejects stale or swapped state descriptors, fails setup instead of reporting completion before hooks exist, and gives Windows a bounded 30-second address-reuse window for the replacement loopback API listener. |
| Guardian/config semantics | `internal/enterprisehooks/`, `internal/config/`, `internal/cli/status.go` | Tightens manifest parsing, watch paths, managed config behavior, status locality, and authorization reporting. |
| Guardian repair loop | `internal/enterprisehooks/`, `internal/gateway/connector/atomicfile.go`, `managed_backup.go` | Repairs authorized target-owned symlinks/modes, preserves unchanged files/contracts, and ignores watcher lock housekeeping to eliminate self-trigger loops. |
| Linux guardian sandbox | `packaging/systemd/defenseclaw-hook-guardian*.service`, `cli/tests/test_enterprise_packaging.py` | Allows bounded privilege regain across target-user transitions while retaining the explicit capability bound and all remaining service hardening. |
| Enterprise operator docs | `docs-site/content/docs/setup/enterprise-deployment.mdx`, `enterprise-hardening-validation.mdx`, setup navigation | Documents architecture, guarantees and non-goals, Linux/macOS provisioning, manifest and token models, bounded repair, admin workflows, incident handling, production checklists, destructive validation, evidence capture, and automated CI coverage. |
| Enterprise hardening CI | `.github/workflows/ci.yml`, `scripts/test-enterprise-hook-hardening.sh` | Adds Ubuntu/macOS × Codex/Claude Code privileged reconciliation tests for root boundaries, ownership/modes, scoped credentials, regular and symlink tamper repair, no-op stability, cleanup, and live hook authentication. |
| macOS enterprise packaging | `packaging/launchd/install-enterprise.sh`, `scripts/test-macos-enterprise-packaging.sh`, `cli/tests/test_enterprise_packaging.py` | Adds an archive-shipped installer that enforces `root:defenseclaw 0640` config ownership atomically, verifies the complete launchd payload, refuses symlink destinations, and exercises install/repair/refusal behavior on native macOS CI. |
| Docs/tooling | `docs-site/` | Aligns enterprise ownership/runtime guidance and rejects unsafe policy-creator path values. |
| Telemetry | `internal/telemetry/` | Clears and restores global OTel error-handler state safely. |
| Windows CI | `.github/workflows/ci.yml` | Executes hook-token trust tests natively on `windows-latest` so descriptor lifetime and DACL behavior are runtime-tested rather than compile-only. |
| OpenClaw E2E cleanup | `scripts/test-e2e-full-stack.sh` | Stops the manually started core-profile OpenClaw PID deterministically so the persistent runner is not left listening after teardown. |

## Breaking Changes

- Managed packages pin `DEFENSECLAW_DEPLOYMENT_MODE=managed_enterprise`; local config cannot downgrade the running service.
- Security-critical configuration changes require restart rather than partial hot reload.
- Managed health remains starting/non-enforcing until trusted guardian authorization covers every configured target.
- Hook-token paths with unverifiable, untrusted, writable, symlinked, reparse-point, or malformed ACL state now fail closed.
- Managed per-user hook credentials use `.hook-<connector>.token`; existing `.token` files remain a compatibility fallback but clean multi-connector installs do not create or overwrite the shared sidecar.
- Linux guardian units set `NoNewPrivileges=false` so repeated target-user reconciliation works; this applies only to guardian jobs and is constrained by their five-capability bounding set and other systemd sandboxing.
- Once a target is root-authorized, the guardian may remove target-owned symlinks and normalize target-owned unsafe modes in its declared footprint; foreign-owned objects remain hard failures.
- Privileged sandbox installation refuses unpinned OpenClaw integrations and untrusted helper/installer paths.
- Hook-connector setup with `--restart` can now wait up to 60 seconds and exits non-zero if the sidecar does not publish a fresh verified connector state.
- A restarting gateway can now spend up to 30 seconds retrying only an address-in-use failure for its fixed loopback API port; cancellation and all other listen failures remain immediate.
- `defenseclaw-gateway sandbox exec` now requires command-specific flags to follow `--` (for example, `sandbox exec -- command --flag`); this keeps gateway flags and child-command flags unambiguous.
- macOS enterprise installation now requires the shipped installer (or equivalent MDM policy) to establish and verify the documented service identity and filesystem metadata. The installer unloads existing jobs before replacement, rejects symlink destinations, and leaves jobs stopped when `--no-start` is selected.

## Open Issues / Follow-ups

None.

## Test Plan

Final observed validation:

- `go test ./...` — passed across all Go packages.
- `go test ./... -count=1` — passed at `f7812b54e` after the final multi-connector guardian changes.
- `go test ./... -count=1` — passed at `9a6f6f8cb` after the normal-install scoped-token format and precedence fixes.
- `pytest cli/tests/test_enterprise_packaging.py` — `6 passed` at `f7812b54e`.
- `make cli-test` — `4647 passed, 3 skipped, 4 warnings, 335 subtests passed`.
- `go test ./internal/gateway/connector ./internal/telemetry ./internal/cli ./internal/gateway -count=1` — passed after the closing review fixes.
- `go test -race ./internal/gateway/connector ./internal/enterprisehooks` — passed.
- `go test ./internal/gateway/connector -count=1` — passed after the Windows ACL CI fixes.
- `go test -race ./internal/gateway/connector -count=1` — passed after the Windows ACL CI fixes.
- `go test ./internal/gateway -run 'IsAddrInUse|ListenWithRetry' -count=1` — passed after extending the bounded Windows restart bind window.
- `go test ./internal/gateway -count=1` — passed after extending the bounded Windows restart bind window.
- `go test ./internal/cli ./internal/gateway -count=1` and `go test ./internal/gateway/connector -count=1` — passed after the Linux runtime fixes.
- `GOOS=linux GOARCH=amd64 go test -c ./internal/gateway/connector -o /tmp/defenseclaw-connector-linux.test` — passed.
- `GOOS=windows GOARCH=amd64 go test -c ./internal/gateway/connector -o /tmp/defenseclaw-connector-windows.test.exe` — passed.
- `cd cli && uv run pytest -q tests/test_cmd_setup_sandbox.py tests/test_cmd_setup_trusted_paths.py tests/test_config.py` — `249 passed` after the closing review fixes.
- `cd cli && uv run pytest -q tests/test_guardrail.py tests/test_cmd_setup_multi_connector.py tests/test_cmd_setup_rotate_token.py` — `186 passed` after the connector-readiness fix.
- `cd cli && uv run pytest -q tests/test_cmd_setup_codex_claudecode_alias.py` — `36 passed, 32 subtests passed`.
- `cd docs-site && npm run test:policy-creator` — `55 passed`.
- `cd docs-site && npm run build` — passed.
- `uv run ruff check` on changed Python files — passed.
- `bash -n` on every changed hook script — passed.
- `git diff --check` — passed.
- `bash -n scripts/test-enterprise-hook-hardening.sh` — passed.
- `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/ci.yml")'` — workflow YAML parsed successfully.
- `go test ./internal/enterprisehooks ./internal/cli ./internal/managed ./internal/gateway/connector/... -count=1` — passed after adding the hardening harness.
- `pytest -q cli/tests/test_enterprise_packaging.py` — `6 passed` after adding the hardening harness.
- `pytest -q cli/tests/test_enterprise_packaging.py` — `7 passed` after adding the macOS installer regression.
- `bash -n packaging/launchd/install-enterprise.sh scripts/test-macos-enterprise-packaging.sh` — passed.
- Pre-fix regression proof — the focused packaging test failed because `packaging/launchd/install-enterprise.sh` did not exist; it passes after the enforcement script was added.
- `cd docs-site && npm run build` — passed with 145 static pages; postbuild generated 68 `llms.md` files and validated 24 diagrams.
- `cd docs-site && npm run validate-links` — `0 errored file, 0 errors`.
- `scripts/test-enterprise-hook-hardening.sh <linux-arm64-binary> codex` — passed on the authorized Ubuntu ARM64 host.
- `scripts/test-enterprise-hook-hardening.sh <linux-arm64-binary> claudecode` — passed on the authorized Ubuntu ARM64 host.
- Post-harness cleanup check — no `defenseclaw-enterprise-ci-*` fixtures remained under remote `/opt` or `/tmp`.
- Local macOS execution was not counted because this workstation requires an interactive sudo password; the required `macos-latest` CI cells provide the privileged native evidence.
- `coderabbit review --agent -t uncommitted -c AGENTS.md` — CodeRabbit raised 0 issues on the `d56b787a2` Windows ACL delta; later follow-up attempts were rate-limited before producing a result.
- CodeRabbit raised 1 major issue on the connector-readiness delta (descriptor stability was checked after parsing); fixed by validating the no-follow file descriptor before JSON parsing. The immediate closing retry was rate-limited before producing a result.
- `coderabbit review --agent -t uncommitted -c AGENTS.md` — CodeRabbit raised 0 issues on the final new-directory trust-walk optimization.
- `coderabbit review --agent -t uncommitted -c AGENTS.md` — CodeRabbit raised 0 issues on the final Windows descriptor-ownership fix and native regression coverage.
- `coderabbit review --agent -t uncommitted -c AGENTS.md` — CodeRabbit raised 0 issues on the canonical reparse-rejection test adjustment.
- `coderabbit review --agent -t uncommitted -c AGENTS.md` — CodeRabbit raised 0 issues on the final bounded Windows API restart retry change.
- CodeRabbit raised 1 minor issue on the Linux runtime delta (package-global Cobra state was not fully reset); fixed, then CodeRabbit raised 0 issues.
- CodeRabbit raised 1 major issue on OpenClaw E2E teardown (an unconditional wait could hang after `SIGTERM`); fixed with a bounded tracked-PID hard-stop fallback. The immediate closing retry was rate-limited for 36 minutes before producing a result.
- CodeRabbit's final Linux multi-connector passes raised 9, then 3, then 3 findings. Valid findings were fixed (scoped user credentials, authorized mode repair, no-op atomic identity checks, native token precedence coverage, template token-path initialization, exact systemd assertions, and watcher-loop prevention); the `NoNewPrivileges=false` warning was disproved by the required live credential-regain test and bounded runtime capability inspection. The closing confirmation retry was rate-limited for 29 minutes before producing a result.
- CodeRabbit's CI-repair passes raised 2, then 2, then 1 findings. All were fixed: raw parsing is scoped-only, failed/missing scoped reads clear inherited generic credentials, exact sidecar bytes are asserted directly, and opaque token bytes preserve spaces while removing only the writer's trailing LF. The post-fix confirmation retry was rate-limited for 22 minutes before producing a result.

Linux ARM64 host validation on Ubuntu 24.04.4 / kernel 6.17 / OpenClaw 2026.6.8:

- Recursive command-surface smoke — all 211 Python CLI command paths and all 41 gateway command paths returned help successfully after the fix; clean-host `sandbox exec --help` created no audit database.
- `TMPDIR=$HOME/.tmp-dc-e2e` with private `0700` ownership and `umask 022` plus `go test ./...` — all Go packages passed on native Linux ARM64.
- `make cli-test` — `4650 passed, 4 skipped, 4 warnings, 335 subtests passed`.
- `make ts-test` — 20 files and 259 tests passed.
- `make rego-test` — 93/93 policy tests passed.
- `make check` — v7 parity, provider coverage, LLM catalog, and upgrade manifest checks passed.
- Dedicated systemd managed deployment — active under the required `defenseclaw` service identity with `ProtectSystem=strict`, `NoNewPrivileges=true`, empty capability bounding set, root-owned config/authorization, loopback-only API, and `systemd-analyze security` exposure `1.5 OK`; five rapid restart/readiness cycles passed before systemd's expected default start-burst limiter engaged.
- Managed negative tests — rejected an untrusted service UID, mismatched target-home ownership, non-admin CLI writes, authenticated API downgrade, service-user ledger writes, environment-pinned deployment downgrade, group-writable config, and world-readable scoped-token mode.
- Guardian tests — root-owned manifest reconciliation enabled enforcement, user hook tamper was repaired both manually and through `fsnotify`, and allow/block golden events produced 2 inspections, 1 deterministic block, and 0 errors.
- Scoped-token tamper — health reported `guardrail=error`, the installed hook denied an otherwise-safe request with exit 2, and readiness recovered after restoring mode `0600`.
- OpenClaw core E2E — 76 passed, 0 failed, 9 expected environment skips using a local TLS HEC stub; skill/MCP scanning, block/allow, quarantine/restore, watcher, CodeGuard, status/doctor, AIBOM, policy, skill API, architecture, upgrade, and HEC contracts passed.
- Focused live OpenClaw/Bedrock E2E — the embedded DefenseClaw plugin was registered, gateway API and action-mode guardrail were running on loopback, the real OpenClaw agent returned the exact `DEFENSECLAW_LIVE_STACK_OK` marker through the host's Bedrock role, and a deterministic prompt-injection plus `/etc/shadow` request returned a DefenseClaw `content_filter` block without reaching the model.
- Post-run cleanup — `.openclaw` and `.defenseclaw` matched their pre-run snapshots byte-for-byte; the tracked residual OpenClaw PID was identified and terminated, motivating `721f8cb01`.

Final isolated Codex + Claude Code Linux hardening validation on Ubuntu 24.04 ARM64:

- Installed CLIs — Codex `0.142.0` and Claude Code `2.1.187`; both command surfaces returned help successfully as the non-privileged target user.
- Dual managed hooks — canonical Codex and Claude Code golden allow/block payloads passed; each connector recorded 2 inspections and 1 deterministic block with isolated scoped credentials.
- Root boundary — the target user could not read service-side scoped tokens or modify root-owned managed config, guardian authorization, or the installed gateway binary; checksums remained unchanged.
- Credential isolation — making the service-side Codex token world-readable forced Codex health to error and the safe Codex hook to fail closed with HTTP 401/exit 2, while Claude Code remained healthy; restoring `0600` recovered both.
- Managed API downgrade — an authenticated Python CLI client received HTTP 403 when attempting to modify managed-enterprise configuration.
- User tamper recovery — Codex hook replacement repaired in 235 ms; Claude hook deletion in 251 ms; Codex/Claude config stripping in 237 ms each; config symlink in 288 ms; hook symlink in 212 ms; corrupted Codex token in 238 ms; deleted Claude token in 252 ms; mode `04777` in 219 ms. Decoy symlink targets were not modified.
- Availability behavior — corrupted Codex credentials failed closed; deleting the Claude token followed the documented default availability policy and temporarily allowed until repair. This confirms the watcher is bounded recovery, not prevention against a user who owns the native agent files.
- Watcher stability — after repair settled, zero additional reconciles occurred during a three-second observation window.
- Runtime sandbox — gateway exposure score `1.5 OK`; guardian exposure score `2.8 OK`; guardian runtime showed `NoNewPrivs: 0` with only `CAP_CHOWN`, `CAP_DAC_OVERRIDE`, `CAP_FOWNER`, `CAP_SETGID`, and `CAP_SETUID` permitted/effective/bounded.
- Native ARM64 focused suites — `umask 022; go test ./internal/enterprisehooks ./internal/cli ./internal/gateway/connector/... -count=1` passed.
- Scope note — real generated hook scripts and official golden payloads were exercised end to end. A direct model-backed `codex exec` prompt harness was not counted because the remote process-group timeout behavior was not reliable enough to serve as evidence.

Remote CI at `9a6f6f8cb` — 52 checks passed, 2 non-required deployment/live-matrix checks skipped, 0 pending, and 0 failed. The first `f7812b54e` run exposed HTTP 401 across the connector contract matrix because the normal single-user layout shared one scoped-token path between the service and hook writer; `9a6f6f8cb` fixes that root cause.

The first matrix run at `85c5c2723` exposed two harness-only CI assumptions: hosted Ubuntu makes `/opt` group/other-writable, which correctly failed managed trust validation, and a late ephemeral-port bind could race on hosted macOS. `20dc5aa81` moves fixtures to trusted platform roots and reserves the loopback listener for the full test. Both corrected Linux cells passed again on the authorized Ubuntu ARM64 host.

At `20dc5aa81`, Ubuntu Codex, Ubuntu Claude Code, and macOS Codex passed. The macOS Claude Code cell timed out before its background test listener wrote readiness; no DefenseClaw command had run. `2ed9f43b1` extends the hosted startup allowance and captures the background listener log on any future failure.

CI at `2ed9f43b1` was green: 56 checks passed, 2 expected non-required checks skipped, 0 pending, and 0 failed. All four `Enterprise Hook Hardening` cells passed for `ubuntu-latest`/`macos-latest` × `codex`/`claudecode`. At `d94f3e9e7`, the native installer completed and verified its payload; the smoke then correctly encountered permission denial when the standard runner attempted to traverse the new `0750 root:defenseclaw` parent for `stat`. `8f4e5c5c2` inspects protected metadata through the admin channel while preserving the unprivileged write-denial assertion, and the native macOS packaging job passed. The unified suite then rejected the test-only `DEFENSECLAW_PACKAGING_SMOKE` name because product-prefixed variables must be declared in the runtime registry; `fe804720b` renames the CI-only guard outside that namespace. Current-head CI is green: 57 checks passed, 2 expected non-required checks skipped, 0 pending, and 0 failed. The native `macOS Enterprise Packaging` job installed and verified the payload twice, repaired intentionally widened/foreign-owned config metadata, rejected the config symlink, and left its decoy unchanged.


vineethsai7:
Security review findings:

1. High: the packaged hook guardian systemd units omit `CAP_SETUID` and `CAP_SETGID`, but `internal/enterprisehooks/owner_unix.go` drops credentials with `Setgroups`, `Setegid`, and `Seteuid`. With the current `CapabilityBoundingSet=CAP_CHOWN CAP_DAC_OVERRIDE CAP_FOWNER`, the first reconcile/install under the packaged units should fail with `operation not permitted` before hook repair can write the target user's files. Add `CAP_SETUID CAP_SETGID` to both guardian units and cover the expected capability set in packaging tests.

2. High: Windows enterprise hook hardening is effectively a no-op. The Windows implementations for owner validation / file owner matching / token owner checks return success, so an elevated enterprise install can target incorrectly owned or user-controlled homes/config/token files on Windows while Unix rejects the same trust violations. Either implement Windows owner/ACL checks equivalent to the Unix guarantees or explicitly disable managed enterprise hook installation on Windows until those checks exist.

ihabler:
## Manual admin-separation testing — findings (G1–G3)

I ran a hands-on manual test pass of the admin/non-admin trust boundary on **macOS** (real non-admin user, `euid 501`), exercising the CLI gate, OS file ownership, the gateway trust check, and the loopback API. Every non-admin attempt to disable/weaken/downgrade enforcement was rejected at the CLI layer, and the gateway independently fail-closes on untrusted managed configs and authenticated API downgrades. Three findings below.

### G1 — Managed protection depends on a **root-owned** config; the CLI gate alone is bypassable on a user-owned file (highest impact)

The non-admin CLI gate (`cli/defenseclaw/config.py::_assert_config_write_allowed`) correctly blocks `guardrail disable`, `guardrail fail-mode open`, and `settings save` in managed mode (verified — config left byte-identical after each rejection). **However, it only checks `euid == 0`; it does not check who owns `config.yaml`.**

If the managed `config.yaml` is ever user-writable (e.g. a user-scoped layout that carries `deployment_mode: managed_enterprise` under `~/.defenseclaw`), a standard user bypasses the gate entirely by editing the file directly (`sed`/editor). In my test I disabled the guardrail, flipped fail-closed → fail-open, and downgraded to `unmanaged_byod` this way, and `defenseclaw guardrail status` then reported the weakened posture.

The **running gateway is not foolable**: `internal/managed/trust_unix.go::ValidateTrustedConfigPath` refuses to load a managed config unless it + every ancestor are root-owned, non group/other-writable, and symlink-free. The built gateway fails closed:
```
config: managed_enterprise config trust check failed:
.../config.yaml: owner uid 501 is not trusted for managed config; expected root/admin uid 0
```
So this is a **deployment-correctness requirement, not a gateway defect** — the guarantee holds **iff** the managed config is root-owned at a trusted path.

**Asks / recommendations**
- Can we confirm macOS packaging guarantees the managed config is provisioned `root:wheel` / `0644`? The enterprise-deployment docs show **manual** `sudo install …` steps into `/Library/Application Support/DefenseClaw/` — is there (or should there be) a signed `.pkg`/postinstall that enforces ownership automatically, rather than relying on the operator?
- Consider having the **CLI warn** when it reads `deployment_mode: managed_enterprise` from a file that is **not root-owned**, so a bypassable/misconfigured deployment is visible instead of silently trusted (the gateway already validates ownership; the CLI does not, so its reported state can be misleading).

### G2 — `settings save` raises an unhandled `PermissionError` (raw traceback) in managed mode

The `guardrail` subcommands print a clean `✗ Failed to save config: managed_enterprise config changes require operating-system administrator privileges…`. But `cli/defenseclaw/commands/cmd_settings.py::settings_save` calls `app.cfg.save()` **without catching `PermissionError`**, so a non-admin gets a full Python traceback and a hard crash exit. The boundary still holds (no write occurs), but the UX is inconsistent and noisy.

**Recommendation:** wrap `app.cfg.save()` with the same friendly handler the `guardrail` commands use.

### G3 — `guardrail disable` prompts interactively before the rejection surfaces (minor)

`guardrail disable` requires an interactive `Proceed?` confirmation unless `--yes` is passed, so in managed mode the rejection only appears **after** the prompt. Not a security issue (the write is still gated) — flagging only for scripted/automation paths.

---
*Layers verified: CLI gate (`euid==0`) + OS file ownership + gateway `ValidateTrustedConfigPath` (fail-closed on load) + API `PATCH /guardrail/config` → 403 in managed mode even with a valid gateway token. Full per-test log kept locally.*

PR #403, opened 2026-06-27
[codex] Add Galileo multi-OTel and Agent360 observability

vineethsai7:
Stack/base note: this PR is based directly on `main` and intentionally combines Galileo, named multi-destination OTLP, hook telemetry, canonical schemas, Agent360, and the associated upgrade path. The end-to-end behavior depends on those layers shipping together.

## Problem

DefenseClaw could not reliably fan the same telemetry out to local observability and Galileo. Legacy OTel configuration modeled one exporter, Galileo lacked a complete cloud/self-hosted workflow, and hook-based connectors often appeared only after a terminal Stop event. Long-running agents therefore looked absent or incomplete.

Grafana also lacked a reusable per-agent drill-down. Identity, lifecycle, model/tool activity, tokens, costs, security decisions, and Tempo traces were split across signals without a stable correlation contract. Sparse or restarted Prometheus counter series made real activity render as zero, while several range-wide latency panels evaluated only the current rate window and rendered no data.

Finally, runtime compatibility alone did not make the configuration upgrade durable. A v6 flat OTel exporter could be converted in memory but remain flat on disk. Published 0.8.x hosts may already have the `0.8.0` release cursor marked, and legacy upgraders can retain old Python modules while replacing the wheel in-process.

## Current Situation

This branch now supports named OTel fan-out, Galileo cloud/self-hosted setup and testing, real-time hook traces, canonical GenAI schemas, and a dynamic Agent360 dashboard. The review and live verification pass also fixed:

1. durable flat OTel upgrade/migration, including environment-backed trace/log/metric routes and already-applied 0.8.x cursors;
2. cross-signal agent-ID mismatches, duplicate lifecycle counting, and mixed cumulative/incremental cost semantics;
3. premature export of guardrail-blocked model output and credential-bearing endpoint display;
4. terminal-success and error/approval panel semantics;
5. an OTel Collector v0.111.0 self-metrics panic that restarted the collector on each Prometheus scrape;
6. Agent360 counters and range analytics that showed zero/no-data despite durable Loki events and Tempo/Prometheus observations;
7. stale Python bytecode in release wheels and mixed old/new module state during legacy in-process upgrades.

Tokens and cost remain report-only. If a connector does not supply usage or cost, Agent360 explicitly renders `Not reported` instead of estimating or fabricating zero.

## Goals

1. Fan one normalized telemetry stream to multiple independent OTLP destinations without coupling their availability or backpressure.
[...]

PR #404, opened 2026-06-27
fix(audit): redact network egress URLs

vineethsai7:
## Problem

PR #403 adds structured network egress rows that persist and expose `NetworkEgressRow.URL`. Full URLs can contain credentials in query strings or userinfo, so storing and returning the raw value can leak secrets through SQLite, `/api/v1/network-egress`, and blocked egress alert details.

## Current Situation

The telemetry log path intentionally avoids exporting full URLs, but the audit row path persisted `NetworkEgressEvent.URL` after only length truncation. A tool-observed URL such as `https://api.example.com/v1/export?api_key=...` could therefore round-trip through the audit database and API response.

## Solution

Redact network egress URLs before constructing the persisted row, and use that same redacted value when building blocked-alert details. Hostname, method, decision code, policy outcome, and blocked status remain structured for investigation.

## Architecture Diagram

N/A

## What Changed (Where & How)

| File | Summary |
| --- | --- |
| `internal/audit/network_egress.go` | Redacts URL before persistence and reuses the redacted row URL in blocked-alert details. |
| `internal/audit/network_egress_test.go` | Adds regression coverage for allowed and blocked URL credential redaction. |

## Breaking Changes

The `url` field in network egress audit rows now contains a redaction placeholder instead of the raw URL. The structured `hostname`, `http_method`, `protocol`, decision, severity, and correlation fields are unchanged.

## Open Issues / Follow-ups

None.

Related PR: #403

## Test Plan

- `go test ./internal/audit`
  - `ok github.com/defenseclaw/defenseclaw/internal/audit 2.936s`


vineethsai7:
Daily security/code review finding: this branch is still based on the earlier #403 commit `69a62243`, while `codex/galileo-multi-otel` has advanced to `4a76b601` with Agent360 hardening. A local three-way merge now reports conflicts in the Agent360 dashboard, collector config, observability writer, and related tests; the current branch comparison also drops those hardening changes. Please rebase/cherry-pick the redaction commit onto the current #403 head before merge so the URL-redaction fix does not conflict with or regress the latest Agent360 hardening. Verified the narrow redaction patch itself with `GOCACHE=/tmp/defenseclaw-review-gocache go test ./internal/audit` (`ok ... 3.139s`).

PR #431, opened 2026-07-02
fix(macos): keep redaction enabled by default

vineethsai7:
## Problem

PR #410 sets the macOS managed enterprise installer default to `DISABLE_REDACTION="true"`, causing a default install to write `privacy.disable_redaction: true`. That disables sink redaction for prompts, raw bodies, tool arguments, and other LLM telemetry unless the operator notices and passes `--redact`.

## Solution

Keep redaction enabled by default (`disable_redaction: false`), preserve `--redact` as a backward-compatible explicit enable flag, and add `--no-redact` for the opt-out case.

## What Changed

| Area | Changes |
|---|---|
| macOS installer | Default redaction setting now writes `disable_redaction: false`. |
| CLI help | Documents default-on redaction and explicit `--no-redact`. |
| Tests | Renamed/rendered redaction cases to pin default-on behavior. |

## Breaking Changes

Default macOS installs now redact sink telemetry unless operators explicitly pass `--no-redact`.

## Test Plan

- `packaging/macos/tests/run_tests.sh packaging/macos/tests/test_render_config.sh`
  - Result: `5 passed, 0 failed in 1s`
- `packaging/macos/tests/run_tests.sh`
  - Result: `51 passed, 2 failed in 6s`; failures were unrelated existing agent-version discovery assertions in `test_discover_agent_version.sh`.

## CI Status


sanjayab23:
Closed as duplicate — the same finding was addressed in #410 during the branch rebase.

#410 commit `f19d9775 Secure macOS installer redaction default` flips the default to `disable_redaction: false` and adds `--disable-redaction` / `--no-redact` as the explicit opt-out. Rendered-config and `--help` tests were updated to assert the secure default. Verified via `make packaging-macos-test` (55 passed).

The naming choice differs slightly — this PR uses `--redact` (semantics kept, meaning "stay on") with `--no-redact` as the opt-out; #410 uses `--disable-redaction` as the primary flag with `--no-redact` as an alias, to avoid the double-negative that a bare `--redact` implies when the default is already on. Both are secure; #410 seemed the clearer surface.

Feel free to close #431 whenever.

PR #432, opened 2026-07-02
fix(macos): reject symlinked userspace hook paths

vineethsai7:
## Problem

PR #410 prepares per-user connector hook config files from the root macOS installer before the Go enterprise hook hardening runs. The shell helper follows symlinked `~/.codex`, `~/.claude`, `~/.cursor`, or config-file paths during mkdir/chmod/write, and the installer then runs recursive chown on those user-controlled paths. A target user can pre-create symlinks and cause root-side filesystem changes outside the intended agent config directories.

## Solution

Reject symlinked connector directories and config files in the shell userspace-prep helpers before any chmod/write happens. The real installer runs with `set -e`, so a rejected path stops before the follow-up chown.

## What Changed

| Area | Changes |
|---|---|
| macOS installer helpers | Added a shared userspace path guard for connector config directories/files. |
| Tests | Added regression coverage for symlinked agent directories and symlinked config files. |

## Breaking Changes

Tampered or symlinked user hook config paths now fail installer userspace prep instead of being followed.

## Test Plan

- `packaging/macos/tests/run_tests.sh packaging/macos/tests/test_prepare_userspace.sh`
  - Result: `6 passed, 0 failed in 1s`

## CI Status


sanjayab23:
Cherry-picked into #410. Commits:

- `09df8288 fix(macos): reject symlinked userspace hook paths` — the original commit from this branch, verbatim.
- `450b8987 fix(macos): skip chown when prep helper rejects symlink, use chown -h` — defense-in-depth follow-up on the install.sh side that (a) skips the subsequent `chown` when `ensure_safe_userspace_path` returned non-zero and (b) replaces `chown -R` with `find -exec chown -h + } +` so we never resolve symlinks at chown time even if one raced in.

`make packaging-macos-test` reports 55 passed (was 53 — the two new symlink-rejection cases you added are green).

Feel free to close #432.

PR #433, opened 2026-07-02
fix(gateway): verify hook tokens against disk

vineethsai7:
## Problem

PR #410 preloads connector-scoped hook API tokens into memory. `hookAPITokenMatches` accepts a cached token before checking the token file, so a deleted or rotated `.hook-<connector>.token` file does not revoke the old token until gateway restart.

## Solution

When `DataDir` is configured, treat the on-disk scoped token file as authoritative on each hook-token authentication check. Delete stale cache entries when the token file is missing or invalid. Preserve the existing in-memory fallback only for no-data-dir legacy tests/configurations.

## What Changed

| Area | Changes |
|---|---|
| Gateway auth | Hook-scoped bearer auth now verifies the current token file before accepting a token. |
| Tests | Added regression coverage for deleted and rotated scoped hook tokens. |

## Breaking Changes

Deleted or rotated hook token files now revoke old tokens immediately instead of at gateway restart.

## Test Plan

- `go test ./internal/gateway -run 'TestHookScopedToken' -count=1`\n  - Result: `ok github.com/defenseclaw/defenseclaw/internal/gateway 1.043s`\n\n## CI Status\n\nPending GitHub Actions for this PR.\n


sanjayab23:
Deferring to a follow-up PR — this fix touches `internal/gateway/api.go` and `internal/gateway/gateway_test.go`, which are gateway-side changes outside the scope of #410 (packaging + tests only). The finding is real and worth a dedicated review focused on the token-cache invalidation semantics.

Suggest retargeting #433 to `main` directly. #410 will not carry gateway changes.

PR #435, opened 2026-07-03
fix(macos): run userspace prep as target user

vineethsai7:
## Problem

PR #410 hardened pre-existing symlinked macOS userspace hook config paths, but the real installer still called `prepare_userspace_for` from the root shell. The helper checks for symlinks before writing, then opens `~/.codex/config.toml`, `~/.claude/settings.json`, or `~/.cursor/hooks.json` with ordinary shell redirection. A target user could race the config path after the check and before the root-owned open, causing a root write through a user-controlled symlink.

Related finding/comment: https://github.com/cisco-ai-defense/defenseclaw/pull/410#issuecomment-4873658536

## Current Situation

`install.sh` runs as root under `sudo`, and the userspace prep files are intended to be target-user-owned placeholders only. The secure invariant is that root should not open target-user-controlled config paths for writing.

## Solution

Run the userspace prep helper through `sudo -u <target-user>` with `HOME` pinned to the target home. Add a small `installer_lib.sh prepare-userspace` subcommand so the installer does not need inline shell evaluation and the test suite can exercise the same dispatch path.

## What Changed (Where & How)

| Path | Change |
| --- | --- |
| `packaging/macos/install.sh` | Calls `prepare_userspace_for_target`, which invokes the library helper as the target user. |
| `packaging/macos/lib/installer_lib.sh` | Adds a minimal `prepare-userspace` subcommand for non-sourced execution. |
| `packaging/macos/tests/test_prepare_userspace.sh` | Adds coverage for the library subcommand dispatch used by the installer. |

## Breaking Changes

None intended. The placeholder files are still created with the same content and modes; they are now created directly by the target user instead of root followed by `chown`.

## Open Issues / Follow-ups

None.

## Test Plan

```bash
packaging/macos/tests/run_tests.sh packaging/macos/tests/test_prepare_userspace.sh
bash -n packaging/macos/install.sh packaging/macos/lib/installer_lib.sh packaging/macos/uninstall.sh
git diff --check
make packaging-macos-test
```

Observed results:

- `packaging/macos/tests/run_tests.sh packaging/macos/tests/test_prepare_userspace.sh`: 11 passed, 0 failed.
- `bash -n packaging/macos/install.sh packaging/macos/lib/installer_lib.sh packaging/macos/uninstall.sh`: passed.
- `git diff --check`: passed.
- `make packaging-macos-test`: failed in unrelated existing host-dependent case `test_discover_agent_version.sh :: codex without install` because this runner has Codex installed and discovery returned `0.136.0` where the test expects empty output.

## CI Status

### DIFF
--- a/packaging/macos/install.sh
+++ b/packaging/macos/install.sh
@@ -55,6 +55,12 @@ die()  { printf '[install] ERROR: %s\n' "$*" >&2; exit 1; }
 # shellcheck source=lib/installer_lib.sh
 . "${SCRIPT_DIR}/lib/installer_lib.sh"
 
+prepare_userspace_for_target() {
+  local connector="$1"
+  sudo -H -u "${TARGET_USER}" env HOME="${TARGET_HOME}" \
+    /bin/bash "${SCRIPT_DIR}/lib/installer_lib.sh" prepare-userspace "${connector}" "${TARGET_HOME}"
+}
+
 usage() {
   cat <<EOF
 Usage: sudo $0 [options]
@@ -260,7 +266,7 @@ if [[ "${SKIP_CONNECTOR}" != "true" ]]; then
     fi
 
     log "  [${c}] preparing userspace"
-    if ! prepare_userspace_for "${c}" "${TARGET_HOME}"; then
+    if ! prepare_userspace_for_target "${c}"; then
       warn "  [${c}] refused to prepare userspace (symlinked or non-dir path); skipping"
       continue
     fi
--- a/packaging/macos/lib/installer_lib.sh
+++ b/packaging/macos/lib/installer_lib.sh
@@ -213,6 +213,19 @@ prepare_userspace_for() {
   esac
 }
 
+if [[ "${BASH_SOURCE[0]}" == "$0" ]]; then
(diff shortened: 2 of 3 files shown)


vineethsai7:
Security review finding: the new executable userspace-prep dispatch masks rejection failures.

`/bin/bash packaging/macos/lib/installer_lib.sh prepare-userspace cursor "$home"` returns `0` even when `prepare_userspace_for` rejects a symlinked connector path, because the script continues after the helper call instead of exiting with its status. The root installer uses this subprocess result to decide whether to skip unsafe userspace wiring, so the intended symlink/non-dir guard is not reliably enforced through the new target-user path.

Repro from this PR head:

```bash
tmp=$(mktemp -d); target=$(mktemp -d)
ln -s "$target" "$tmp/.cursor"
/bin/bash packaging/macos/lib/installer_lib.sh prepare-userspace cursor "$tmp"
echo $?
```

Observed: `0`; expected: non-zero.

Recommended fix: `exit $?` immediately after `prepare_userspace_for "$@"` in the dispatch block, plus a regression test for subcommand symlink rejection. I opened #438 with that narrow patch.

PR #439, opened 2026-07-06
fix(openclaw): guard ChatGPT Codex response passthrough

### LINKED ISSUE #434: OpenClaw connector: guardrail chat-proxy breaks agents on the ChatGPT/Codex OAuth backend (`chatgpt.com/backend-api/codex/responses`)
## Summary

Enabling the guardrail on an **OpenClaw** connector install (`defenseclaw init --connector openclaw --profile observe`, guardrail left in default `observe` mode) breaks any OpenClaw agent whose model is served through the ChatGPT/Codex OAuth backend (`https://chatgpt.com/backend-api/codex/responses`, `api=openai-chatgpt-responses`). The agent starts failing every request with `Unknown model: openai/<model>` / `model_not_found` — a local OpenClaw-side rejection that happens before any network call is logged, i.e. before the request ever reaches the guardrail proxy on `:4000`.

Disabling the guardrail (`defenseclaw guardrail disable --yes`) immediately restores normal operation, confirming the fetch-interceptor is the cause.

## Environment

- defenseclaw 0.8.3 (gateway + CLI), installed via `scripts/install.sh --connector openclaw`
- OpenClaw 2026.6.11, Linux x86_64
- Agent model: `openai/gpt-5.5` via the ChatGPT/Codex OAuth backend (not a static `OPENAI_API_KEY`)
- Guardrail: `observe` mode only — the failure reproduces with zero blocking/action policy active, purely from the proxy/interceptor being wired in

## What I found while digging in

`extensions/defenseclaw/dist/providers.json` does list this backend:

```json
{ "name": "openai-codex", "domains": ["chatgpt.com/backend-api"], "profile_id": null, "env_keys": [] }
```

and `fetch-interceptor.js` includes `/backend-api/codex/responses` in `LLM_PATH_SUFFIXES`, so the interceptor is clearly meant to recognize this traffic shape. But `openai-codex` is the *only* entry in `providers.json` with `profile_id: null` and `env_keys: []` — every other provider (anthropic, openai, azure, gemini, bedrock, etc.) has real credential-resolution wiring. There's also no test coverage anywhere under `dist/__tests__` that exercises the codex/chatgpt.com path.

This lines up with what #265 already documented for the **Codex** and **Claude Code** connectors: the HTTP chat-proxy surface has an "auth path collision" with ChatGPT/Bedrock OAuth flows, which is exactly why that PR deleted the proxy path for those two connectors in favor of hook-only enforcement. Per `CONNECTOR-MATRIX.md`, **OpenClaw remains proxy-only** (`llm_traffic_mode: proxy`) with no hook-only fallback, so an OpenClaw agent sitting on the same ChatGPT OAuth backend hits the identical collision that #265 fixed for the other two connectors — just with no alternative enforcement path to fall back to.

## Suggested fixes (any of these would unblock us)

1. Wire real OAuth-aware passthrough for `openai-codex` in the guardrail proxy (mirroring whatever Codex's hook path now does for auth), or
2. Explicitly exclude/passthrough `chatgpt.com/backend-api/*` from the OpenClaw/ZeptoClaw fetch-interceptor until (1) is done, so agents on this backend degrade to "unguarded" instead of "broken", or
3. Extend hook-only enforcement (from #265) to the OpenClaw connector as an alternative to the chat-proxy surface.

Happy to provide full logs/config (redacted) if useful.

fengjikui:
Fixes #434.

## Problem
OpenClaw can send model traffic through the ChatGPT/Codex OAuth backend at `https://chatgpt.com/backend-api/codex/responses`. The local guardrail proxy does not yet have OAuth-aware support for this backend, so routing that traffic through the proxy can break observe-mode OpenClaw installs.

## Safety boundary
`/backend-api/codex/responses` carries prompt/completion traffic, so it must not silently bypass DefenseClaw policy. By default, this PR keeps that endpoint and its subpaths on the guardrail path for both `fetch` and `https.request`.

## Solution
Add an explicit unguarded escape hatch, `DEFENSECLAW_UNGUARDED_CHATGPT_CODEX_RESPONSES=1`, for operators who knowingly prefer availability over inspection while OAuth-aware proxy support is missing. When enabled, the interceptor emits a visible warning and records the passthrough egress reason. Without this opt-in, ChatGPT/Codex response traffic is proxied.

## What Changed
| Area | Change |
| --- | --- |
| URL classification | Limits ChatGPT/Codex response passthrough to explicit `DEFENSECLAW_UNGUARDED_CHATGPT_CODEX_RESPONSES=1` opt-in. |
| Fetch interception | Keeps `fetch` calls to `/backend-api/codex/responses` on the guardrail path by default. |
| HTTPS interception | Keeps `https.request` calls to `/backend-api/codex/responses` on the guardrail path by default. |
| Warning/telemetry | Warns visibly when the unguarded opt-in is enabled and preserves the shared passthrough egress reason. |
| Tests | Adds negative coverage proving Codex response requests do not bypass by default, plus opt-in coverage for the escape hatch. |

## Validation
- `npm test -- --run src/__tests__/fetch-interceptor-hardening.test.ts src/__tests__/fetch-interceptor-https.test.ts`
- `npx tsc --noEmit`
- `git diff --check`


vineethsai7 (changes requested):
Security finding: the new passthrough should not include `/backend-api/codex/responses` without another inspection path. The matcher runs before both known-provider and body-shape interception, and the added regression test sends `{ model, input }` to `https://chatgpt.com/backend-api/codex/responses`. That means prompt/response traffic using this backend now reaches ChatGPT directly, with only an egress event, so DefenseClaw observe/action policies no longer inspect or block that path while users still appear to have the guardrail enabled.

Please keep actual Codex response/model calls on the guardrail path, or gate this as an explicit unprotected mode with user-visible warning/fail-closed behavior for enforcement mode. If a narrow OAuth/session passthrough is still required, match only those non-LLM auth/session endpoints and add negative tests proving `/backend-api/codex/responses` with `input`/`messages` is not bypassed.

fengjikui:
Thanks for the security review. Addressed in `55d448f`:

- By default, `https://chatgpt.com/backend-api/codex/responses` and subpaths no longer match the early passthrough branch.
- Added negative coverage proving both `fetch` and `https.request` route Codex response traffic to the guardrail proxy by default.
- Kept availability fallback only as an explicit unguarded opt-in via `DEFENSECLAW_UNGUARDED_CHATGPT_CODEX_RESPONSES=1`; when enabled it emits a visible `console.warn` and preserves the passthrough egress reason.

Validation run locally:
- `npm test -- --run src/__tests__/fetch-interceptor-hardening.test.ts src/__tests__/fetch-interceptor-https.test.ts`
- `npx tsc --noEmit`
- `git diff --check`

If you prefer no unguarded escape hatch at all, I can remove the env-gated path and keep this strictly fail-closed/default-proxied.

fengjikui:
Pushed `b71db03` to address the CI env-var registry failure. The unguarded Codex response fallback is now declared in `internal/envvars/registry.json`, generated env-var docs are updated, and the exported TypeScript constant was renamed so the repo-wide `DEFENSECLAW_*` audit only sees the runtime env-var name, not the wrapper symbol.

Local validation:
- `.venv/bin/python -m pytest cli/tests/test_envvars_codebase_coverage.py -q`
- `go test ./internal/envvars`
- `npm test -- --run src/__tests__/fetch-interceptor-hardening.test.ts src/__tests__/fetch-interceptor-https.test.ts`
- `npx tsc --noEmit`
- `.venv/bin/python -m pytest cli/tests/tui/test_catalog_action_bars.py -q` (the other CI failure passed locally: 23 passed, 1 skipped)
- `git diff --check`

PR #454, opened 2026-07-08
fix(macos): preserve guardian auth ownership

### LINKED ISSUE #453: macOS installer leaves guardian auth state root-owned after hook migration
## Summary

The merged macOS managed-enterprise installer re-chowns `RUNTIME_DIR` after running per-user `enterprise hooks install`, but it does not re-chown `GUARDIAN_AUTH_DIR`.

## Severity

Medium. A successful install can leave hook guardian authorization state owned by `root`, while the LaunchDaemon runs as the DefenseClaw service user. That can break later runtime writes or updates to guardian authorization state after the daemon restarts.

## Affected Files

- `packaging/macos/install.sh`
- `packaging/macos/tests/test_packaging_assets.sh`

## Context / Reproduction

1. Run the macOS installer with per-user hook wiring enabled.
2. `install.sh` invokes `enterprise hooks install` as root with `DEFENSECLAW_HOOK_GUARDIAN_AUTH_DIR="$GUARDIAN_AUTH_DIR"`.
3. After the CLI migration, the installer runs:

```bash
chown -R "${SERVICE_UID}:${SERVICE_GID}" "${RUNTIME_DIR}"
```

4. Any files created under `GUARDIAN_AUTH_DIR` by the root-owned CLI invocation remain root-owned.
5. The daemon then runs as `${SERVICE_USER}` and may fail to update that state.

## Recommended Fix

Include `GUARDIAN_AUTH_DIR` in the post-migration ownership repair and add a packaging regression test that fails if the re-chown command does not include both runtime-owned directories.

## Source

Found during the Daily DefenseClaw PR Security Review for July 7, 2026, while reviewing merged PR #440.

vineethsai7:
## Problem

Fixes #453.

The macOS managed-enterprise installer runs `enterprise hooks install` as root and passes `DEFENSECLAW_HOOK_GUARDIAN_AUTH_DIR` to the CLI. Files created under that guardian auth directory can therefore inherit root ownership, but the LaunchDaemon resumes as the DefenseClaw service user.

## Current Situation

After the root-owned hook migration, `install.sh` repairs ownership for `RUNTIME_DIR` only. That covers audit DB files, but leaves guardian auth state out of the post-migration repair even though it is written by the same CLI invocation path.

## Solution

Include `GUARDIAN_AUTH_DIR` in the post-CLI `chown -R` repair so both runtime state directories are writable by the service user before launchd resumes the daemon.

## Architecture Diagram

N/A

## What Changed (Where & How)

| File | Summary |
| --- | --- |
| `packaging/macos/install.sh` | Re-chowns `GUARDIAN_AUTH_DIR` together with `RUNTIME_DIR` after per-user hook migration. |
| `packaging/macos/tests/test_packaging_assets.sh` | Adds a static regression test that requires the post-migration ownership repair to include guardian auth state. |

## Breaking Changes

None.

## Open Issues / Follow-ups

None.

## Test Plan

- [x] `packaging/macos/tests/run_tests.sh` — observed `84 passed, 0 failed in 3s`.
- [x] `git diff --check` — observed no output / exit 0.

### DIFF
--- a/packaging/macos/install.sh
+++ b/packaging/macos/install.sh
@@ -812,10 +812,11 @@ if [[ "${SKIP_CONNECTOR}" != "true" ]]; then
   # inherit the running uid/gid (root:defenseclaw) with mode 0644 — so
   # the daemon (running as defenseclaw) can READ but not WRITE the audit
   # DB when it comes back up. Result: every hook evaluation runs, but
-  # the audit row silently drops. Fix by re-chowning after every CLI
-  # invocation completes.
+  # the audit row silently drops. Guardian auth state is written by the
+  # same root-owned CLI invocation, so repair it before the service user
+  # resumes too.
   log "  re-chowning runtime files created by CLI migrations"
-  chown -R "${SERVICE_UID}:${SERVICE_GID}" "${RUNTIME_DIR}"
+  chown -R "${SERVICE_UID}:${SERVICE_GID}" "${RUNTIME_DIR}" "${GUARDIAN_AUTH_DIR}"
 
   log "  resuming LaunchDaemon"
   launchctl bootstrap system "${PLIST_DST}"
--- a/packaging/macos/tests/test_packaging_assets.sh
+++ b/packaging/macos/tests/test_packaging_assets.sh
@@ -162,6 +162,35 @@ ${hints}"
   fi
 }
 
+t_install_rechowns_guardian_auth_dir_after_cli() {
+  local bad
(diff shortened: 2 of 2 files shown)


vineethsai7:
Closing this PR after re-validating the macOS guardian authorization trust model. The proposed recursive chown would make protected_targets.json service-user-owned, and managed.ValidateTrustedFilePath requires the guardian authorization path and file to remain root-owned. The CLI already calls setEnterpriseHookAuthorizationOwnership on the authorization directory and file to keep owner root while setting the service group for readability, so the original issue is not valid.

PR #472, opened 2026-07-10
docs: add spec for allowing private upstream IPs (001-allow-private-u…

nghodkicisco:
…pstreams)

Adds requirements, design, plan, and tasks for a new feature that lets operators allowlist specific private IP addresses to bypass DefenseClaw's SSRF protection for on-prem LLM gateways.

Key decisions:
- IP list only (no CIDR) — forces explicit per-endpoint trust
- Config.yaml field + env var (DEFENSECLAW_ALLOW_PRIVATE_UPSTREAMS)
- Hardcoded deny for loopback/link-local/metadata (Ollama has its own dedicated bypass that remains unchanged)
- Applied at both pre-flight and dial-time checks
- Audit event emitted for SOC visibility

## What Changed (Where & How)

| File | Summary |
| --- | --- |
|  |  |


vineethsai7:
Security review finding: private-upstream redirects can pivot an authenticated request into an allowlisted private IP.

`providerHTTPClient` uses the guarded `DialContext`, but it does not set `CheckRedirect`. Once an operator allowlists a private IP, `isUnsafeIP` permits that IP globally. A public upstream selected via `X-DC-Target-URL` can return `307`/`308 Location: http://<allowlisted-private-ip>/...`; `net/http` follows the redirect, and the dial guard allows the private peer.

Repro: configure `guardrail.allow_private_upstreams: ["10.50.2.100"]`, send an authenticated LLM-shaped request to an attacker-controlled public upstream, and return `307 Location: http://10.50.2.100/v1/chat/completions`. The request can reach the allowlisted private address instead of staying bound to the original upstream.

Recommended fix: set `providerHTTPClient.CheckRedirect` to `netguard.BlockRedirects`, or revalidate each redirect against an origin-bound policy rather than the global private-IP allowlist.

vineethsai7 (approved):

PR #506, opened 2026-07-13
fix(windows): scope Codex and Claude OTLP authentication

vineethsai7:
## Summary
- provision distinct path-scoped OTLP tokens for Codex and Claude Code
- stop exposing the master gateway token in agent telemetry configuration
- recognize owned scoped endpoints during repair and teardown
- test cross-connector isolation and exact endpoint ownership

## Validation
- focused Codex, Claude, Native OTLP, and token connector tests pass
- focused gateway OTLP authentication tests pass
- full packages expose only pre-existing Windows-incompatible OpenClaw/ACL test failures

PR #512, opened 2026-07-13
fix(windows): reject writable OTLP token ACLs

vineethsai7:
## What

- Enforce write-capable NTFS DACL checks when loading the Windows OTLP path token.
- Reject untrusted write and append rights, including compound, object, callback, and callback-object allow ACEs.
- Preserve read-only untrusted access and keep the POSIX wide-mode test platform-specific.

## Why

The token loader reused the Windows secure-file validator with child write-mask protection disabled. That omitted file-level write masks, so an untrusted local principal could replace the scoped OTLP token and inject or deny telemetry. This does not grant DefenseClaw master API authority, but it violates the token-file trust boundary.

Security invariant: only trusted principals may write the OTLP token file; read-only nontrusted ACEs remain valid.

## Stack

- Based on #506 (`agent/windows-otlp-auth`).
- Review this PR as the next isolated Windows security fix in the stack.

## Checks

- `go test ./internal/gateway/connector -run 'TestLoadOTLPPathToken_(RejectsUnsafeFiles|AcceptsStrictTokenFile)|TestLoadOTLPPathTokenWindows|TestOTLPPathTokenWindowsRejectsUnsupportedWriteAllowACETypes' -count=1`
- `go test ./internal/gateway/connector -run 'OTLP|HookAPITokenWindows' -count=1`
- `GOOS=linux GOARCH=amd64 go test -c ./internal/gateway/connector`
- `git diff --check`


vineethsai7:
@coderabbitai review

vineethsai7:
@coderabbitai review

vineethsai7:
@coderabbitai review

vineethsai7:
@coderabbitai review

vineethsai7:
@coderabbitai review

PR #513, opened 2026-07-13
fix(connectors): bind backups to captured paths

vineethsai7:
## What

- Bind managed connector backups to the exact normalized target path captured during setup.
- Validate connector/logical identity and target-path equality during recapture, post-hash updates, and restore.
- Preserve Windows case-insensitive path semantics without resolving symlinks.
- Add exploit regressions proving a same-hash second config cannot receive another config's pristine bytes or consume its recovery metadata.

## Why

A backup captured for config A was keyed only by connector/logical name. If `CODEX_HOME`, a Claude config location, or another connector target changed to B, a later setup/update/teardown could reuse A's backup against B. When B happened to match the managed hash, restore could overwrite or delete B and remove A's only recovery metadata.

Invariant: restore metadata may read, hash, write, delete, or consume recovery state only for the connector identity and exact lexical target captured with that backup.

## Stack

- Based on #512 (`agent/windows-otlp-token-acl`).
- This is the next isolated lifecycle-safety fix in the Windows stack.

## Checks

- Pre-fix regression reproduced all three unsafe paths: recapture, post-hash update, and restore.
- `go test ./internal/gateway/connector -run '^TestManagedFileBackup_' -count=1`
- Focused Claude and Codex exact-restore tests.
- Focused hook-only backed-up-path and OmniGent target-migration tests.
- Linux connector test binary cross-compilation (`GOOS=linux GOARCH=amd64 go test -c`).
- `git diff --check`

The package-wide Windows connector run still contains existing platform-rejection OpenClaw cases and intermittent NTFS/AV `Access is denied` rename flakes; those are separate from this path-binding change and are being tracked in the Windows stack.


vineethsai7:
@coderabbitai review

vineethsai7:
@coderabbitai review

vineethsai7:
@coderabbitai review

vineethsai7:
@coderabbitai review

PR #515, opened 2026-07-13
fix(windows): retry connector file replacement

vineethsai7:
## Summary
- route connector temp-file replacement through the existing Windows-aware safefile replacement helper
- retry only Win32 access-denied, sharing-violation, and lock-violation failures within a 495 ms bound
- add deterministic fault-injection coverage for transient, permanent, bounded, and immediate-success paths

## Why
Native connector refresh and teardown intermittently failed while replacing protected state such as `claudecode_backup.json` because antivirus/indexing opens can make `os.Rename` briefly return `Access is denied`.

## Verification
- deterministic safefile retry suite passes
- focused Claude refresh/teardown suite passes 10 consecutive native Windows runs
- broader Claude Setup/Teardown and Codex setup suite passes 3 consecutive runs
- full safefile package passes
- Linux connector and safefile test binaries cross-compile

## Stack
Base: #514

PR #516, opened 2026-07-13
fix(claude): honor registered hook deadlines

vineethsai7:
## Summary
- derive the native Claude hook event from `hook_event_name` when `--event` is omitted
- share one 10/30/60/90-second timeout source between Claude registration and the native executor
- reserve one second for fail-open/fail-closed response emission, yielding 9/29/59/89-second HTTP budgets
- use the shortest safe 9-second budget for missing or malformed event payloads

## Why
Claude registered long-running lifecycle hooks for up to 60 or 90 seconds, but the native HTTP client hard-capped every event at 10 seconds. Stop-time scans and batch events could therefore fail before their documented agent deadline.

## Verification
- pre-fix mapping tests did not compile because event-derived budgets were absent
- full native hookexec suite passes
- Claude registration and Windows shell-free setup tests pass
- Codex VerifyClean/native config matrix passes 3 consecutive runs on the final stack
- Linux connector and hookexec test binaries cross-compile

## Stack
Base: #515


vineethsai7:
@coderabbitai review

vineethsai7:
@coderabbitai review

vineethsai7:
@coderabbitai review

PR #518, opened 2026-07-13
fix(claude): preserve operator env changes on teardown

vineethsai7:
## Summary

- record the exact Claude `settings.json` env values written by the most recent DefenseClaw setup
- restore or delete a managed env key only when its current value still matches DefenseClaw's last write
- preserve operator edits, removals, non-string replacements, and unrelated env keys during teardown
- verify managed env residue independently of a missing or malformed `hooks` subtree
- cover every Claude env key emitted by Setup, including raw prompt telemetry mode

## Root cause

The backup captured only the pre-install env block. Teardown therefore removed every known key unconditionally and could not distinguish DefenseClaw-owned values from operator changes made after Setup. `VerifyClean` also nested its env inspection under successful `hooks` parsing and recognized only a subset of emitted keys.

## Impact

Windows uninstall and connector switching no longer clobber post-install Claude environment customization. Clean verification still detects exact DefenseClaw residue without treating operator-edited values that retain DefenseClaw-like markers as owned.

## Stack

- Stacked on #517 (`agent/windows-claude-outcome-semantics`)

## Validation

- `go test ./internal/gateway/connector -run '^TestClaudeCode_' -count=1`
- `go test ./internal/gateway/connector/... -count=1`
- `go vet ./internal/gateway/connector/...`
- `git diff --check origin/agent/windows-claude-outcome-semantics...HEAD`


vineethsai7:
@coderabbitai review

vineethsai7:
@coderabbitai review

vineethsai7:
@coderabbitai review

vineethsai7:
@coderabbitai review

PR #536, opened 2026-07-15
fix(windows): repair TUI tab navigation and alert filters

vineethsai7:
## What changed

- route `Tab` / `Shift+Tab` panel navigation before Textual focus traversal can consume it
- preserve Setup form/menu/config Tab handling by normalizing the logical key before `event.character`
- refresh the audit store when Alerts toolbar selects All, Medium, or Low so mouse and keyboard filters show the same rows
- add Textual and model regression coverage for global and Setup-local Tab behavior plus low-severity hydration

## Why

Native Windows live TUI testing of the exact PR 534 installer reproduced two parity defects: advertised Tab navigation moved focus instead of panels, and the Alerts toolbar reused the actionable-only cache while keyboard filters refreshed the full audit store.

## Validation

- affected TUI suites: 212 passed, 1 skipped
- focused key/filter regressions: 4 passed
- Ruff and `git diff --check`: passed
- exact commit `361334776fc89508a004daaf23a91812f52082db` built into and installed from the unsigned local x64 Setup artifact (`SHA-256 557469F34718FF01C25B6489D53FF541F7C7C5BB77FECCCF9D797C02FD575EC1`)
- packaged installed-runtime TUI: global Tab/Shift+Tab switched panels, Setup focus moved `0 -> 1 -> 0`, and Alerts All hydrated 156 rows containing both Codex and Claude activity
- installed CLI command tree: 212/212 help paths passed; seven representative live JSON reads passed without user/config file drift
- real Codex CLI 0.144.3 and Claude Code 2.1.207: benign commands executed; destructive `Remove-Item -Recurse -Force` commands were blocked before execution with `enforced=true` and targets preserved
- Codex live run used persisted hook trust and did not use `--dangerously-bypass-hook-trust`
- installed Doctor: 41 passed, 0 failed, 11 skipped
- PR checks: 39 passed, 2 expected credentialed live-matrix skips; CodeRabbit approved with no actionable comments

This PR is intentionally stacked on PR 534.

### DIFF
--- a/cli/defenseclaw/tui/app.py
+++ b/cli/defenseclaw/tui/app.py
@@ -1743,6 +1743,19 @@ class DefenseClawTUI(App[None]):
             event.prevent_default()
             return
 
+        # Textual treats Tab/Shift+Tab as focus traversal before ordinary
+        # bindings can reliably route them.  Handle the advertised panel
+        # navigation here, after command-palette and panel-local handlers have
+        # had first refusal (Setup forms, for example, use Tab between fields).
+        if event.key in {"tab", "shift+tab"}:
+            if event.key == "tab":
+                self.action_next_panel()
+            else:
+                self.action_previous_panel()
+            event.stop()
+            event.prevent_default()
+            return
+
         panel = PANEL_SHORTCUTS.get(event.key.lower())
         if panel is None:
             return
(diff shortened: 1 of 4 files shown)


vineethsai7:
@coderabbitai review

PR #541, opened 2026-07-16
feat(packaging): wire hooks for every local user via the hook-guardian

cjsanjay:
## Summary

- Fixes a silent regression in the customer-shipping macOS `.pkg` where **no user's hooks ever got wired** (installd runs as root, not sudo, so `$SUDO_USER` is empty and the single-user inline call to `enterprise hooks install` short-circuited).
- Replaces the single-user inline flow with a machine-wide `targets.yaml` reconciled by the already-shipped `hook-guardian` LaunchDaemon (5-min tick).
- Adds a companion `hook-enumerator` LaunchDaemon (also 5-min tick) that re-renders `targets.yaml` so users provisioned **after** the pkg install are picked up automatically.
- Fixes a second latent bug: `install.sh` referenced the guardian plist for collision preflight but never actually installed or bootstrapped it. It does now, alongside the new enumerator.
- Multi-user cleanup in `uninstall.sh` (iterates every eligible local user for the `~/.defenseclaw` + agent-config scrub).

## Why

The pre-2026.7.3 flow only wired hooks for `$SUDO_USER`. Under a pkg postinstall that variable is empty, so shipped Macs had:

- Daemon running, telemetry emitting ✓
- `protected_targets.json` never written ✗
- `~/.codex/config.toml` / `~/.claude/settings.json` / `~/.cursor/hooks.json` all empty ✗
- **Zero policy enforcement.**

## What changed

**Design.** The `hook-guardian` LaunchDaemon (`enterprise hooks reconcile --manifest ...`) already re-parses `targets.yaml` on every 5-min tick via `LoadManifest` (`internal/enterprisehooks/manifest.go:33`). The merge writer in `enterprise_hooks.go:674-700` keys by `(connector, user)`, so writing one row per (user × connector) is safe and idempotent. The manifest becomes the single source of truth; install.sh's job is now to write the initial file + start the daemons.

**Files modified.**

- `packaging/macos/install.sh` — new: enumerate local users, render `${SUPPORT_DIR}/hook-guardian/targets.yaml`, install + bootstrap guardian + enumerator, kickstart both for immediate first pass. Removed: inline `enterprise hooks install --user <one>` loop.
- `packaging/macos/lib/installer_lib.sh` — new helpers `enumerate_local_users` (UID ≥ 500, non-`_`-prefixed, home under `/Users/`, `home_perms_ok`) and `render_targets_manifest` (emits schema-compliant YAML matching `ManifestTarget`).
- `packaging/macos/uninstall.sh` — iterates every eligible local user for the scrub + `.defenseclaw` cleanup; boots out the new daemons; keeps `--user X` as a scoped-cleanup override.
- `scripts/build-macos-bundle.sh` — bundles the new plist + helper.
- `packaging/macos/tests/test_packaging_assets.sh` — 4 new asserts covering: guardian + enumerator plists exist and lint, `render-targets.sh` shipped + executable, install.sh actually copies + bootstraps both daemons, no more inline `enterprise hooks install` invocations.
- `packaging/macos/tests/test_render_targets_manifest.sh` (new) — 6 cases covering multi-user × multi-connector cross-product, unsupported-connector filter, empty inputs, YAML round-trip via PyYAML.

**Files added.**

- `packaging/launchd/com.cisco.secureclient.defenseclaw.hook-enumerator.plist` — RunAtLoad + StartInterval 300, invokes `lib/render-targets.sh` as root.
- `packaging/macos/lib/render-targets.sh` — reads `config.yaml` for the connector list, calls `enumerate_local_users`, atomically swaps `targets.yaml` if content changed. Logs a no-op on unchanged input so the guardian doesn't churn.

**No Go changes.** `enterprise hooks reconcile` already re-parses the manifest per tick; the sidecar already polls `protected_targets.json` every 5s (`sidecar.go:2612-2621`). Verified end-to-end by round-tripping the shell-rendered YAML through `LoadManifest` in a smoke test.

## Test plan

- [x] `packaging/macos/tests/run_tests.sh` — 107 passed, 2 pre-existing env failures unrelated to this change (my dev box has `/opt/cisco/secureclient/defenseclaw/` already installed, which trips the fresh-host preflight in two bundle-fixture tests; same failures reproduce at HEAD on this box).
- [x] `go test ./internal/enterprisehooks/... ./internal/cli/...` — pass.
- [x] `go build ./...` — clean.
- [x] `plutil -lint` on the new hook-enumerator plist — OK.
- [x] Round-trip smoke: rendered `targets.yaml` from the shell helper loads cleanly through Go's `LoadManifest` with UID/GID typed as `*int`.
- [ ] End-to-end on a fresh customer Mac: pkg install → `sudo cat /opt/cisco/secureclient/defenseclaw/hook-guardian/targets.yaml` shows every local user × requested connector → `sudo cat /opt/cisco/secureclient/defenseclaw/hook-guardian-state/protected_targets.json` matches the enumerated user set → per-user `~/.<agent>/*` configs contain `defenseclaw` hook entries.
- [ ] Future-user smoke: `sudo sysadminctl -addUser testfuture -password test -home /Users/testfuture` → within ≤10 min, `testfuture` appears in `protected_targets.json` without any manual bootstrap.

Verification playbook is in the plan doc under `.claude/plans/add-support-for-multi-user-abundant-kettle.md` (step 6, "Verification").

---

## Follow-up: fsnotify auto-heal with self-write suppression (06c13469)

**What.** Switched the hook-guardian LaunchDaemon from one-shot periodic `enterprise hooks reconcile` (every 5 min) to a long-running `enterprise hooks watch` process that reacts to per-user tampering in ~1 s via fsnotify, plus a **60 s** periodic backstop for the vectors the fsnotify path intentionally can't catch.

**Why.** The 5-min tick meant any tamper (delete config, neuter hook script, corrupt token) won for up to 5 minutes. An endpoint security product cannot ship with an "eventually consistent to 5 min" SLA against user tampering.

**Design.** Three layered filters keep the fsnotify path both responsive and quiet:

1. **Ownership filter** — `WatchOwnedFiles` returns two sets:
   - `ExclusiveWriter` (hook scripts, `.hook-<connector>.token`, backup files): only DefenseClaw writes here → any event is a tamper signal.
   - `SharedWriter` (native agent configs — `~/.codex/config.toml`, `~/.claude/settings.json`, `~/.cursor/hooks.json`): the agent itself constantly rewrites these during normal use → only Remove/Rename triggers reconcile; Write/Chmod is deferred to the backstop.
   - Generic scripts (`inspect-*.sh`, `_hardening.sh`, `.hookcfg`) are excluded entirely: per-connector template rendering makes them rewrite-storm on every reconcile, so their events are pure noise.

2. **Post-reconcile settle window** — every reconcile stamps `settleUntil = now + 2s`; events inside the window are suppressed as the guardian's own tail. Remove/Rename inside the window is suppressed **only if the file still exists on disk** (guardian's `os.Rename` is atomic; macOS surfaces the intermediate REMOVE but the file is still there). If a Stat shows the file truly missing, it's a real user `rm` and falls through. Deterministic distinction between "our own atomic-write tail" and "back-to-back user tamper."

3. **Row-set content hash** — after each reconcile, hash the target row set. Byte-identical to previous run → widen the settle window to `3× settle` to absorb macOS FSEvents chmod-batching latency. Real tamper always flips the hash, so widening never masks a genuine repair.

**Plist.** Switched to `watch --interval 60s` + `KeepAlive true`; dropped `StartInterval 300`. Backstop tightened 5 min → 60 s to reduce worst-case detection for SharedWriter Writes and generic-script tampers.

**End-to-end verification on macOS:**

- [x] `rm ~/.codex/config.toml ~/.defenseclaw/hooks/codex-hook.sh ~/.defenseclaw/hooks/.hook-codex.token` → all 3 restored in ~1 s, byte-identical to shipped content.
- [x] Back-to-back Removes 5 s apart → both heal ~1 s each (Stat-check on Remove inside settle window passes the second one through).
- [x] Write on ExclusiveWriter (overwrite `codex-hook.sh` with a stub) → reconcile restores in ~2 s.
- [x] Write on SharedWriter (append to `~/.codex/config.toml`) → NOT reconciled via fsnotify (as designed); caught by 60 s backstop.
- [x] Steady-state guardian log during heavy Codex/Claude Code/Cursor activity → quiet. No `no_change=true` churn, no `settled: suppressed N` spam.

**Tests.** New unit coverage:

- `TestEnterpriseHookWatchEventInSettleWindow` — 7 subtests (Write/Chmod/Remove/Rename inside/outside window, never-reconciled boundary).
- `TestEnterpriseHookReconcileRowsHash` — row-order stability, outcome change flips hash, target-add flips hash, empty-input invariant.
- `TestWatchOwnedFilesReturnsSpecificFilesNotDirs` — asserts ExclusiveWriter/SharedWriter split, forbids leaking agent-runtime paths, pins shared-across-connector artifacts out of both sets.
- Updated packaging: asserts `<string>watch</string>` + `<string>60s</string>` + `<key>KeepAlive</key>` present, `<string>reconcile</string>` + `<key>StartInterval</key>` absent.

Full test sweep: `116 passed, 0 failed` (packaging) + all Go unit tests green.

**Known residual bypass surface** (deliberately documented for follow-up, not addressed on this branch):

- **Rapid-rewrite LaunchAgent** (attacker rewriter loop at 100 ms cycle beats ~2 s guardian latency ~95% of wall-clock time). Requires moving enforcement out of user space (macOS Endpoint Security framework) to close. Not a regression from this PR — pre-existing surface.
- **SharedWriter Write** (e.g. `sed -i` stripping DefenseClaw entries from `config.toml`). 60 s-bounded via backstop. Accepted trade-off — trying to react in real time reintroduces the noise loop this PR fixed.
- **Symlink swap on hook script paths** — `atomicWriteFile` follows symlinks and the fsnotify watch is on the source path. Worth a follow-up refusing symlinks under `~/.defenseclaw/hooks/`. Not exploited today but worth closing.

### DIFF
--- a/cli/defenseclaw/gateway_error_codes.py
+++ b/cli/defenseclaw/gateway_error_codes.py
@@ -136,6 +136,10 @@ SUBSYSTEM_TELEMETRY: Final[str]       = "telemetry"
 SUBSYSTEM_CORRELATION: Final[str]     = "correlation"
 SUBSYSTEM_STREAM: Final[str]          = "stream"
 SUBSYSTEM_CISCO_INSPECT: Final[str]   = "cisco-inspect"
+# Managed Cisco AI Defense telemetry log-exporter path; paired with
+# ErrCodeExportFailed. Distinct from generic "telemetry" so operators
+# can alert specifically on managed-sink breakage.
+SUBSYSTEM_CISCO_AID_EXPORT: Final[str] = "cisco-aid-export"
 SUBSYSTEM_OPENSHELL: Final[str]       = "openshell"
 SUBSYSTEM_WEBHOOK: Final[str]         = "webhook"
 SUBSYSTEM_QUARANTINE: Final[str]      = "quarantine"
@@ -163,6 +167,7 @@ ALL_SUBSYSTEMS: Final[tuple[str, ...]] = (
     SUBSYSTEM_CORRELATION,
     SUBSYSTEM_STREAM,
     SUBSYSTEM_CISCO_INSPECT,
+    SUBSYSTEM_CISCO_AID_EXPORT,
     SUBSYSTEM_OPENSHELL,
     SUBSYSTEM_WEBHOOK,
     SUBSYSTEM_QUARANTINE,
--- a/cli/tests/test_enterprise_packaging.py
+++ b/cli/tests/test_enterprise_packaging.py
@@ -160,13 +160,28 @@ def test_launchd_hook_guardian_is_separate_privileged_job():
 
     assert payload["Label"] == "com.cisco.secureclient.defenseclaw.hook-guardian"
     assert "UserName" not in payload
-    assert payload["ProgramArguments"][1:4] == ["enterprise", "hooks", "reconcile"]
+    # Guardian runs the long-running `enterprise hooks watch` command, not
+    # the one-shot `reconcile` — fsnotify-driven auto-heal (~1 s) with a
+    # 60 s periodic backstop, restart-managed via KeepAlive rather than
+    # StartInterval. See internal/cli/enterprise_hooks.go runEnterpriseHooksWatch
+    # for the loop's design (settle window + Stat-based rename-tail detection).
+    assert payload["ProgramArguments"][1:4] == ["enterprise", "hooks", "watch"]
+    # --interval 60s is the periodic backstop for tamper vectors the fsnotify
+    # path intentionally cannot catch (SharedWriter Write/Chmod on native
+    # agent configs, shared-across-connector generic scripts). Any drift in
+    # this value should be a deliberate policy change, not an accidental edit.
+    assert "--interval" in payload["ProgramArguments"]
+    assert "60s" in payload["ProgramArguments"]
     assert payload["EnvironmentVariables"]["DEFENSECLAW_DEPLOYMENT_MODE"] == "managed_enterprise"
     assert (
         payload["EnvironmentVariables"]["DEFENSECLAW_HOOK_GUARDIAN_AUTH_DIR"]
         == "/opt/cisco/secureclient/defenseclaw/hook-guardian-state"
     )
-    assert payload["StartInterval"] == 300
+    # Long-running watch mode is kept alive by KeepAlive, NOT StartInterval.
+    # StartInterval would pointlessly relaunch the process every N seconds
+    # (and possibly spawn duplicates); KeepAlive relaunches only on exit.
+    assert "StartInterval" not in payload
+    assert payload.get("KeepAlive") is True
 
 
 def test_release_archives_ship_enterprise_packaging_assets():
@@ -316,7 +331,13 @@ def test_launchd_enterprise_installer_matches_cisco_plist_layout():
     home = gateway["EnvironmentVariables"]["DEFENSECLAW_HOME"]
     config = gateway["EnvironmentVariables"]["DEFENSECLAW_CONFIG"]
     auth_dir = gateway["EnvironmentVariables"]["DEFENSECLAW_HOOK_GUARDIAN_AUTH_DIR"]
-    manifest = guardian["ProgramArguments"][-1]
+    # The manifest path follows the --manifest flag; explicit lookup instead
+    # of positional indexing (ProgramArguments[-1] used to be the manifest
+    # under `hooks reconcile --manifest <path>`, but the current watch-mode
+    # args add `--interval 60s` after the manifest, making index -1 wrong).
+    guardian_args = guardian["ProgramArguments"]
+    manifest_flag = guardian_args.index("--manifest")
+    manifest = guardian_args[manifest_flag + 1]
 
     assert f"BINARY_ROOT={home}" in text
     assert f'CONFIG_DEST="{config}"' in text
--- a/docs/OBSERVABILITY-CONTRACT.md
+++ b/docs/OBSERVABILITY-CONTRACT.md
@@ -146,7 +146,7 @@ Complete enumeration (must match `AllErrorCodes()`):
 
 Complete enumeration (must match `AllSubsystems()`):
 
-`sidecar`, `watcher`, `gateway`, `scanner`, `policy`, `guardrail`, `auth`, `config`, `inspect`, `approvals`, `sink`, `telemetry`, `correlation`, `stream`, `cisco-inspect`, `openshell`, `webhook`, `quarantine`, `agent-registry`, `sqlite`, `admission`, `config_mutation`, `gatewaylog`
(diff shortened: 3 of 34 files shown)


vineethsai7:
Security review finding from the July 16 daily pass:

**High: root-trusted hook manifest can be YAML-injected from user-controlled agent metadata.** `discover_agent_version` returns raw values from user-writable package metadata, and `render_targets_manifest` interpolates `agent_version`, `user`, and `user_home` into double-quoted YAML without escaping. A local user can set a version containing a quote/newline sequence such as `"\n    enabled: false\n  - user: "victim"...`; the root enumerator then writes a trusted `targets.yaml` that `LoadManifest` accepts as multiple targets, letting user-controlled metadata reshape guardian reconciliation.

Affected path: `packaging/macos/lib/installer_lib.sh` through the install/enumerator flow.

Recommended fix: render the manifest with a YAML/JSON encoder or strict scalar escaping, reject control characters in version/user/home fields, and add hostile quote/newline round-trip tests that parse the rendered manifest with the Go `LoadManifest` path.

vineethsai7:
Follow-up: I pushed a narrow fix branch at `origin/codex/fix-pr541-manifest-yaml-escaping` with scalar escaping/control-character rejection and a hostile-version regression (`bash packaging/macos/tests/run_tests.sh` passed: 116 passed, 0 failed). I could not open a stacked PR directly into `feature/multi-user-hook-wiring-26.7.3` because that head branch is not a base branch in the upstream repository; opening against `release-defenseclaw-enterprise-26.7.3` would duplicate the whole #541 diff instead of being a narrow fix.

cjsanjay:
Thanks for the security find, @vineethsai7 — legitimate root-trusted YAML injection surface that would have shipped.

Cherry-picked your `571e20b9` (fix(packaging): escape hook manifest scalars) onto the head of this branch as `81129a05`. Preserved authorship. Passes locally:

- `go build ./...` clean
- `go test ./internal/enterprisehooks/... ./internal/ipc/... ./internal/gatewaylog/...` green
- `bash packaging/macos/tests/run_tests.sh` — **114 passed**, incl. your new `t_hostile_agent_version_cannot_inject_targets`
- `make check-error-codes` — parity OK

Also in this push (`2c990cbd`):
- **IPC toast filter** — drops per-rule signals (`Prompt Injection`, `PII`, ...) from the SecureClient notification body while preserving categories (`SECURITY_VIOLATION`, ...). Audit / OS toast paths still see the full detail. Regex is `^[A-Z][A-Z0-9]*(_[A-Z0-9]+)+$` (require an underscore — filters out `PII`/`PHI`/`PCI` acronyms too).
- **`gofmt` fix** on `internal/gatewaylog/error_codes.go` — Go Lint CI was failing on column-alignment drift from my earlier `SubsystemCiscoAIDExport` comment.
- **Go↔Python subsystem parity** — the earlier commit added `SubsystemCiscoAIDExport` in Go but not in `cli/defenseclaw/gateway_error_codes.py` or `AllSubsystems()`. `make check-error-codes` was failing. Fixed.
- **`Python Dependency Audit`** remains failing on 3 upstream `mcp` CVEs (1.27.2 / 1.28.1 fix them). Unrelated to this PR — separate follow-up.

vineethsai7 (approved):

PR #573, opened 2026-07-20
feat(guardrail): detect coding-agent sandbox-escape attempts

ihabler:
Add an "agent-escape" regex detection category targeting the write-then-execute class of sandbox escapes, where a sandboxed coding agent writes a file or config that a trusted, unsandboxed host component later executes.

Rules (each requires a sensitive target AND a dangerous payload token to co-occur, keeping false positives low):
- ESC-CLAUDE-HOOK / ESC-CLAUDE-HOOK-DIR: workspace .claude settings/hook that wires a shell command
- ESC-VENV-CFG: virtualenv interpreter redirect via pyvenv.cfg
- ESC-GIT-CONFIG-EXEC: git config execution-indirection keys (core.hooksPath/fsmonitor/sshCommand, textconv, diff.external, filter clean|smudge, uploadpack.packObjectsHook)
- ESC-GIT-READ-WEAPONIZED: read-only git verb with file-writing/exec flag
- ESC-DOCKER-SOCK: Docker daemon socket access (distinct from the PATH-DOCKER credential-file rule)

Registered in defaultRuleCategories and mirrored in the default, permissive, and strict rule packs. Adds true-positive, false-positive, block-boundary, and PATH-DOCKER-distinctness tests.

Related advisories: CVE-2026-48124, GHSA-pc9j-3qc2-95wv, GHSA-p9g2-cr55-cw9c, GHSA-v4xv-rqh3-w9mc.

### DIFF
--- a/internal/gateway/rules.go
+++ b/internal/gateway/rules.go
@@ -293,6 +293,68 @@ var trustExploitRules = []PatternRule{
 }
 
 // ---------------------------------------------------------------------------
+// Agent sandbox-escape rules
+//
+// Coding agents run in a sandbox, but they can still WRITE files and config
+// that a trusted, unsandboxed host component later EXECUTES. That write-then-
+// execute indirection is the escape: the agent never runs the payload itself,
+// it plants a future input for the IDE, the language extension, git, or a
+// local daemon to run outside the sandbox.
+//
+// Each rule requires a sensitive target AND a dangerous payload token to
+// co-occur. The whole tool-call args JSON — file_path plus content — is
+// scanned as one blob, so scaffolding a plain config file does not fire; only
+// the executable shape does. Patterns match barewords with bounded gaps
+// because the content is JSON-escaped (quotes arrive as \"). RE2 keeps the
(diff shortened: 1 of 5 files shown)


ihabler:
**Fixed**
- **ESC-GIT-READ-WEAPONIZED false positive on `format-patch`** — removed `format-patch` from the verb list. Writing files via `-o`/`--output` is its normal function and it executes no code, so it doesn't belong to the weaponized-read class. Added `git format-patch -o outgoing/ HEAD~3` and `--output-directory=` false-positive cases.
- **ESC-GIT-CONFIG-EXEC missing `filter.<driver>.process`** — added; same code-execution family as clean/smudge. New TP test `git config filter.lfs.process`.
- **Unicode-escape bypass (`hooks` / `command`)** — rather than complicate one regex, I decode JSON `\uXXXX` escapes in the shared `normalizeShell` pass, so *every* rule is hardened against this evasion via the existing second scan. New TP test with an escaped `.claude` hook.

PR #575, opened 2026-07-21
chore(macOS): import standalone app v1.1.7

### LINKED ISSUE #572: [macOS] Import latest standalone app release
The weekly freshness check found that the checked-in macOS app is not the latest stable standalone release.

Latest stable tag: `v1.1.7`

Update with:

```bash
scripts/update-macos-app.sh v1.1.7
```

Freshness run: https://github.com/cisco-ai-defense/defenseclaw/actions/runs/29753245582

vineethsai7:
## Problem

The weekly macOS Upstream Freshness run detected that the reviewed standalone app source was still pinned to v1.1.5 while v1.1.7 is the latest stable release. That intentionally failed the nightly freshness gate and opened #572.

## Current Situation

The monorepo carries Cisco-specific installation context, bounded subprocess output, the Cisco bundle identifier/version, and ad-hoc-by-default signing behavior on top of the standalone app. Importing v1.1.7 therefore required a three-way merge instead of copying upstream files wholesale.

Draft PR #511 addresses command cancellation separately, but it does not advance the immutable upstream pin or import v1.1.7's redacted audit-detail handling.

## Solution

Import the immutable v1.1.7 source commit through the repository updater and resolve the four integration conflicts by preserving both sides' safety properties:

- retain Cisco installation mutation controls, protected subprocess environment, bounded output capture/streaming, version, bundle identifier, and unsigned signing fallback;
- add bounded SIGINT → SIGTERM → SIGKILL cancellation, race-safe run reservation, and descendant-held pipe termination;
- hydrate bounded alert previews only when selected and parse redacted detail placeholders without treating placeholder metadata as event fields;
- retain and execute the new upstream cancellation and structured-detail regression harnesses on future imports.

Closes #572.

## Architecture Diagram

N/A.

## What Changed (Where & How)

| File | Summary |
| --- | --- |
| `macos/DefenseClawMac/upstream.lock.toml`, `UPSTREAM.md` | Pin stable v1.1.7 to immutable commit `f8a79fe0ce86facd81205b7d6ff8a480c8e5a3b4`. |
| `macos/DefenseClawMac/DefenseClawMac/**` | Import v1.1.7 cancellation and redacted audit-detail behavior while preserving Cisco runtime and output-safety integration. |
| `macos/DefenseClawMac/Tests/**`, `script/test_*.sh` | Add deterministic cancellation and structured-detail regression coverage. |
| `scripts/update-macos-app.sh`, `Makefile` | Preserve the new harnesses in later imports and run them in `macos-app-test`. |

## Breaking Changes

None.

## Open Issues / Follow-ups

None.

## Test Plan

- `DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer make macos-app-test` → all 14 macOS script suites passed; unsigned Release build completed with `CODE_SIGNING_ALLOWED=NO` and `** BUILD SUCCEEDED **`. The host emitted its known non-blocking out-of-date CoreSimulator warning.
- `python3 scripts/check-macos-upstream.py --json` → `status: current`, pinned/latest `v1.1.7@f8a79fe0ce86`.
- `make check-version-sync` → `version sync OK: 0.8.6 (source epoch 2, runtime 8)`.
- `python3 scripts/macos_license_headers.py` → `macOS app license headers OK`.
- `uv run --frozen pytest cli/tests/test_ci_workflow_efficiency.py -q` → `4 passed in 0.49s`.
- `bash -n scripts/update-macos-app.sh` → passed.
- `scripts/update-macos-app.sh v1.1.7` → `macOS app is already pinned to v1.1.7@f8a79fe0ce86`.
- `git diff --check` → passed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -791,6 +791,8 @@ macos-app-test:
 	macos/DefenseClawMac/script/test_first_run_connector_selection.sh
 	macos/DefenseClawMac/script/test_numeric_safety.sh
 	macos/DefenseClawMac/script/test_output_safety.sh
+	macos/DefenseClawMac/script/test_cli_cancellation.sh
+	macos/DefenseClawMac/script/test_structured_detail_parser.sh
 	macos/DefenseClawMac/script/test_secret_file_safety.sh
 	macos/DefenseClawMac/script/test_runtime_install_filesystem.sh
 	macos/DefenseClawMac/script/test_app_state_signal_safety.sh
--- a/macos/DefenseClawMac/DefenseClawMac/DataLayer/AuditStore.swift
+++ b/macos/DefenseClawMac/DefenseClawMac/DataLayer/AuditStore.swift
@@ -189,6 +189,16 @@ actor AuditStore {
         ).map(decodeAuditEvent)
     }
 
+    /// Loads the complete record for an inspector selection. Alert list rows
+    /// intentionally carry only a bounded details preview.
+    func event(id: String) -> AuditEvent? {
+        guard tableExists("audit_events"), !id.isEmpty else { return nil }
+        return query(
+            "SELECT * FROM audit_events WHERE id = ? LIMIT 1",
+            binds: [id]
+        ).first.map(decodeAuditEvent)
+    }
+
     func scanFindings(runID: String? = nil, target: String? = nil, limit: Int = 20) -> [ScanFindingEvent] {
         guard tableExists("scan_results") else { return [] }
         var conditions = ["raw_json IS NOT NULL", "raw_json != ''"]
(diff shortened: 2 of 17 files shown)


vineethsai7:
Security review finding: this branch inherits the command-cancellation descendant process leak that was already raised on #511.

Cancellation sends SIGINT/SIGTERM/SIGKILL only to `active.process.processIdentifier`. A command can spawn a child or background process that survives after the direct process exits or is killed, leaving privileged or long-running CLI work alive while the UI reports cancellation.

Please launch each command in its own process group/session and signal the group, with regression coverage for a long-lived descendant.

vineethsai7:
Security review finding: a late Cancel can replace the real result with a removable cancelled row.

If the process finishes and clears `runStates` before the store continuation updates the row, Cancel returns `.notFound`; the UI marks the row `.cancelled`, and `clearCompleted()` can remove it before the real exit/output lands. That loses the authoritative command result.

Please treat `.notFound` during this race as `.finishing`, or otherwise keep the row active until the `run` continuation records the true result.

vineethsai7:
Security review finding: active cancelling/finishing rows can still be pruned by the 500-entry cap.

The insertion pruning path removes entries with `status != .running`, so `.cancelling` and `.finishing` rows are treated as completed and can be dropped under load before their final output/status lands. `clearCompleted()` already uses `!status.isActive`; this insertion path should use the same predicate.

vineethsai7:
Security review finding: alert-list details truncation relies on duplicate SQLite column overwrite behavior.

`SELECT *, substr(COALESCE(details, ''), 1, 4096) AS details` depends on the row decoder overwriting the original `details` column with the later alias. If column order or the row dictionary behavior changes, list rows can silently return unbounded `details`; the truncated value also has no marker showing it is incomplete.

Please select an explicit column list and return a marked truncated `details` value for list rows.

PR #579, opened 2026-07-21
fix(managed_enterprise): IPC stats accuracy + idempotent installer + default action mode

cjsanjay:
## Summary

Managed-enterprise reliability sweep on release-defenseclaw-enterprise-26.7.3, five logically separate changes bundled because they share the same "managed rollouts should Just Work on the second install" theme:

### 1. IPC stats API — ActiveAlerts + TotalScans now reflect reality

`GetStatsSnapshot`'s `ActiveAlerts` and `TotalScans` stayed pinned at 0 in managed_enterprise deployments even while connector hooks (Claude Code / Codex / agent hook) actively inspected and blocked traffic.

- **ActiveAlerts:** `gateway.logConnectorHookAuditEnvelope` hardcoded `audit_events.severity=INFO` (a deliberate noise-reduction choice, since every tool call produces a row). The real verdict severity was living on `structured_json`. Teach `Store.GetCounts()` to OR in a JSON-extract branch scoped to `action='connector-hook'`. Column and JSON branches are disjoint by column value; no double-count.
- **TotalScans:** in managed_enterprise the hook lane runs through `inspectManagedAIDOnly` → `mergeWithAIDVerdict(nil, aid)`, which populated only `verdict.Findings` (stringy) and never `DetailedFindings`. `emitInspectVerdictFindings` early-returned on empty `DetailedFindings`, so `scan_results` never grew. Synthesize one `RuleFinding` per lane finding in `mergeWithLaneVerdict` (RuleID = `<tag>:<name>`, Severity from the lane verdict). Applies symmetrically to the LLM-judge lane.

### 2. Idempotent installer

The managed installers (`packaging/macos/install.sh`, `packaging/launchd/install-enterprise.sh`) previously `die`-d on any existing marker — including a stray `~/.defenseclaw` in any local account — with "existing DefenseClaw installation detected; no changes were made. This installer is fresh-install-only." A second run made no changes, drifted hosts stayed drifted.

The refuse-on-existing loops are replaced with a classify-then-reconcile flow:
- **Current-generation markers** → in-place reinstall. Binary + config.yaml + guardian manifest + both plists atomically overwritten via `rename(2)`. Current-gen launchd jobs booted out before rebootstrap so the atomic replace hits a quiescent target.
- **Legacy paths** (pre-Cisco layout: `/Library/DefenseClaw`, `com.defenseclaw.gateway.plist`, etc.) → relocated under `LOG_DIR` with a `.pre-<version>-<timestamp>` suffix. Best-effort, never deleted — operators keep forensic access. Legacy launchd labels are unloaded.
- **Per-user `~/.defenseclaw`** → left untouched entirely. The hook-guardian daemon reconciles that layer on its own 60 s tick, and wiping it would drop audit history.
- **Runtime state** (`audit.db`, `judge_bodies.db`, `device.key`, `hook-guardian-state/`) → preserved across reinstalls.
- **Symlinks** under a DefenseClaw-owned install destination are still refused (an atomic rename over a symlink would corrupt the pointee; a symlink there can only be placed via privileged tampering).

The pure-function helpers `install_file_no_replace` / `create_install_directory_no_replace` are renamed to `install_file_reconcile` / `create_install_directory_reconcile` in the bundle installer; `install-enterprise.sh` keeps the historical function names but their bodies reconcile in place. A new `move_legacy_aside()` helper in `installer_lib.sh` (with `--dry-run` mode) is shared by both.

### 3. Default mode = `action` in managed_enterprise installer

`packaging/macos/install.sh`'s `DEFAULT_MODE` flips from `observe` to `action`. That installer only ever writes `deployment_mode: managed_enterprise`, so the flip is managed-scoped. Operators who want an observe pilot still pass `--mode observe`.

Unmanaged installers (`scripts/install.sh`, `scripts/install.ps1`, `defenseclaw quickstart`, `defenseclaw init`) keep `observe`. Go `DefaultConfig()` and Python `GuardrailConfig.mode` dataclass defaults are untouched — zero test churn on the large Python/TUI/Go config test suites.

### 4. ai_discovery walks every eligible user home (managed_enterprise)

Root-cause found while capturing discovery events for the AID dashboard schema: under managed_enterprise the daemon runs as root, so `os.UserHomeDir()` resolves to `/var/root` — and every per-user detector (editor_extension, mcp_server, config paths, shell history, applications) walks a directory that contains no user data. On real endpoints the discovery service silently missed every human user's dotfiles. Post-fix the same PoC box went from 4 category types (process/desktop-app only) to 11 (adds editor_extension, mcp_server, supported_connector, package_dependency, shell_history_match, provider_domain, workspace_artifact).

Fix:
- New `AIDiscoveryOptions.HomeDirs` (config: `ai_discovery.home_dirs`), seeded with `HomeDir` so single-user/dev installs are unchanged.
- `homesToScan()` helper; `expandCandidatePath`, `scanRoots`, `detectEditorExtensions`, `detectApplications`, `detectShellHistory` iterate every home.
- Installer + hook-enumerator populate `home_dirs` from the same `enumerate_local_users` pass that renders `hook-guardian/targets.yaml`, so per-user detectors and per-user hook wiring stay in lockstep as users are added/removed. New `apply_ai_discovery_home_dirs` helper in `installer_lib.sh` is idempotent + atomic (no-op when the resolved list is unchanged, so enumerator ticks don't churn config mtime).

### 5. Skill / rule / plugin discovery signals

`SignalSkill`, `SignalRule`, and `SignalPlugin` were declared and allow-listed but no detector emitted them — the categories showed up as empty slots in the discovery dashboard even on endpoints with skills/rules/plugins configured for one of the supported agents.

- Three new detectors (`detectSkills` / `detectRules` / `detectPlugins`) mirror `detectMCPPaths` but require the target path to exist AND, if a directory, contain at least one entry. Empty scaffold directories that agents create pre-emptively don't fire — a signal means "this surface is configured", not "the scaffold exists". Non-directory targets (a scalar like `~/.claude/CLAUDE.md`) still count as populated.
- Three new `AISignature` fields (`skill_paths`, `rule_paths`, `plugin_paths`) added to `codex`, `claudecode`, `cursor`, `hermes`, `openclaw`, `windsurf` — the six connectors that actually expose those surfaces.

## Also in this PR (unblocking CI on the release branch)

- Roll back `0.8.4.1` → `0.8.4` on the release branch: the `0.8.4.1` version broke every canonical-X.Y.Z semver validator (`source_release_identity.py`, `check-upgrade-manifest`, `check-version-sync`, `test-upgrade-release.sh`). The "Drop 2" distinction is carried in git history rather than the version string.
- Bump `extensions/defenseclaw/package-lock.json` `js-yaml` `4.2.0 → 4.3.0` to clear `GHSA-52cp-r559-cp3m` (YAML merge-key quadratic-CPU DoS). Prod audit clean; dev-only `esbuild` GHSA is ignored by `npm audit --omit=dev`.

## Test plan

- [x] `go build ./...`
- [x] `go test ./internal/audit/... ./internal/ipc/... ./internal/gateway/... ./internal/config/... ./internal/inventory/... ./internal/scanner/...`
- [x] `bash packaging/macos/tests/run_tests.sh` — 135 passed, 0 failed (includes the 7 `test_move_legacy_aside.sh` cases, `DEFAULT_MODE=action` pin, and unchanged `install.sh` + `installer_lib.sh` render tests)
- [x] `pytest cli/tests/test_enterprise_packaging.py cli/tests/test_install_docs_static.py` — 52 passed
- [x] End-to-end multi-user ai_discovery capture on a live managed_enterprise install: 15 of 15 declared signal categories emit; the previously-blind editor_extension / mcp_server / package_dependency / shell_history_match / provider_domain / workspace_artifact / skill / rule / plugin all now fire from the real user's home
- [ ] End-to-end managed_enterprise install verification on a disposable macOS host: fresh install → reinstall → drifted-config reinstall → legacy-path reinstall (script at `scripts/test-macos-enterprise-packaging.sh`, gated on `MACOS_ENTERPRISE_PACKAGING_SMOKE=1`)
- [ ] Live sanity: after reinstall, submit a known-blocking tool call through Claude Code; assert the block is enforced (not would-block) and that both `ActiveAlerts` and `TotalScans` on `GetStatsSnapshot` advance


vineethsai7:
Security review finding: `internal/ipc/notifier_broadcast.go` can still drop a retained notification for a late subscriber.

In `subscribe`, a late subscriber receives every retained record on replay, but it is only added to `pending` when `ackedByAny == false`. That means this sequence can lose a record: subscriber A gets retained record R; subscriber B subscribes before A acks and receives R in its replay buffer, but B is not added to `pending` because `ackedByAny` may already be true; A then acks and `ackSubscriber` evicts R. If B disconnects before `stream.Send(R)` succeeds, `cancel()` has no pending marker to release and R is gone, so the next subscriber never sees it. This violates the comment/contract that records remain retained until every currently-live subscriber has successfully sent them.

Please either add replayed late subscribers to `pending` for any retained record they are handed, or avoid replaying already-delivered records to late subscribers after they are eligible for eviction. A focused regression test would be: publish R, have A receive/ack enough to set `ackedByAny`, have B subscribe and receive R without acking, then let A/remaining pending eviction run and cancel B; the next subscriber should still replay R if B never acked it.

vineethsai7:
Security review finding: `--override-endpoint` accepts plaintext or credential-bearing AID endpoints that can leak enterprise auth material.

`resolve_aid_endpoint` accepts values like `http://host`, and the installer only warns before writing the endpoint into managed config. The rendered config comments indicate CMID bearer tokens are sent to `${aid_endpoint}/api/v1/...`, so a plaintext or userinfo-bearing override can expose enterprise credentials.

Please port the stricter validation from the mainline packaging path: require an HTTPS bare origin only, with no credentials, path, query, fragment, whitespace, quotes, or backslash. Add installer tests rejecting `http://`, `https://user@host`, paths, query strings, and fragments.

vineethsai7:
Security review finding: the idempotent reinstall path can relocate legacy material through an untrusted logs backup root.

The reinstall flow creates and uses `LOGS_DIR` as the legacy backup root before the later trusted-directory checks. If `/Library/Logs/Cisco` or `/Library/Logs/Cisco/SecureClient` is a symlink or otherwise untrusted, the root installer follows it during `mkdir -p` / `mv` and can move legacy config, logs, or audit material into an attacker-controlled location before validation fails later.

Please validate `/Library/Logs`, `/Library/Logs/Cisco`, `/Library/Logs/Cisco/SecureClient`, and `LOGS_DIR` with the same no-symlink, root-owned, no-write-ACL policy before any `mkdir -p` or `move_legacy_aside`. `move_legacy_aside` should also reject an untrusted backup root, with tests for symlinked backup roots.

vineethsai7:
Security review finding: managed-enterprise hook inspections that allow with zero findings still undercount `TotalScans`.

The patch synthesizes `DetailedFindings` only when `aid.Findings` is non-empty, but `emitInspectVerdictFindings` still returns before calling `scanner.EmitInspectFindings` when there are zero detailed findings. An AID allow/no-findings hook inspection did run a scan, but no `scan_results` row is written, so IPC `TotalScans` remains undercounted.

Please let verdict-bearing hook inspections call `scanner.EmitInspectFindings` even with an empty findings slice, since that helper already supports zero-finding scan rollups. Add a managed AID allow/no-rules test asserting `GetCounts().TotalScans` increments.

cjsanjay:
@vineethsai7 addressed in a8202521.

The `if !ackedByAny` fast path in `subscribe()` is gone — every late subscriber now unconditionally joins each retained record's `pending` set on replay. Combined with the existing `releasePendingLocked` invariant (`cancel` releases the marker but does NOT flip `ackedByAny`), the ring evicts only when every live subscriber has actually acked, so a cancel-without-ack from a late subscriber keeps the record retained for the next arrival.

Two new regression tests exercise the exact scenario from the finding:

- `TestBroadcastLateSubscriberCancelWithoutAckDoesNotDropAckedRecord` — sub A publishes/acks R with sub C also live, sub B late-subscribes and picks up R on replay, sub C acks, sub B cancels without acking. The record is fully delivered on the wire to A and C, so eviction here is correct — but the pending-tracking machinery covers B honestly (B's cancel releases its pending marker exactly like A's/C's acks would have).
- `TestBroadcastLateSubscriberCancelWithoutAnyAckKeepsRetained` — the "no ack ever landed" companion where no subscriber acked R. Two churning subscribers each subscribe→cancel without acking; a third subscriber must still see R on replay. This is the invariant the finding's proposed regression asked for.

Also updated the pre-existing `TestBroadcastRetainsForOtherSubscribers` to reflect the corrected contract (record now stays retained until every live subscriber — original + late — has acked).

cjsanjay:
@vineethsai7 addressed in a8202521.

`_valid_aid_endpoint_url` in `packaging/macos/lib/installer_lib.sh` is rewritten as a strict HTTPS bare-origin check:

- HTTPS only (no `http://` — plaintext would leak the CMID bearer on the wire).
- No userinfo (`user@` or `user:pass@`).
- No path, query, or fragment. The daemon appends `/api/v1/inspect/defense_claw` itself; an operator-supplied path would double-append or silently override the routing.
- Whitespace, double quotes, and backslashes rejected via a pre-filter case statement (belt-and-braces for the YAML-scalar landing zone).
- Optional `:port` on the host.

Both sources go through the same validator — `--override-endpoint` and the JSON `cisco_ai_defense_endpoint` field in AVC's `env_config.json`. The dead plaintext-http warning in `install.sh` is deleted since it's now unreachable behind the tighter regex. The `--help` and error strings both name the "HTTPS bare origin" contract explicitly.

Five new installer tests cover the exact rejection cases from the finding:

- `t_install_override_endpoint_rejects_plaintext_http` — `http://example.com` → exit 1 with "HTTPS bare origin" message.
- `t_install_override_endpoint_rejects_userinfo` — `https://user@` and `https://user:pass@` both rejected.
- `t_install_override_endpoint_rejects_path_query_fragment` — `/api`, `/api/v1/inspect`, `?tenant=x`, `#frag` all rejected.

Plus symmetric config-file-side coverage inside `t_resolve_aid_endpoint_precedence`: a hostile `env_config.json` with `"cisco_ai_defense_endpoint": "http://example.com"` (or userinfo, or a path) now produces rc 2 rather than being silently accepted.

cjsanjay:
@vineethsai7 addressed in a8202521.

Two layers of defense added:

**Outer (install.sh):** `_assert_trusted_logs_chain_or_die` walks `/`, `/Library`, `/Library/Logs`, `/Library/Logs/Cisco`, `/Library/Logs/Cisco/SecureClient` — and then `LOGS_DIR` itself once it exists — applying the same no-symlink / root-owned / no-group-other-write / no-write-capable-ACL policy that the env_config.json trust check uses. This runs BEFORE any `mkdir -p` or `move_legacy_aside`. An ancestor that doesn't exist yet is fine (we create it below with the right perms); an ancestor that DOES exist must be trusted. After the `mkdir -p` sequence we re-verify LOGS_DIR to catch a racing symlink between creation and use.

**Inner (installer_lib.sh):** `move_legacy_aside` itself now rejects a symlinked `backup_root` with a new rc 4, as second-line-of-defense against future callers that skip the outer trust check. A new `t_symlinked_backup_root_rejected` test stages a symlinked backup root pointing at attacker-controlled space and asserts:

- rc 4 returned.
- Legacy source is untouched (no relocation happened).
- The symlink target is empty (no data flowed through the link).

The trust helpers (`_refuse_symlink_or_die`, `_assert_no_write_acl_or_die`, `_assert_trusted_dir_or_die`) live inline in `install.sh` and are gated on `DC_INSTALLER_SKIP_ROOT_CHECK=1` — the same seam the euid check already uses — so tmpdir-based tests can still exercise the arg-parse / render_config surface without needing a root-owned fixture. The real-install path always runs the full trust walk.

cjsanjay:
@vineethsai7 addressed in a8202521.

`emitInspectVerdictFindings` in `internal/gateway/hook_findings_emit.go` had `if verdict == nil || len(verdict.DetailedFindings) == 0 { return }`. Relaxed to `if verdict == nil { return }` so a non-nil verdict with zero findings flows through to `scanner.EmitInspectFindings`, which handles zero-finding rollups exactly as you noted — emits an `EventScan` summary + a `scan_results` row (finding_count=0) without any `scan_findings` rows. The IPC `TotalScans` counter now reflects every managed AID hook inspection regardless of whether AID returned a hit.

Two new tests pin the invariant:

- `TestEmitInspectVerdictFindings_ZeroFindingsStillCountsAsScan` — non-nil `ToolInspectVerdict{Action:"allow", Severity:"NONE", DetailedFindings:nil}` through `emitInspectVerdictFindings` → assert `store.GetCounts().TotalScans == 1` (was 0 before the fix). Also asserts `Alerts == 0` so an allow verdict doesn't inadvertently raise an alert row.
- `TestEmitInspectVerdictFindings_NilVerdictSkipped` — nil verdict = no inspection = counter stays at 0. This is the guard that stays.

Note: I intentionally did NOT touch `emitGuardrailScanVerdictFindings` (the proxy-side companion) — its "no scan_results for empty findings" guard is a deliberate quality bar for LLM proxy traffic, doesn't share the managed_enterprise counter concern, and would be a larger behavioral change to touch. The hook path is where the reported managed_enterprise TotalScans undercounting actually manifests.

packaging/launchd/install-enterprise.sh
```diff
@@ -390,59 +415,144 @@ done
...
+        /bin/launchctl bootout "system/${_legacy_label}" >/dev/null 2>&1 \
+            || warn "launchctl bootout system/${_legacy_label} failed; the legacy plist will still be relocated below"
+    fi
+done
+unset _legacy_label
+
+# Ensure LOG_DIR exists early so the legacy relocation below has a
+# landing zone. Recreated with the right ownership later during the
+# mutation phase; a bare directory here is enough.
+for _log_parent in /Library/Logs "$LOG_VENDOR_DIR" "$LOG_PRODUCT_DIR"; do
```
vineethsai7, line 527: This legacy/documented installer path still relocates legacy state before proving the logs backup root is trusted. `packaging/macos/install.sh` now walks `/Library/Logs/Cisco/SecureClient/...` before any mkdir/mv, but this script creates `LOG_DIR` and moves `/Library/DefenseClaw`, `/Library/Application Support/DefenseClaw`, logs, and legacy plists into it at lines 507-548, then only validates the logs chain later at lines 568-582. Because this script is still documented and covered by `scripts/test-macos-enterprise-packaging.sh`, a symlinked or ACL-writable `/Library/Logs/Cisco` can still cause the root installer to move legacy config/audit material into an attacker-controlled target before validation fails. Please port the same pre-mutation logs-chain trust check here, and make the smoke/static tests assert it runs before the early mkdir/mv relocation block.

vineethsai7 (commented):
Daily security review finding.

cjsanjay:
Pushed two follow-up commits addressing an ai_discovery blindness bug found while capturing discovery events for the AID dashboard schema PoC — plus filling in three previously-empty signal categories.

**`8ceae65f` — fix(ai_discovery): walk every eligible user home in managed_enterprise**

Under managed_enterprise the daemon runs as root, so `os.UserHomeDir()` resolves to `/var/root`. Every per-user detector (editor_extension, mcp_server, config paths, shell history, applications) was walking a directory that contains no user data — the discovery service silently missed every human user's dotfiles on real endpoints.

- New `AIDiscoveryOptions.HomeDirs` (config: `ai_discovery.home_dirs`), seeded with `HomeDir` so single-user installs are unaffected.
- `homesToScan()` helper; every detector that reads `HomeDir` now iterates via it (also updates `expandCandidatePath`, `scanRoots`, `scanRootsForRelative` so `~/…` catalog patterns expand across all homes).
- New `apply_ai_discovery_home_dirs` helper in `installer_lib.sh`. Installer + hook-enumerator populate `home_dirs` from the same `enumerate_local_users` pass that renders `hook-guardian/targets.yaml`, so per-user detectors and per-user hook wiring stay in lockstep as users are added/removed. Atomic + idempotent (no-op when list unchanged, so enumerator ticks don't churn config mtime).

Verified on my host: previously-blind categories now fire — editor_extension, mcp_server, supported_connector, package_dependency, shell_history_match, provider_domain, workspace_artifact. Went from 4 emitted categories to 11.

**`5d2b3b75` — feat(ai_discovery): emit skill / rule / plugin inventory signals**

`SignalSkill` / `SignalRule` / `SignalPlugin` existed as allow-listed category constants but nothing emitted them. Three new detectors (`detectSkills` / `detectRules` / `detectPlugins`) mirror `detectMCPPaths` with one extra check: the target path must exist AND, if a directory, contain at least one entry. Empty scaffold directories agents create pre-emptively don't fire — signal means "this surface is configured", not "the scaffold exists". A scalar file (e.g. `~/.claude/CLAUDE.md`) still counts as populated.

New `AISignature` fields `skill_paths` / `rule_paths` / `plugin_paths` seeded on the six connectors that actually expose those surfaces: codex (skills / rules / plugins), claudecode (skills / rules), cursor (rules), hermes (skills / plugins), openclaw (plugins), windsurf (rules).

Post-fix the AID discovery-events capture covers 15 of 15 declared signal categories. Full breakdown from the reference capture (`/tmp/dc-discovery-logs.ndjson`):

```
active_process        (existing)
desktop_app           (existing)
local_ai_endpoint     (existing)
local_model           (existing)
editor_extension      (fixed by #4)
mcp_server            (fixed by #4)
supported_connector   (fixed by #4)
package_dependency    (fixed by #4)
shell_history_match   (fixed by #4)
provider_domain       (fixed by #4)
workspace_artifact    (fixed by #4)
skill                 (new via #5)
rule                  (new via #5)
plugin                (new via #5)
env_var_name          (still needs `launchctl setenv` in plist — separate follow-up)
```

**Tests**
- `go build ./...` clean
- `go test ./internal/inventory/... ./internal/config/... ./internal/ipc/...` pass
- `bash packaging/macos/tests/run_tests.sh` — 135 pass (unchanged; `install.sh` and `installer_lib.sh` diffs cleared existing coverage)
- End-to-end packaging smoke: shell-side `apply_ai_discovery_home_dirs` round-trips through PyYAML clean across initial-render → 2-user → idempotent re-apply → user-removed → empty

**One follow-up worth flagging**
`env_var_name` still comes up empty in the capture because launchd-launched root daemons don't inherit user shell env. `sudo launchctl setenv OPENAI_API_KEY …` reaches the daemon on the current login but doesn't persist across reboots. If we want that category populated, plist should declare the AI-provider env names it should observe (values still redacted). Splitting to a separate ticket since it's orthogonal to the home-dirs fix.

packaging/launchd/install-enterprise.sh
cjsanjay: Addressed in `7caa8aed`.

Ported the pre-mutation logs-chain trust check from `packaging/macos/install.sh` (where it landed in `a8202521`) to `packaging/launchd/install-enterprise.sh`. Every `/Library/Logs/…` ancestor is now proven trusted via the existing `assert_trusted_system_dir` / `assert_existing_secure_dir_or_absent` primitives **before** the early `mkdir -p` and legacy-relocation `mv` block. LOG_DIR is re-verified post-mkdir so a racer that swaps in a symlink between the initial check and the relocation still can't slip through.

Also added ordering assertions in `test_launchd_enterprise_installer_enforces_managed_config_trust_boundary` (`cli/tests/test_enterprise_packaging.py`): the gate offset must precede both the mkdir-landing block and the `moved legacy path aside` log line, and the gate block itself must reference every ancestor (`/Library`, `/Library/Logs`, `$LOG_VENDOR_DIR`, `$LOG_PRODUCT_DIR`, `$LOG_DIR`) by name.

Tests: `pytest cli/tests/test_enterprise_packaging.py` — 8 passed; `bash packaging/macos/tests/run_tests.sh` — 135 passed. `bash -n` on the enterprise installer passes.

PR #586, opened 2026-07-22
fix(gateway): preserve extra_content on streaming tool calls (Gemini …

nghodkicisco:
…thought_signature)

Re-applies the changes from PR #366 (reverted in #580).

- Bump bifrost/core to v1.7.3 which includes ExtraContent on ChatAssistantMessageToolCall and SecretVar type for credentials
- Add ExtraContent field to ChatMessage so thought_signature passes through the guardrail proxy on streaming deltas
- Add extractExtraParams to forward extra_body (google.thinking_config) to upstream providers via Bifrost
- Update CI audit ignore list for transitive Python/npm CVEs

This ensures Gemini multi-turn tool calls with extended thinking do not fail with "missing thought_signature" when traffic flows through the DefenseClaw guardrail proxy.

## What Changed (Where & How)

| File | Summary |
| --- | --- |
|  |  |


internal/gateway/provider_bifrost.go
```diff
@@ -844,6 +841,9 @@ func fromBifrostStreamChunk(resp *schemas.BifrostChatResponse) StreamChunk {
 					msg.ToolCalls = raw
 				}
 			}
+			if len(d.ExtraContent) > 0 {
+				msg.ExtraContent = d.ExtraContent
```
vineethsai7, line 845: Security/behavior regression: streaming responses now preserve `d.ExtraContent`, but request history conversion still drops `ChatMessage.ExtraContent`. The next direct-provider request goes through `toBifrostMessages` and only copies role/name/content/tool-call fields, so Gemini `thought_signature` can still be lost before Bifrost sends the history upstream. Please map `m.ExtraContent` into the Bifrost message/assistant message field supported by `schemas.ChatMessage`, and add a round-trip test that an assistant message with `extra_content` survives `ChatMessage` JSON unmarshal -> `toBifrostMessages`.

vineethsai7 (commented):
Daily security review finding.

PR #633, opened 2026-07-29
fix(managed_enterprise): dynamic env_config reload, Claude Code detector, config-reload restart classification

cjsanjay:
## Summary

Bundles several managed_enterprise fixes surfaced by QA on macOS 26 / Mac 26.5 hosts.

### 1) env_config dynamic reload
`internal/config/env_config.go` (new), `internal/gateway/config_manager.go`

The AVC-authored `/opt/cisco/secureclient/defenseclaw/env_config.json` may arrive AFTER the installer completes, but the gateway used to bake the endpoint value in at boot and hold it for the process lifetime. ConfigManager now watches `env_config.json` alongside `config.yaml`, overlays `cisco_ai_defense_endpoint` on top of the yaml value, and reloads on any fsnotify event.

Malformed env_configs are **logged + retained** (health check surfaces the reason) rather than silently overwriting a working endpoint — bad-endpoint exfiltration is exactly the vector the validation step defends against.

### 2) Config-reload restart classification
`internal/gateway/config_manager.go`

Fixes three latent bugs that surfaced the moment env_config triggered a real reload. All three shared the same failure mode: reload rejected with `config reload requires gateway restart for: <field>`.

- **cisco_ai_defense wasn't hot-reloadable.** Any endpoint change tripped the restart guard even though `inspectorNeedsRebuild` + `otelNeedsReload` already handle the change. Added under managed_enterprise gate; OSS path keeps the pre-existing behavior.
- **Gateway.Token synthesised at boot, not persisted.** `ensureGatewayTokenSynthesis` generates a token into the runtime snapshot but never writes it into `config.yaml`, so every reload compared runtime-token vs on-disk-empty and forced a gateway restart. Existing preservation at `applyConfigReload:1165` ran AFTER `diffConfigs`, too late. Moved preservation into the pre-diff normalization step.
- **Gateway.NoTLS / SandboxHome / ClawHome same shape.** All three are `mapstructure:"-"` (never in yaml) and set post-Load, tripping the same false-restart signal on standalone-mode hosts. Extended the preservation set to cover every runtime-only Gateway field.

### 3) Claude Code detector — newer install layouts
`packaging/macos/lib/installer_lib.sh`, `packaging/macos/tests/test_discover_agent_version.sh`

Newer Claude Code native installers publish `versions/<X.Y.Z>` as executable **files** (not directories) and skip the `current` symlink, using `~/.local/bin/claude -> ../share/claude/versions/<X.Y.Z>` as the PATH shim. The dir-only glob was returning empty version, producing:

```
connector claudecode agent version "" is not verified against a known hook contract;
using connector default hook contract
```

Added file-shape support to `_native_claudecode_version_from_dir` and a new `_native_claudecode_version_from_bin` shim resolver covering `~/.local/bin`, `/usr/local/bin`, `/opt/homebrew/bin`, `/opt/claude/bin`.

### 4) Inventory / AI discovery updates
`cli/defenseclaw/inventory/*`, `internal/inventory/*`

Multi-install agent-version discovery on the Python side, seen-vs-new state tracking for AI discovery events, and misc detector coverage.

## Test plan

- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `go test ./internal/config/... ./internal/gateway/...` — all pass (~42s)
- [x] `packaging/macos/tests/run_tests.sh` — **151/151 pass** including new agent-version detector + multi-install cases
- [x] Verified end-to-end on a managed_enterprise host: env_config.json arrives post-install, ConfigManager reloads without a restart, inspect client rebuilds against the fresh endpoint (previously blocked with `config reload requires gateway restart for: cisco_ai_defense, gateway`)

## Backwards compatibility

- OSS reload path is untouched — every new preservation branch is gated on `managed.IsManagedEnterprise(next.DeploymentMode)`.
- `Provider` / `Config` interfaces unchanged; existing callers need no modifications.
- Old Claude Code install layouts (directory-shape versions/, `current` symlink) continue to resolve exactly as before — the file-variant support is additive.

### DIFF
--- a/cli/defenseclaw/config.py
+++ b/cli/defenseclaw/config.py
@@ -2147,6 +2147,7 @@ class AIDiscoveryConfig:
     confidence_policy_path: str = ""
     require_trusted_binary_paths: bool = False
     trusted_binary_prefixes: list[str] = field(default_factory=list)
+    home_dirs: list[str] = field(default_factory=list)
 
 
 @dataclass
@@ -4865,6 +4866,7 @@ def _merge_ai_discovery(raw: dict[str, Any] | None) -> AIDiscoveryConfig:
         confidence_policy_path=str(raw.get("confidence_policy_path", "") or ""),
         require_trusted_binary_paths=bool(raw.get("require_trusted_binary_paths", False)),
         trusted_binary_prefixes=[str(v) for v in (raw.get("trusted_binary_prefixes", []) or [])],
+        home_dirs=[str(v) for v in (raw.get("home_dirs", []) or [])],
     )
 
 
--- a/cli/defenseclaw/connector_paths.py
+++ b/cli/defenseclaw/connector_paths.py
@@ -93,6 +93,36 @@ import time) means a typo in ``guardrail.connector`` surfaces in
 producing wrong paths.
 """
 
+KNOWN_AGENT_KINDS: tuple[str, ...] = KNOWN_CONNECTORS + (
+    # Discovery-only agents. These have AI-signature entries in
+    # ai_signatures.json (mirrored between Go and Python copies) but no
+    # DefenseClaw enforcement path, so they do NOT appear in
+    # KNOWN_CONNECTORS (which is the enforcement-connector allow-list).
+    # They ARE first-class agents on the discovery/inventory side:
+    # ``defenseclaw agent discover`` reports them, and the wire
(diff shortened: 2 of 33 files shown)


cjsanjay:
Superseded by a new PR against release-defenseclaw-enterprise-26.7.3 with the same commit plus follow-up polish.

PR #634, opened 2026-07-29
fix(managed_enterprise): dynamic env_config reload, Claude Code detector, config-reload restart classification

cjsanjay:
Supersedes #633 (closed) — same core commit, plus a follow-up polish commit and retargeted to the release branch.

## Summary

Bundles several managed_enterprise fixes surfaced by QA on macOS 26 / Mac 26.5 hosts.

### 1) env_config dynamic reload
`internal/config/env_config.go` (new), `internal/gateway/config_manager.go`

The AVC-authored `/opt/cisco/secureclient/defenseclaw/env_config.json` may arrive AFTER the installer completes, but the gateway used to bake the endpoint value in at boot and hold it for the process lifetime. ConfigManager now watches `env_config.json` alongside `config.yaml`, overlays `cisco_ai_defense_endpoint` on top of the yaml value, and reloads on any fsnotify event. Filename and location are fixed — the watcher only picks up the canonical path.

Malformed env_configs are **logged + retained** (health check surfaces the reason) rather than silently overwriting a working endpoint — bad-endpoint exfiltration is exactly the vector the validation step defends against.

### 2) Config-reload restart classification
`internal/gateway/config_manager.go`

Fixes three latent bugs that surfaced the moment env_config triggered a real reload. All three shared the same failure mode: reload rejected with `config reload requires gateway restart for: <field>`.

- **cisco_ai_defense wasn't hot-reloadable.** Any endpoint change tripped the restart guard even though `inspectorNeedsRebuild` + `otelNeedsReload` already handle the change. Added under managed_enterprise gate; OSS path keeps the pre-existing behavior.
- **Gateway.Token synthesised at boot, not persisted.** `ensureGatewayTokenSynthesis` generates a token into the runtime snapshot but never writes it into `config.yaml`, so every reload compared runtime-token vs on-disk-empty and forced a gateway restart. Existing preservation at `applyConfigReload:1165` ran AFTER `diffConfigs`, too late. Moved preservation into the pre-diff normalization step.
- **Gateway.NoTLS / SandboxHome / ClawHome same shape.** All three are `mapstructure:"-"` (never in yaml) and set post-Load, tripping the same false-restart signal on standalone-mode hosts. Extended the preservation set to cover every runtime-only Gateway field.

### 3) Claude Code detector — newer install layouts
`packaging/macos/lib/installer_lib.sh`, `packaging/macos/tests/test_discover_agent_version.sh`

Newer Claude Code native installers publish `versions/<X.Y.Z>` as executable **files** (not directories) and skip the `current` symlink, using `~/.local/bin/claude -> ../share/claude/versions/<X.Y.Z>` as the PATH shim. The dir-only glob was returning empty version, producing:

```
connector claudecode agent version "" is not verified against a known hook contract;
using connector default hook contract
```

Added file-shape support to `_native_claudecode_version_from_dir` and a new `_native_claudecode_version_from_bin` shim resolver covering `~/.local/bin`, `/usr/local/bin`, `/opt/homebrew/bin`, `/opt/claude/bin`.

### 4) Inventory / AI discovery + docs polish
`cli/defenseclaw/inventory/*`, `internal/inventory/*`, `docs-site/`, `packaging/macos/PACKAGING.md`

Multi-install agent-version discovery on the Python side, seen-vs-new state tracking for AI discovery events, and misc detector coverage. PACKAGING.md corrected to reflect the fixed env_config.json path contract (QA flagged the previous wording as misleading — the doc showed `env_config.preview.json` / `env_config.prod.json` variants and `--config-file` on every command, but the watcher only reads the canonical path).

## Test plan

- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `go test ./internal/config/... ./internal/gateway/... ./internal/inventory/... ./internal/ipc/...` — all pass
- [x] `packaging/macos/tests/run_tests.sh` — **151/151 pass** including new agent-version detector + multi-install cases
- [x] Verified end-to-end on a managed_enterprise host: env_config.json arrives post-install, ConfigManager reloads without a restart, inspect client rebuilds against the fresh endpoint (previously blocked with `config reload requires gateway restart for: cisco_ai_defense, gateway`)

## Backwards compatibility

- OSS reload path is untouched — every new preservation branch is gated on `managed.IsManagedEnterprise(next.DeploymentMode)`.
- `Provider` / `Config` interfaces unchanged; existing callers need no modifications.
- Old Claude Code install layouts (directory-shape versions/, `current` symlink) continue to resolve exactly as before — the file-variant support is additive.

### DIFF
--- a/cli/defenseclaw/config.py
+++ b/cli/defenseclaw/config.py
@@ -2147,6 +2147,7 @@ class AIDiscoveryConfig:
     confidence_policy_path: str = ""
     require_trusted_binary_paths: bool = False
     trusted_binary_prefixes: list[str] = field(default_factory=list)
+    home_dirs: list[str] = field(default_factory=list)
 
 
 @dataclass
@@ -4865,6 +4866,7 @@ def _merge_ai_discovery(raw: dict[str, Any] | None) -> AIDiscoveryConfig:
         confidence_policy_path=str(raw.get("confidence_policy_path", "") or ""),
         require_trusted_binary_paths=bool(raw.get("require_trusted_binary_paths", False)),
         trusted_binary_prefixes=[str(v) for v in (raw.get("trusted_binary_prefixes", []) or [])],
+        home_dirs=[str(v) for v in (raw.get("home_dirs", []) or [])],
     )
 
 
--- a/cli/defenseclaw/connector_paths.py
+++ b/cli/defenseclaw/connector_paths.py
@@ -93,6 +93,36 @@ import time) means a typo in ``guardrail.connector`` surfaces in
 producing wrong paths.
 """
 
+KNOWN_AGENT_KINDS: tuple[str, ...] = KNOWN_CONNECTORS + (
+    # Discovery-only agents. These have AI-signature entries in
+    # ai_signatures.json (mirrored between Go and Python copies) but no
+    # DefenseClaw enforcement path, so they do NOT appear in
+    # KNOWN_CONNECTORS (which is the enforcement-connector allow-list).
+    # They ARE first-class agents on the discovery/inventory side:
+    # ``defenseclaw agent discover`` reports them, and the wire
+    # payload's ``agent_kind`` gets populated for their signals via the
+    # promotion table in internal/inventory/ai_catalog.go.
+    #
+    # Keep this list in sync with ``promotedAgentKinds`` in the Go
+    # catalog — the connector slug values here (aider, continue, cline,
+    # claudedesktop) must match the values in that map for dashboards
+    # to join cleanly across the two sides.
+    "aider",
+    "continue",
+    "cline",
+    "claudedesktop",
+)
+"""All agent slugs the CLI's discovery / inventory paths know about.
+
+Superset of :data:`KNOWN_CONNECTORS`. Callers that need to iterate over
+"every agent DefenseClaw can *see*" (agent-discovery cache, inventory
+report renderers, TUI agent list) should use this constant. Callers
+that need "every connector DefenseClaw can *enforce through*"
+(manifest validation, MCP-writer dispatch, hook installers) should
+continue to use :data:`KNOWN_CONNECTORS`.
+"""
+
+
 HOOK_ONLY_CONNECTORS: frozenset[str] = frozenset(
     {
         "hermes",
--- a/cli/defenseclaw/inventory/_semver.py
+++ b/cli/defenseclaw/inventory/_semver.py
@@ -37,9 +37,17 @@ from functools import cmp_to_key
 
 # Matches the first "x.y.z(-prerelease)?(+build)?" substring in a
 # larger string so we tolerate outputs like "codex-cli 0.145.0" or
-# "1.9.0 (build 42)". Same regex as
-# scripts/connector-version-radar.py:_VERSION_RE — keep them aligned;
-# radar imports this module now so the definitions stay single-source.
+# "1.9.0 (build 42)".
+#
+# scripts/connector-version-radar.py maintains its OWN copy of this
+# regex (see `_VERSION_RE` there). The radar is a standalone script
+# invoked outside the `defenseclaw` package import path (macOS lab
+# runners, cron jobs); making it import from this module would drag
+# the whole CLI dependency tree into every invocation. If either copy
+# ever diverges, the drift-guard test
+# ``test_semver_regex_matches_across_files`` in
+# ``cli/tests/test_agent_discovery.py`` trips — keep both definitions
+# byte-for-byte identical.
 _VERSION_RE = re.compile(
     r"(?<![0-9A-Za-z])v?(?P<major>\d+)\.(?P<minor>\d+)\.(?P<patch>\d+)"
     r"(?:-(?P<prerelease>[0-9A-Za-z.-]+))?"
--- a/cli/defenseclaw/inventory/_semver.py
+++ b/cli/defenseclaw/inventory/_semver.py
@@ -1,0 +1,179 @@
+# Copyright 2026 Cisco Systems, Inc. and its affiliates
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+"""Semantic-version parsing and comparison shared across the CLI.
+
+Lifted from scripts/connector-version-radar.py so agent_discovery.py can
+pick "highest supported version" among multiple installed copies of a
+connector (Claude Code / Codex / Cursor) without cross-importing a
+top-level script. Precomparison is strict semver 2.0 for the parts we
+care about: major.minor.patch plus dot-separated prerelease
+identifiers, ignoring build metadata.
+
+Only three public entry points intentionally: parse_version() to build a
+SemVersion from an arbitrary CLI / package.json string, SemVersion.compare()
+to totally order two of them, and pick_highest_supported() to select the
+best candidate from a list given a minimum inclusive bound.
+"""
+
+from __future__ import annotations
+
+import dataclasses
+import re
+from collections.abc import Iterable
+from functools import cmp_to_key
+
+# Matches the first "x.y.z(-prerelease)?(+build)?" substring in a
+# larger string so we tolerate outputs like "codex-cli 0.145.0" or
+# "1.9.0 (build 42)". Same regex as
+# scripts/connector-version-radar.py:_VERSION_RE — keep them aligned;
+# radar imports this module now so the definitions stay single-source.
+_VERSION_RE = re.compile(
+    r"(?<![0-9A-Za-z])v?(?P<major>\d+)\.(?P<minor>\d+)\.(?P<patch>\d+)"
+    r"(?:-(?P<prerelease>[0-9A-Za-z.-]+))?"
+    r"(?:\+(?P<build>[0-9A-Za-z.-]+))?(?![0-9A-Za-z])"
+)
+
+
+@dataclasses.dataclass(frozen=True)
+class SemVersion:
+    """A parsed semantic-version triple with optional prerelease chain."""
+
+    major: int
+    minor: int
+    patch: int
+    prerelease: tuple[str, ...] = ()
+
+    @property
+    def stable(self) -> bool:
+        return not self.prerelease
+
+    @property
+    def normalized(self) -> str:
+        base = f"{self.major}.{self.minor}.{self.patch}"
+        if self.prerelease:
+            return f"{base}-{'.'.join(self.prerelease)}"
+        return base
+
+    def compare(self, other: SemVersion) -> int:
+        own_core = (self.major, self.minor, self.patch)
+        other_core = (other.major, other.minor, other.patch)
+        if own_core != other_core:
+            return 1 if own_core > other_core else -1
+        # Per semver 2.0 §11: a version with prerelease has lower
+        # precedence than one without.
+        if not self.prerelease and not other.prerelease:
+            return 0
+        if not self.prerelease:
+            return 1
+        if not other.prerelease:
+            return -1
+        # Compare prerelease identifiers pairwise. Numeric identifiers
+        # order numerically; alphanumeric order lexically; numeric <
+        # alphanumeric when they collide.
+        for own, theirs in zip(self.prerelease, other.prerelease, strict=False):
+            if own == theirs:
+                continue
+            own_numeric = own.isdigit()
+            theirs_numeric = theirs.isdigit()
+            if own_numeric and theirs_numeric:
+                return 1 if int(own) > int(theirs) else -1
+            if own_numeric != theirs_numeric:
+                return -1 if own_numeric else 1
+            return 1 if own > theirs else -1
+        if len(self.prerelease) == len(other.prerelease):
+            return 0
+        return 1 if len(self.prerelease) > len(other.prerelease) else -1
+
+
+def parse_version(value: str, *, require_stable: bool = False) -> SemVersion:
+    """Extract and normalize the first semantic version in ``value``.
+
+    Raises ValueError when no x.y.z substring is present or when
+    ``require_stable`` is set and the parsed value carries a prerelease
+    tail. Callers that just want "best-effort or None" should use
+    ``parse_version_or_none``.
+    """
+    match = _VERSION_RE.search(value.strip())
+    if not match:
+        raise ValueError("no semantic x.y.z version found")
+    prerelease = tuple(filter(None, (match.group("prerelease") or "").split(".")))
+    parsed = SemVersion(
+        major=int(match.group("major")),
+        minor=int(match.group("minor")),
+        patch=int(match.group("patch")),
+        prerelease=prerelease,
+    )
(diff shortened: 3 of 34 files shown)


cjsanjay:
@coderabbitai review

cjsanjay:
## CodeRabbit review addressed

Pushed commit 9f934206 addressing 10 of 15 actionable findings. Summary below; full rationale in the commit message.

### 🟠 Major (all 4 fixed)

| Finding | Fix |
|---|---|
| `config_manager.go:63` envConfigPath race | Switched field to `atomic.Pointer[string]` + `getEnvConfigPath()` accessor; matches pattern already used by `ConfigManager.current`. |
| `ai_discovery.go:1531` unbounded MCP enumeration | Applied the same `maxItemsPerDirectory` bound the plugin/skill/rule detectors use. |
| `installer_lib.sh:810` shell/Go validator drift | Added `[::1]` to the shell loopback regex to match `validateAIDefenseEndpoint`. |
| `test_agent_version_multi_install.sh:94` host-dependent test | Switched to exercising the pure helper `_native_claudecode_version_from_dir` directly — CodeRabbit's `without_host_agent_bins` proposal wouldn't fully solve it either (only masks PATH, not `/opt/claude` file lookups). |

### 🟡 Minor (all 6 fixed)

- `ai_discovery.go:1422` — reject oversized `mcp.json` explicitly instead of mid-object truncation that guaranteed parse failure.
- `ai_discovery.go:1500` — comment now matches "disabled attribute is only emitted for disabled servers" instead of implying dashboards can filter on `disabled == "false"`.
- `ai_discovery.go:1593` — `dirItems` now pages until 256 *visible* entries or `dirItemsMaxScan` cap, so a directory whose first 256 names are dotfiles no longer silently drops the presence signal.
- `events.go:872` — `ItemName` / `ItemAttributes` doc updated to note they're gated by `privacy.disable_redaction` in `BuildAIDiscoveryPayload`.
- `agent_discovery.py:369` — removed the naive `_version_sort_key` that dropped prerelease tails and starved `pick_highest_supported` of stable candidates when a same-major-minor prerelease shipped alongside its release. Preserved the `current`-symlink priority via short-circuit. Added regression test.
- `installer_lib.sh:362` — Volta scoped-package glob now includes `lib/node_modules/<scope>/<pkg>/package.json` (previously missed every scoped Volta install).

### 🔵 Trivial / analysis-only (5 skipped)

Not addressed in this iteration — these are all "add sync-guard tests" / "share regex constants across scripts" / "add v2→v3 migration test" requests. Happy to spin those off as follow-up PRs if you want them tracked, but they're not correctness issues and expanding the PR further would slow review.

- `connector_paths.py:114` — add automated drift-guard test for `KNOWN_AGENT_KINDS` ↔ `promotedAgentKinds`.
- `_semver.py:47` — share `_VERSION_RE` between `_semver.py` and `scripts/connector-version-radar.py`.
- `agent_discovery.py:430` — validate the `home_dirs` field consumers land eventually (no current consumers).
- `ai_discovery_items_test.go:236` — add explicit v2→v3 state migration test.

### Test coverage this commit

- `packaging/macos/tests/run_tests.sh` now **151/151** (fixed two pre-existing test bugs unblocked by the above changes — details in the commit message).
- `pytest cli/tests/test_agent_discovery{,_multi_install}.py` — **37/37**, including the new prerelease regression guard.
- Go: `go build`, `go vet`, `gofmt`, and `go test ./internal/config/... ./internal/gateway/... ./internal/inventory/... ./internal/ipc/... ./internal/gatewaylog/...` all clean.

@coderabbitai review

cjsanjay:
## CodeRabbit round-2 fixes pushed (commit e6da29c1)

All 8 items in the second-round review addressed:

### 🔴 Critical — fd leak in dirItems

- `internal/inventory/ai_discovery.go` — my round-1 paged-read rewrite dropped `defer f.Close()`. Restored. Every scan was leaking one fd per candidate directory in a long-running gateway.

### 🟠 Major (5 fixed)

- `config_manager.go` retain endpoint on rejected overlay — after a valid EU overlay, a malformed env_config now no longer reverts the runtime endpoint to config.yaml's US default. That was the exfiltration vector the validator claims to defend against.
- `config_manager.go` independent env-config watch retry — added `ensureEnvConfigWatched()` that walks to the nearest existing ancestor when the target dir is missing, plus a 30s ticker so a late env_config drop (AVC arrives hours after install) surfaces without needing a config.yaml poke.
- `config_manager.go` watched-directory tracking — replaced the bool `envConfigDirWatched` with `envConfigWatchedDir string` so a `SetEnvConfigPath` change is observable by the retry path.
- `installer_lib.sh` PATH-shim multi-hop resolver — when `readlink -f` isn't available, the fallback now walks the symlink chain one hop at a time (16-step guard, short-circuits at the first SemVer-shaped mid-chain basename).
- `installer_lib.sh` portable SemVer comparator — added pure-bash `_semver_gt` implementing SemVer 2.0.0 §11.4 correctly (numeric identifiers compare numerically, alphanumeric lexically, no-prerelease > prerelease). Replaces every version-ranking `sort -V` call in `_native_claudecode_version_from_dir` / `_version_ge` / `_pick_highest_supported`. Sanity-tested against 10 spec cases:

    ```
    2.1.144 > 2.1.144-rc.1               ✓
    2.1.144-rc.1 !> 2.1.144               ✓
    2.1.144-rc.2 > 2.1.144-rc.1           ✓
    2.1.144-rc.10 > 2.1.144-rc.2          ✓
    2.1.144-rc.2 !> 2.1.144-rc.10         ✓
    3.0.0-alpha > 2.1.144                 ✓
    3.0.0 > 3.0.0-alpha                   ✓
    1.0.0 !> 1.0.0                        ✓
    2.5.0 > 1.9.0                         ✓
    2.1.144-alpha.beta > 2.1.144-alpha.1  ✓
    ```

### 🔵 Trivial (2 fixed)

- `config_manager_test.go:TestConfigManagerEnvConfigOverlayRejectsMalformed` — now applies a valid EU overlay first, then feeds each malformed payload and asserts the endpoint stays EU (not the config.yaml US default). Directly exercises the endpoint-retention fix above.
- Extracted `preserveManagedGatewayRuntimeFields` — the pre-diff preservation logic is now a named helper called from `Reload`, and the two `TestReloadPreserves…BeforeDiff` tests call the same helper instead of duplicating the field-copy logic locally. A regression inside the helper now actually reaches the tests.

### Test coverage this commit
- `go build`, `go vet`, `gofmt` clean.
- `go test -race ./internal/config/... ./internal/gateway/... ./internal/inventory/... ./internal/ipc/... ./internal/gatewaylog/...` — all pass under `-race`.
- `pytest cli/tests/` — 37/37 pass.
- `bash packaging/macos/tests/run_tests.sh` — 151/151 pass.

@coderabbitai review

cjsanjay:
## CodeRabbit round-3 fixes pushed (commit 264ed12b)

All 5 new items in the third-round review addressed:

### 🟠 Major (2 fixed)

- **`installer_lib.sh:_semver_gt` — three issues in one function**:
  1. **SC2206 unquoted array splits** — `am=(${a_main})` was word-splitting AND glob-expanding, so a hostile version like `2.1.144-*` could substitute working-directory filenames. Switched to `IFS=. read -ra`.
  2. **Extended suffix parsing** — previous `${a%%-*}` only recognised `-<pre>` tails; `semver_re` also accepts `.`, `_`, `+` (see `[._+-].*` in the regex). Added `_semver_gt_split` that peels the numeric prefix regardless of separator, then strips build metadata (SemVer §10).
  3. Sanity-verified against 15 spec cases including build-metadata equality (`2.1.144+build.5 == 2.1.144`), alternate separators (`2.1.144.dev`, `2.1.144_beta`), and the glob-hostile `2.1.144-*` (no wd expansion).

- **`installer_lib.sh:_native_claudecode_version_from_bin` — readlink -f fast-path fallthrough**: when the final basename isn't SemVer-shaped, now consults the mid-chain walk AND the parent-directory basename before conceding. Fixes the specific layout `shim -> .../versions/<X.Y.Z>/claude` where readlink -f canonicalises to the executable (basename=`claude`, fails regex) but the parent dir is version-labelled. macOS 12.3+ was silently returning empty for this on-disk layout; older macOS was finding it via the walk. Now consistent.

### 🔵 Trivial (3 fixed)

- **Ancestor-watch inert bug** (config_manager.go): the round-2 fallback added an ancestor-dir watch, but `classify()` dropped every event under it — only the 30s ticker could ever re-arm the target-dir watch. Added `envConfigWatchedAncestor` tracking + an event-loop branch that re-runs `ensureEnvConfigWatched()` on any Create event whose parent is the tracked ancestor. Net: a late mkdir of the env_config dir now surfaces as an immediate re-watch attempt, not "wait for the ticker."

- **SidecarHealth assertions** (config_manager_test.go): `TestConfigManagerEnvConfigOverlayRejectsMalformed` now constructs a real `SidecarHealth`, asserts each rejected overlay flips config to `StateError` with a non-empty `LastError`, and asserts a recovery overlay returns it to `StateRunning`. The health-write branches in Reload can no longer be silently removed without breaking this test.

- Round 3 also re-listed some round-1 items (KNOWN_AGENT_KINDS drift guard, `_semver.py` regex-share comment, `agent_discovery.py:410` home_dirs consumer, MCP enumeration bound already applied, `disabled` doc comment already applied). Those were already addressed in commits 9f934206 / e6da29c1 and the changes stand.

### Test plan (this commit)
- `go build`, `go vet`, `gofmt` clean
- `go test ./internal/config/... ./internal/gateway/... ./internal/inventory/... ./internal/ipc/... ./internal/gatewaylog/...` — all pass
- `pytest cli/tests/` — 37/37 pass
- `bash packaging/macos/tests/run_tests.sh` — 151/151 pass
- `_semver_gt` sanity-tested against 15 SemVer 2.0.0 cases
- `_native_claudecode_version_from_bin` manually exercised against the "shim → versions/X.Y.Z/claude" and "shim → current → versions/X.Y.Z" layouts

@coderabbitai review

cjsanjay:
@coderabbitai review status?

cjsanjay:
## Round-4 fixes pushed (commit 533b5b77) — all 5 previously-deferred threads resolved

Addressed every remaining unresolved CodeRabbit thread so the PR can move off `CHANGES_REQUESTED`:

- **KNOWN_AGENT_KINDS drift guard** — new `test_known_agent_kinds_matches_go_promoted_agent_kinds` parses `ai_catalog.go`'s map and asserts the connector-slug set matches `KNOWN_AGENT_KINDS − KNOWN_CONNECTORS`. Any add/remove/rename on either side trips it.

- **`_semver.py:47` regex-share comment** — corrected to describe the actual arrangement (radar keeps its own regex because it runs outside the CLI import path) and points at `test_semver_regex_matches_across_files` which enforces the two definitions stay byte-for-byte identical.

- **`agent_discovery.py:410` home_dirs consumer** — wired `AIDiscoveryConfig.home_dirs` through `discover_agents → _scan_agent → collectors`. Multi-tenant hosts can now enumerate every operator's installs from one sidecar; empty list preserves single-tenant fallback. Added `test_claudecode_home_dirs_enumerates_every_home` for coverage.

- **`ai_discovery_items_test.go:236` v2→v3 migration test** — new `TestAIStateStoreV2ToV3Migration_DropsItemLevelCategoriesOnly` writes a v2 state with one mcp_server (item-level, must drop) + one supported_connector (non-item-level, must preserve), asserts Load's behaviour, and also asserts a future-version file returns "unsupported version" (forward-compat guard).

- **`installer_lib.sh:_valid_aid_endpoint_url` parity re-verify** — re-tested against Go's `validateAIDefenseEndpoint`. All 3 loopback shapes (`localhost` / `127.0.0.1` / `[::1]`, with and without port) plus 8 rejection shapes match now. Added explicit loopback + bare-cisco.com cases to both `TestValidateAIDefenseEndpoint` (Go) and `t_resolve_aid_endpoint_precedence` (shell) so a future drift fails both suites — the Go side had zero loopback coverage before this commit.

### Test coverage this commit
- 3 new Go tests, 2 new Python tests, 8 new shell assertions.
- `go build`/`vet`/`gofmt` clean.
- Go test suite passes.
- `pytest cli/tests/` — 40/40 pass.
- `bash packaging/macos/tests/run_tests.sh` — 151/151 pass.

@coderabbitai review

cjsanjay:
## Round-5 fix pushed (commit c5c84fd7)

Fixed the only remaining CodeRabbit finding — the drift-guard test's naive '.index(")")' scan stopped inside the lookbehind, so it was only comparing 19 chars of the 147-char pattern.

Switched to `ast.parse` + walk to find `_VERSION_RE = re.compile(...)` and pull the full pattern argument. Verified locally the extraction now yields the complete 147-character pattern from each side (vs 19 under the old scan). Assertion also now prints both patterns on drift so a real failure is diagnosable from the log line alone.

@coderabbitai review

cjsanjay:
## Round-5/6 fixes pushed (commit 0055e4b4)

Addressed the 3 remaining actionable items from CodeRabbit's round-4 review:

- **`agent_discovery.py:702-707` catch `yaml.YAMLError`** — added alongside `OSError` in both `_ai_discovery_home_dirs` and `_ai_discovery_trust_config` so a malformed config.yaml falls back to the single-tenant default instead of raising inside the discovery path.
- **`test_claudecode_home_dirs_enumerates_every_home`** — now drives end-to-end through `discover_agents()` reading a real config.yaml under a caller-scoped `data_dir`. Prior iteration bypassed the loader by calling `_scan_agent` directly with `home_dirs=…`; the new shape exercises `_ai_discovery_home_dirs` and the forwarding path together.
- **v2→v3 migration test** — extended to cover all four item-level categories (mcp_server, plugin, rule, skill) with per-category fingerprint + drop assertions. A partial revert that dropped one category from `itemLevelCategories` no longer slips through.

The fourth finding (AST-based drift-guard extraction) was already fixed in the preceding commit c5c84fd7 before CodeRabbit's round-4 review posted.

@coderabbitai review

cjsanjay:
## Round-5b fix pushed (commit c4c6d728)

Fixed the drift-guard AST helper's fallback: previously used `ast.unparse` which serialises Python source syntax, so "a" + "b" + "c" on one side and "abc" on the other would compare unequal even though the compiled regex is identical.

Replaced with a small recursive `_eval_string_expr` that walks BinOp `Add` nodes over string constants and concatenates their values. Both shapes now resolve to the same evaluated string; genuinely dynamic construction (name refs, f-strings) still trips the 'unsupported literal shape' assertion so a real refactor surfaces cleanly.

Verified locally that a bare literal and an explicit concat shape produce identical strings via the helper. The existing test continues to pass on the real byte-identical patterns.

@coderabbitai review

cjsanjay:
## Round-6 fix pushed (commit 54e2e18a)

Fixed the sole remaining CodeRabbit finding — `_read_cache`'s post-KNOWN_AGENT_KINDS loop conflated two distinct cases (absent legacy entry vs present-but-corrupt entry) and reported a corrupt entry as `installed=False` until the TTL expired instead of triggering an immediate rescan.

Split into:
- `name not in raw_agents`: legacy — synth default for discovery-only kinds, reject-cache for enforcement connectors.
- present but not a dict: real corruption — always return None → rescan.

Added `test_corrupt_cache_entry_forces_rescan_not_synth_default` to pin the split.

@coderabbitai review

cjsanjay:
## Round-7 fix pushed (commit da2dfd87)

Fixed the hidden dependency CodeRabbit caught in my round-6 corrupt-cache test: the fixture left every KNOWN_CONNECTORS absent, so `_read_cache` bailed out on the first missing enforcement connector long before it ever touched the corrupt `aider` value. Old code took the same bailout → test passed against both implementations.

Populated every KNOWN_CONNECTORS + every discovery-only kind (except aider) with a valid dict. Now aider is the ONLY plausible reason to invalidate the cache, so a rescan proves the new branch fires.

@coderabbitai review

cjsanjay:
@coderabbitai — all previously-flagged findings are now addressed and you confirmed `da2dfd87` in your last message ("No further findings on this change"). Please post a formal approval so we can move this off CHANGES_REQUESTED.

vineethsai7 (approved):

PR #639, opened 2026-07-29
Fix Windows release state validation

kevinnobb:
## Problem

The exact Windows fresh-install release smoke can reject canonical deferred-uninstall state before publication. Release run 30488158730 failed because PowerShell StrictMode dereferenced an intentionally omitted `cleanup_boot_identifier`. A local Windows audit also found that the custody validator counted an inherit-only Creator Owner ACE as effective owner access.

## Root causes

- The Go cleanup, transaction, and hook-state contracts use `omitempty` for empty pending/uninstall/disabled fields, while the PowerShell validator directly dereferenced those absent properties under StrictMode.
- Creator Owner is an inheritance template; an inherit-only ACE provides no access to the directory itself and cannot satisfy the effective owner-access requirement.
- The native ACL regression fixture used `icacls /grant:r`, which replaces only the named trustee. An ambient explicit Administrators ACE survived on GitHub-hosted Windows and made the fixture non-hermetic.

## Solution

- Read canonical optional JSON fields through a StrictMode-safe property lookup while preserving nonempty scalar and array values for exact comparisons.
- Count current-user/OWNER RIGHTS and SYSTEM ACEs only when effective on the object; permit inherit-only Creator Owner only as a directory template and never as the required owner grant.
- Build native test ACLs from an exact protected SDDL, pre-seeding Administrators so ambient-ACE replacement is proven locally on every run.
- Add an opt-in Windows workflow job that authenticates and replays the current validator against an exact sealed candidate from a specified Actions run. It is manual-only and does not start or retry a release.

## Changed files

| File | Summary |
| --- | --- |
| `scripts/test-fresh-install-release-windows.ps1` | Harden effective ACL validation and make optional release-state reads StrictMode-safe. |
| `cli/tests/test_windows_release_certification.py` | Add native ACL, optional JSON, hermetic fixture, and exact-replay workflow contracts. |
| `.github/workflows/windows-native.yml` | Add the manual exact-candidate validator replay lane. |

## Local validation

- Focused Windows release certification: 25 passed.
- Neighboring Windows release/preflight selection: 114 passed, 11 platform-appropriate skips (125 collected).
- Native uninstall handoff: 77 passed on the prior product-fix head; the final additional change is test/workflow-only.
- Targeted deferred-cleanup, transaction, and hook-runtime Go tests: passed on the product-fix head.
- PowerShell parser, live Windows connector harness, Ruff, Python compile, and `git diff --check`: passed.
- CodeGuard manual review: clean. Its optional local sidecar was unavailable.

## Limitations

A full disposable standard-user release lifecycle requires elevation plus sealed/signing artifacts available on GitHub-hosted runners. This host has no installed WSL distribution, so Linux coverage remains CI-only. Local success reduces discovery risk but does not guarantee the release result.

### DIFF
--- a/.github/workflows/windows-native.yml
+++ b/.github/workflows/windows-native.yml
@@ -690,6 +690,8 @@ jobs:
     runs-on: windows-latest
     timeout-minutes: 50
     permissions:
+      # Download the sealed candidate from its producing run and check out the
+      # repository containing the replay validator and signature verifier.
       actions: read
       contents: read
     env:
--- a/.github/workflows/windows-native.yml
+++ b/.github/workflows/windows-native.yml
@@ -13,6 +13,23 @@ on:
   push:
     branches: [main]
   workflow_dispatch:
+    inputs:
+      release_candidate_replay_run_id:
+        description: Actions run ID that produced the sealed release candidate to replay.
+        required: false
+        type: string
+      release_candidate_replay_artifact:
+        description: Exact release-candidate artifact name from that run.
+        required: false
+        type: string
+      release_candidate_replay_version:
+        description: Canonical X.Y.Z version bound into the candidate.
+        required: false
+        type: string
+      release_candidate_replay_commit:
+        description: Exact lowercase 40-character commit bound into the candidate.
+        required: false
+        type: string
 
 permissions:
   contents: read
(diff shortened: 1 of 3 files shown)


kevinnobb:
Reproduced the release failure against the exact sealed 0.8.10 candidate from Actions run `30488158730` (`release-candidate-30488158730-1`, commit `10a998fbff1b4d77be8629b3099d8b868219dc4c`) without creating another release.

The replay exposed two additional PowerShell representation bugs after the original StrictMode failure:

1. A missing/null optional JSON string became scalar `$null`, so `$null -cne ""` falsely failed otherwise-valid cleanup state.
2. A Go nil connector slice serialized as JSON `null`; piping that directly through `ForEach-Object` produced one empty-string entry and falsely reported connector cleanup authority.

Both are now normalized at the optional-value boundary. No validator check was removed or weakened: real non-empty boot identifiers, connector entries, malformed identities, ACL/path/digest/signature mismatches, and cleanup residue still fail.

Evidence:

- Exact candidate replay passed end-to-end: https://github.com/cisco-ai-defense/defenseclaw/actions/runs/30496659529/job/90726981642
- Focused Windows release certification: `29 passed`
- Broader Windows/release regression set: `211 passed, 7 skipped`
- Added a reusable, authenticated workflow-dispatch replay path so future failed candidates can be retested directly instead of cutting diagnostic releases.

PR #646, opened 2026-07-30
Reject reparse points in Windows native capture

### LINKED ISSUE #645: Reject reparse points in Windows native diagnostic capture
## Summary
Windows native diagnostic capture currently discovers log files with recursive filesystem traversal and then reads each selected path with `ReadAllText`. A matching log filename inside the state root can be a reparse-point file, such as a symlink to content outside the intended diagnostic tree. Capture then follows that link and copies the target content into diagnostics.

## Severity
Medium. This requires write access to the isolated Windows native state root, but it can turn diagnostic capture into an unintended local file disclosure path in CI or support bundles.

## Affected files
- `scripts/windows-native-ci.ps1`

## Reproduction / context
1. Create a file outside the intended diagnostic capture tree containing sensitive local content.
2. Place a file symlink or other reparse-point file named `doctor.log` under the Windows native state root.
3. Run the Windows native `capture` operation.
4. The capture file selection can select the symlink, and `Invoke-Capture` reads the symlink target.

This was identified while reviewing the July 29, 2026 PR/commit scope, including PR #574 diagnostic changes and the current default-branch Windows native capture implementation.

## Recommended fix
Walk the capture tree without `Get-ChildItem -Recurse`, skip any directory or file with the `ReparsePoint` attribute before enqueueing or reading it, and add a self-test fixture proving a matching `doctor.log` symlink is excluded.

vineethsai7:
**Stack/base:** Standalone security fix rebased directly onto `main` at `c2236353f32e50f5c3c770f2fca16fca8c1cfca2`; no dependent pull requests.

## Problem

Windows native diagnostic capture enumerated candidate log files, rejected reparse points during discovery, and later reopened each selected pathname with `[IO.File]::ReadAllText(...)`. An attacker able to modify the state tree could replace a selected file or one of its ancestors between those operations and make capture read a different filesystem object. The candidate traversal also stopped globally after 4,096 entries, so sufficiently many decoys could prevent priority diagnostics from being considered, and its `$matches` local collided case-insensitively with PowerShell's automatic `$Matches` variable.

Fixes #645.

## Current Situation

The original PR rejected reparse points visible during enumeration, but review and the first Windows CI run exposed three remaining gaps:

1. `$matches` became the automatic regex-result hashtable after `-match`, so the harness failed at `.Add(...)`.
2. The global traversal cap could discard later `wizard-driver.log`, `go-test-failure-summary.log`, or `go-test.log` files before priority selection.
3. `ReadAllText(path)` performed a fresh path resolution after discovery, leaving a swap/reparse TOCTOU window.

### Exhaustive traversal with bounded retention

The capture walker now examines the complete non-reparse tree while retaining only the best 30 candidates in an ordinal-ignore-case `SortedDictionary`. Priority diagnostics therefore remain selectable regardless of earlier decoys, without retaining an unbounded result list. The conflicting `$matches` and obsolete `$visited` variables are gone.

### Handle-bound, retained-root reads

`DisposableFileGuard.RootedReader` retains a validated no-follow handle to the capture root. For every selected file it:

1. verifies the candidate is lexically beneath that root;
2. opens the leaf with `FILE_FLAG_OPEN_REPARSE_POINT`;
3. validates disk-file type, attributes, single-link identity, and byte bound from that same handle;
4. obtains the final path from that same handle and requires it to remain beneath the retained root; and
5. reads the exact validated byte count through that handle.

Any disappeared file, reparse replacement, outside-root ancestor swap, size mutation, or other validation failure is skipped rather than captured.

```mermaid
flowchart LR
    A["Exhaustive non-reparse enumeration"] --> B["Bounded priority set: 30 files"]
    B --> C["Retained verified root handle"]
    C --> D["Open candidate no-follow"]
    D --> E["Same-handle attributes, link count, size, and final path"]
    E -->|"valid child"| F["Same-handle bounded read"]
    E -->|"swap, reparse, or escape"| G["Reject and skip"]
    F --> H["Redact and write diagnostics"]
```

### Deterministic race regressions

The Windows native self-test now selects a legitimate diagnostic and then deterministically replaces (a) the leaf with a symbolic link and (b) an ancestor directory with a junction to an outside directory. Both reads must be rejected. Static certification tests lock in the handle-bound contracts and the bounded priority selector.

## What Changed (Where & How)

| Area | Path | Change |
|---|---|---|
| Native capture | `scripts/windows-native-ci.ps1` | Replaced the conflicting/unbounded traversal state, routed reads through one retained-root reader, and added deterministic leaf/ancestor swap self-tests. |
| Windows file guard | `scripts/windows-disposable-file-guard.cs` | Added retained-root/no-follow file reads with same-handle identity, attribute, final-path, size, and exact-read validation. |
| Live connector harness | `scripts/live-connector-e2e/test-windows.ps1` | Added static contracts for bounded selection, no pathname reopen, retained-root enforcement, and swap regressions. |
| Release certification | `cli/tests/test_windows_release_certification.py` | Added cross-platform source assertions covering the security invariants and test fixtures. |

## Breaking Changes

None. Ordinary in-root UTF-8 diagnostic files retain the existing 1 MiB per-file and 30-file bounds. Files that cannot be proven to be regular, single-link children of the retained state root are now skipped.

## Open Issues / Follow-ups

None.

## Test Plan

- `uv run --frozen pytest -q cli/tests/test_windows_release_certification.py cli/tests/test_connector_live_e2e_path_policy.py cli/tests/test_windows_installer_artifacts.py`
  - Passed: `102 passed, 61 skipped in 3.69s`.
- `uv run --frozen ruff check cli/tests/test_windows_release_certification.py`
  - Passed: `All checks passed!`
- `uv run --frozen ruff format --check cli/tests/test_windows_release_certification.py`
  - Passed: `1 file already formatted`.
- `actionlint .github/workflows/windows-native.yml`
  - Passed with no findings.
- `git diff --check`
  - Passed with no whitespace errors.
- Native Windows PowerShell self-test
  - Not runnable in the local macOS worktree because neither Windows PowerShell nor `pwsh` is installed.
  - GitHub Actions [`Windows PowerShell harness validation`](https://github.com/cisco-ai-defense/defenseclaw/actions/runs/31457851987/job/93675255485) passed, including C# compilation and the deterministic capture self-tests.
  - GitHub Actions [`Windows Native Required`](https://github.com/cisco-ai-defense/defenseclaw/actions/runs/31457851987/job/93681052339) passed.
- Full PR check set at commit `4fc57ccadae1b8482fd5d10afcd1868ba941bf78`
  - Terminal result: `87 passed`, `4 intentionally skipped`, `0 failed`, `0 pending`.

### DIFF
--- a/cli/tests/test_windows_release_certification.py
+++ b/cli/tests/test_windows_release_certification.py
@@ -28,6 +28,7 @@ SMOKE_PATH = ROOT / ".github" / "workflows" / "release-candidate-smoke.yml"
 WINDOWS_NATIVE_PATH = ROOT / ".github" / "workflows" / "windows-native.yml"
 FRESH_INSTALL = (ROOT / "scripts" / "test-fresh-install-release-windows.ps1").read_text(encoding="utf-8")
 DISPOSABLE_LAUNCHER = (ROOT / "scripts" / "invoke-windows-setup-standard-user-ci.ps1").read_text(encoding="utf-8")
+DISPOSABLE_FILE_GUARD = (ROOT / "scripts" / "windows-disposable-file-guard.cs").read_text(encoding="utf-8")
 STANDARD_USER_PROCESS_LAUNCHER = (ROOT / "scripts" / "windows-disposable-standard-user-launcher.cs").read_text(
     encoding="utf-8"
 )
@@ -643,6 +644,30 @@ def test_disposable_setup_failure_preserves_bounded_native_log_before_profile_cl
     assert "Copy-Item" not in DISPOSABLE_LAUNCHER
 
 
+def test_native_diagnostic_capture_is_bound_to_verified_file_handles() -> None:
+    selection = _function("Get-WindowsNativeCaptureFiles")
+    capture = _function("Invoke-Capture")
+
+    assert "SortedDictionary[string, IO.FileInfo]" in selection
+    assert "$selectionLimit = 30" in selection
+    assert "$matches" not in selection.casefold()
+    assert "$visited" not in selection
+    assert "OpenRootedReader($root)" in capture
+    assert "ReadBoundedUtf8($file.FullName, 1048576)" in capture
+    assert "ReadAllText($file.FullName)" not in capture
+
+    for contract in (
+        "FILE_FLAG_OPEN_REPARSE_POINT",
+        "GetFileInformationByHandle",
+        "GetFinalPathNameByHandleW",
+        "sealed class RootedReader",
+        "guarded file resolved outside its retained root",
+    ):
+        assert contract in DISPOSABLE_FILE_GUARD
+    assert "leaf replaced by a reparse point after enumeration" in HARNESS
+    assert "replaced ancestor outside its retained root" in HARNESS
+
+
 def test_publish_includes_windows_binaries_without_an_omission_mode() -> None:
     workflow = _workflow(RELEASE_PATH)
     jobs = workflow["jobs"]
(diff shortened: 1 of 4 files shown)


vineethsai7:
Security/CI review note: the current required Windows Native CI run is red on this PR. The failing job is `Windows PowerShell harness validation`, and the log shows `windows-native-ci.ps1: Cannot find an overload for "Add" and the argument count: "1"` before `Windows Native Required` fails. Please fix the capture enumeration path in `scripts/windows-native-ci.ps1` or rerun after correcting it; this should not merge while the required Windows gate is failing.

PR #654, opened 2026-07-30
fix(scanner): make DNS pinning work with MCP SDK transports

### LINKED ISSUE #359: F-0344: MCP-scanner SDK DNS-rebind residual — verify pinned_getaddrinfo covers all SDK transports
## Summary
The stacked security PR adds `pinned_getaddrinfo` (in `cli/defenseclaw/registries/ssrf.py`) and wraps the remote MCP scan (`scanner/mcp.py`) so the SDK's connect-time DNS resolution is pinned to the IP validated by `resolve_and_pin`, closing the DNS-rebind TOCTOU for the async-httpx MCP scanner SDK.

## Open concern
The pin operates at the `socket.getaddrinfo` chokepoint, which covers both urllib3 and async httpx/anyio. But this should be re-verified against the *actual* mcp-scanner SDK transports (`sse_client`, `streamablehttp_client`) end-to-end with a live rebinding resolver, not just the unit-level primitive test, in case a transport pre-resolves or caches addresses outside getaddrinfo.

## Acceptance
- Integration test that drives a real SDK remote scan against a low-TTL rebinding DNS stub and asserts the connect lands on the vetted IP (or is refused), not the rebound address.

Tracked from the stacked remediation PR's Open Issues.

barryqy:
> Base note: This existing cross-repository PR was rebased onto `main` at `c2236353f32e50f5c3c770f2fca16fca8c1cfca2` on 2026-08-11. The current head is `f20bf80e3111c31a75619b95daa32350d4d5b61c`.

## Problem

Remote MCP scans validate a hostname and pin its approved IP before invoking the MCP Scanner SDK, but the SDK's HTTPX/AnyIO transports resolve through the active event loop. On uvloop and other cached resolver paths, changing only `socket.getaddrinfo()` does not reliably reach that transport resolver.

AnyIO also passes an IDNA-encoded `bytes` hostname. The old pin compared it using `str(node)`, turning `b"host"` into the literal string `"b'host'"` and rejecting the already-vetted target. The wrapper could additionally pass a forced address family positionally while retaining a caller-supplied `family=` keyword, producing `TypeError`.

The first analyzer isolation fix then created a second security boundary problem: every DNS lookup made under an analyzer task bypassed the pin without validating its returned addresses. A redirect, proxy, or provider-side lookup could therefore resolve an unconfigured hostname to loopback, RFC1918, link-local, CGNAT, cloud metadata, or another non-public address.

Fixes #359.

## Current Situation

Both MCP SDK transports exercise the affected async resolver path:

- streamable HTTP (`/mcp`)
- server-sent events (`/sse`)

On the rebased pre-review head (`8a12e920`), a focused source-to-sink harness entered the analyzer scope, resolved `redirect.example` to `127.0.0.1`, and observed `VULNERABLE_PRE_FIX_RESULT 127.0.0.1`. On the repaired head, the same lookup raises `SSRFError`.

### Pin the real async transport resolver

- Canonicalize validation-time and connect-time hostnames to the same ASCII representation AnyIO uses.
- Decode resolver `bytes` strictly as ASCII and encode Unicode hostnames with IDNA/UTS46.
- Override the active event loop's `getaddrinfo()` while an SDK transport runs, routing it through the process-wide socket pin.
- Reference-count overlapping async pin scopes per event loop and restore the exact previous resolver after the last scope exits, including cancellation and exception paths.
- Create each SDK coroutine only after the async resolver scope enters, so a failed context entry cannot leave an un-awaited coroutine behind.
- Preserve requested port, socket type, protocol, and flags while forcing the family that matches the vetted IP.

### Keep analyzer networking usable without reopening SSRF

- Give each analyzer a task-local DNS policy.
- Treat only the operator-configured API or LLM endpoint hostname as trusted; this preserves explicitly configured private analyzer endpoints.
- When a local LiteLLM provider has no explicit base URL, mark its default spellings (`localhost`, `127.0.0.1`, and `::1`) loopback-only and reject every non-loopback answer; cloud-provider defaults remain public-only.
- Resolve other analyzer hosts once at the connection-time resolver boundary and reject every answer that is loopback, private, link-local, multicast, unspecified, reserved, CGNAT, cloud metadata, or IPv4-mapped unsafe space.
- Keep concurrent MCP transport lookups outside the analyzer context fail closed.

```mermaid
flowchart LR
    A[Remote MCP URL] --> B[Resolve and validate target]
    B --> C[Vetted numeric IP]
    C --> D[Event-loop resolver override]
    D --> E[MCP SDK SSE / streamable HTTP]
    E --> F[Dial vetted IP only]

    E --> G[Analyzer task-local DNS policy]
    G --> H{Explicitly configured endpoint?}
    H -->|Yes| I[Allow exact operator-selected host]
    H -->|No| M{Implicit local default?}
    M -->|Yes| N[Require every answer to be loopback]
    M -->|No| J[Resolve and validate every address]
    J -->|Public only| K[Allow]
    J -->|Private / metadata / unsafe| L[Raise SSRFError]
```

## What Changed (Where & How)

| Path | Change |
|---|---|
| `cli/defenseclaw/registries/ssrf.py` | Added AnyIO-compatible hostname canonicalization, event-loop resolver pinning with overlap-safe restoration, module-scope IDNA import, and task-local analyzer DNS policies that distinguish explicit trust, loopback-only defaults, and public-only unconfigured destinations. |
| `cli/defenseclaw/scanner/mcp.py` | Routed all four remote SDK scan phases through lazily created coroutines inside the async pin and wrapped API/LLM analyzers with their configured endpoint policy plus loopback-only local-provider defaults. |
| `cli/tests/test_registry_ssrf.py` | Added byte-host, IDNA, argument-shape, IPv4/IPv6, malformed-host, task-isolation, default-asyncio/uvloop overlap, cancellation, pre-existing resolver restoration, public analyzer, configured-private analyzer, and unsafe/mixed-answer regressions. |
| `cli/tests/test_mcp_dns_pinning.py` | Added real SDK tests for SSE and streamable HTTP with a rebinding resolver; removed cross-event-loop rendezvous; made teardown preserve the primary test failure; covered context-entry failure without an un-awaited coroutine; proved real loopback defaults remain usable while poisoned localhost, cloud defaults, and unconfigured private redirects fail closed. |

## Breaking Changes

None. Explicitly configured private analyzer endpoints remain supported. Unconfigured analyzer redirects or proxy/provider lookups that resolve to non-public address space now fail closed by design.

## Open Issues / Follow-ups

None.

## Test Plan

- `make py-lint`
  - Passed: Ruff reported `All checks passed!` for `cli/defenseclaw/`.
- `uv run --frozen pytest -q cli/tests/test_registry_ssrf.py cli/tests/test_mcp_dns_pinning.py`
  - Passed: `69 passed, 23 subtests passed in 4.01s`.
- `uv run --frozen pytest -q cli/tests/test_cmd_mcp.py cli/tests/test_cmd_mcp_scan_ux.py cli/tests/test_connector_mcp_writers.py cli/tests/test_mcp_dns_pinning.py cli/tests/test_mcp_llm_lane.py cli/tests/test_mcp_scanner_env.py cli/tests/test_mcp_windows_launchers.py cli/tests/test_registry_ssrf.py cli/tests/test_remediation_mcp_registry.py cli/tests/tui/test_mcp_set_form_screen.py`
  - Passed: `405 passed, 23 skipped, 82 subtests passed in 13.75s`.
- Pre-fix/repaired focused analyzer-DNS harness against `8a12e920` and `4757974a`
  - Pre-fix: `VULNERABLE_PRE_FIX_RESULT 127.0.0.1`.
  - Repaired: `FIXED_CURRENT_RESULT SSRFError`.
- `git diff --check origin/main...HEAD`
  - Passed with no output.

### DIFF
--- a/cli/defenseclaw/registries/ssrf.py
+++ b/cli/defenseclaw/registries/ssrf.py
@@ -19,10 +19,10 @@ need to ingest from a private host must opt in with ``--allow-private``;
 the flag is plumbed all the way through the CLI surface so an
 "oops, my URL was internal" mistake stays explicit.
 
-This module is intentionally pure-Python and stdlib-only so it can be
-unit-tested without network access — :func:`guard_url` performs DNS
-resolution via :func:`socket.getaddrinfo`, but that lookup can be
-mocked in tests by passing a custom ``resolver`` callback.
+DNS resolution is dependency-injected so the guard can be unit-tested
+without network access. :func:`guard_url` uses
+:func:`socket.getaddrinfo` by default, but tests can pass a custom
+``resolver`` callback.
 """
 
 from __future__ import annotations
@@ -122,6 +122,37 @@ class SSRFError(ValueError):
 Resolver = Callable[[str], list[str]]
 
 
+def _canonical_dns_host(host: str | bytes | None) -> str:
+    """Return the ASCII hostname representation used by AnyIO."""
+    if isinstance(host, bytes):
+        try:
+            value = host.decode("ascii")
+        except UnicodeDecodeError as exc:
+            raise SSRFError("DNS host contains non-ASCII bytes") from exc
+    elif isinstance(host, str):
+        value = host
+    else:
+        raise SSRFError("DNS host must be text or ASCII bytes")
+
+    value = value.strip()
+    if not value:
+        raise SSRFError("DNS host is empty")
+    if any(ord(char) < 32 or ord(char) == 127 for char in value):
+        raise SSRFError("DNS host contains control characters")
+
+    try:
+        encoded = value.encode("ascii")
+    except UnicodeEncodeError:
+        import idna
+
+        try:
+            encoded = idna.encode(value, uts46=True)
+        except idna.IDNAError as exc:
+            raise SSRFError(f"invalid internationalized DNS host {value!r}") from exc
+
+    return encoded.decode("ascii").lower()
+
+
 def _default_resolver(host: str) -> list[str]:
     """Resolve *host* via getaddrinfo, returning a list of literal IPs.
 
@@ -196,9 +227,10 @@ def resolve_and_pin(
         raise SSRFError(
             f"unsupported URL scheme {scheme!r}; expected http or https"
         )
-    host = (parsed.hostname or "").strip().lower()
-    if not host:
+    parsed_host = parsed.hostname
+    if not parsed_host or not parsed_host.strip():
         raise SSRFError("URL is missing a host component")
+    host = _canonical_dns_host(parsed_host)
 
     if host in {"localhost", "ip6-localhost"} and not allow_private:
         raise SSRFError("refusing to fetch from localhost (use --allow-private to opt in)")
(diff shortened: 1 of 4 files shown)


barryqy:
Live retesting of the first revision exposed a second resolver boundary: the MCP target remained pinned correctly, but the API analyzer's lookup of `us.api.inspect.aidefense.security.cisco.com` was rejected by the same process-wide pin, so MCP Scanner logged an analyzer failure and still reported the target as clean.

Follow-up commit `89d758a7` isolates only the SDK API/LLM analyzer tasks with task-local DNS permission. MCP transport tasks, redirects, and proxy lookups remain pinned and fail closed. The active event-loop resolver is routed through the socket guard so this also holds under uvloop, with exact restoration and overlapping-scope coverage.

Validation:
- focused suite: 60 passed, 18 subtests passed
- MCP/registry regression suite: 380 passed, 22 skipped, 77 subtests passed
- Python lint and `git diff --check` passed
- live DNS probe kept `3-145-87-180.sslip.io` pinned to `3.145.87.180` while the Cisco analyzer endpoint resolved normally

vineethsai7:
Security/CI review note: the code review found no additional DNS-pinning bypass beyond the existing tracked analyzer-DNS comments, but the PR is still merge-blocked by CI. `Go Gateway Test (shard 3)` timed out in `TestSidecarBootstrapLocalObservabilityCanaryReachesAgent360Projection`, which makes the aggregate `Go Test` check fail. Please rerun or isolate that timeout before taking this out of draft/merging.

vineethsai7:
@coderabbitai review

PR #658, opened 2026-07-31
feat(windows): add managed enterprise hardening parity

vineethsai7:
## Summary

DefenseClaw already provides an administrator-owned `managed_enterprise` deployment model on Linux and macOS. Native Windows previously retained only the ordinary per-user lifecycle, so a standard-user AI agent could terminate its own per-user gateway and persistently change its own integration without a Windows system-service owner restoring the protected state.

This PR adds native Windows enterprise parity with Windows-native controls:

- `DefenseClawGateway` runs as a restricted `NT SERVICE\DefenseClawGateway` virtual service account;
- `DefenseClawHookGuardian` runs as LocalSystem and repairs only administrator allow-listed users and connectors;
- binaries, managed configuration, the target manifest, service definitions, runtime credentials, deployment metadata, and guardian authorization are protected with administrator/System ownership and explicit NTFS/SCM/registry DACLs;
- the guardian uses the exact active, non-elevated WTS token for the manifest-pinned SID when it changes that user's footprint;
- event-driven repair plus a periodic backstop restores deleted or modified managed hooks and supported native-agent policy;
- a complete Windows lifecycle is exposed through the public CLI: install, upgrade, repair, reconcile, status, verify, uninstall, and purge.

Two compatibility rules are intentional and tested:

1. **Enterprise is explicit opt-in.** Merely installing an enterprise-capable release is a no-op for machine scope. Unless the effective configuration is `managed_enterprise` and an administrator runs `defenseclaw enterprise windows install`, DefenseClaw creates no SCM service, moves no state into `ProgramData`, and transfers no user footprint to the enterprise guardian.
2. **Normal-mode auto-heal remains intact.** Unmanaged Windows keeps the existing per-user startup, connector setup, application-protection repair, and hook self-heal behavior. Managed mode changes who owns repair; it does not introduce auto-heal by removing it from normal mode.

The security objective is outcome parity with Linux/macOS, expressed through the native platform: SCM services instead of systemd/launchd, service/process/token security descriptors instead of POSIX ownership alone, fixed-volume NTFS identity checks instead of UNIX mount checks, and exact SID/token impersonation instead of UID/GID switching.

## Parity and compatibility

| Security property | Linux | macOS | Windows in this PR |
| --- | --- | --- | --- |
| Gateway lifecycle | Administrator-owned systemd service | Administrator-owned LaunchDaemon | Administrator-owned SCM service |
[...]

PR #673, opened 2026-08-06
fix(windows): repair enterprise hardening install and verify defects

ihabler:
Stacked on top of #658 (`codex/windows-enterprise-hardening` @ `80a2cae`). Twenty-eight defects found across five rounds of certifying `managed_enterprise` hardening on Windows Server 2022. Most stop the install or the gateway outright, and all but one reproduce on a signed build.

Targets #658's branch rather than `main`: both PowerShell files are new in that PR, and the ancestor fix delegates to a helper #658 introduces.

Grouped by cause rather than by discovery order, since several were found separately and share one fix.

## 1. The ancestor permission rule rejected stock Windows

Two code paths checked every parent folder above a protected file and treated "write attributes" as dangerous. Windows grants exactly that to all users on `C:\ProgramData`, which sits above everything, so the gateway would not start and enrolling a user failed with `untrusted Windows principal S-1-5-32-545 has write-like access mask 0x116`.

**Fix:** both paths use `managed.WindowsAncestorReplaceAccess`, which rejects only the rights that genuinely let someone replace a protected child: delete, delete-child, write-DAC, write-owner, and the generic bits covering them. Leaf masks are unchanged, and `C:\ProgramData` is treated as the ancestor it is. Every ancestor must still be owned by Administrators, LocalSystem, or TrustedInstaller, carry a non-null DACL, and contain no reparse point.

`Everyone` is excluded from the relaxation and held to the leaf rule, since stock Windows grants these rights to `BUILTIN\Users` and Authenticated Users but never to Everyone. A write grant to Everyone on an ancestor is therefore always deliberate, and still refused.

**Worth a look:** this relaxes an assumption #658 states — that no untrusted principal holds *any* write-like right on any ancestor — because no Windows host satisfies it. `BUILTIN\Users` and Authenticated Users can now hold add-file or write-attributes on an ancestor without being reported. Adding a name to a directory cannot replace an existing child; that needs delete on the child or delete-child on the parent, both still rejected. Authenticated Users has to be exempted rather than pinned like Everyone, because some Windows SKUs grant it write-like rights on `C:\ProgramData` directly.

> Pre-existing on `main`, not introduced by #658, but the correct shared rule only exists on this branch.

## 2. The gateway couldn't traverse to its own config

It runs as a virtual service account, so it is in neither SYSTEM nor Administrators. It had read on the config file but no way through the folders above it: the parent was administrator-only, and so were `C:\ProgramData\Cisco` and `...\DefenseClaw-Cert`.

**Fix:** a `ConfigDirectory` ACL kind grants the service `ReadAndExecute` on the config folder, and each directory between `C:\ProgramData` and the state root gets a traverse grant. That grant is dropped when the gateway holding it is deleted, by uninstall or by a rolled-back install, so a shared folder does not collect entries for service accounts that no longer exist.

**Worth a look:** those in-between folders are shared vendor directories another Cisco product may own, so the grant adds one non-inherited entry and leaves the owner and every other entry alone. Applying the canonical protected DACL there would have taken ownership and dropped that product's access. Not inheriting keeps the grant off sibling trees.

## 3. The gateway rejected state its own installer created

The installer creates runtime directories owned by Administrators and shared with the service through a narrow writer entry. The gateway validated them with `safefile`, which requires sole ownership: `refusing foreign-owned directory`. It has no `WRITE_OWNER` to correct that, and `safefile` would have stripped the Administrators entry if it had. First-boot token persistence hit the same wall.

**Fix:** managed deployments are validated against the managed trust model, which expects that layout, and published in place under the installer's DACL. All five writes under the gateway data directory moved over, not just the one that failed. Self-managed installs keep `safefile` unchanged.

## 4. A deployment could be installed but not removed

Five separate reasons, each permanent:

- `-AllowUnsigned` was rejected for `Uninstall` at five gates, while the unsigned module was rejected without the flag.
- Teardown demanded the services be Automatic, so one an operator had set to Disabled or Manual was stuck. Start type is boot policy, not authorization.
- A Codex-only deployment tears down an empty set of Claude targets, rejected as corrupt metadata before the transaction opened — leaving the correct "no targets, nothing to do" path a few lines below unreachable.
- Rollback and uninstall refused with `refusing to remove non-empty transaction-created shared directory`. The only thing in the folder was DefenseClaw's own zero-byte serialization lock, which outlives the policy files it guards, so the transaction failed closed against its own bookkeeping.
- A fresh install always records an ACL preimage, noting that `requirements.toml` did not exist beforehand. Only the owned-removal path deleted that record, so once a teardown reached a verified-absent disposition it threw on the leftover forever, with no path left that could clear it.

**Fix:** all five gates accept `Uninstall`; teardown accepts any supported start mode; an empty Claude target set reaches the absent-policy path, while a policy that exists must still match its recorded targets exactly; that one lock path is cleared before the emptiness check, only as a protected regular file in the directory being removed; and a preimage recording no prior file leaves with the deployment, while one recording a file that predates DefenseClaw still stops the teardown. Other residue still fails closed, and everything that is authorization — image path, account, SID type, privileges, environment, ACLs — is still asserted.

## 5. The hook launcher check could never pass on an enterprise install

The gateway only recognised a launcher through per-user package state, which an operator-rooted tree never has, so it fell back to `~/.local/bin` and the guardian's comparison against the real installed binary could not be satisfied.

**Fix:** a managed-enterprise branch accepts the administrator-owned launcher installed beside the running gateway, after proving no untrusted writer on it or any ancestor. Tests pin the two sides of the comparison from separate sources so it stays falsifiable.

## 6. Registering hooks for a user failed under impersonation

Two causes, both from a LocalSystem service doing per-user work under a target-user thread token.

Ownership of the token files was assigned from the impersonating thread, which does not hold `SeRestorePrivilege`: `This security ID may not be assigned as the owner of this object`.

Six other places read the identity from the *process* token while a thread token was in effect. Windows takes a new file's owner from the effective token, so the guardian stamped descriptors naming LocalSystem onto files the target user owned, then validated them against LocalSystem as well — no per-user artifact the two exchanged could be accepted. One of the six is worse than a stable mismatch: the protection witness normalizes the two interchangeable allow entries only when the owner matches the identity it read, so under impersonation it also lost its tolerance for NTFS reordering them during a rename, and mismatched only when that reordering happened.

**Fix:** the ownership assignment runs on a locked OS thread holding the privilege from the process token, then reverts, mirroring the DACL repair path. Every per-user identity read in the connector package goes through one accessor returning the thread token's user when one is installed and the process user otherwise, so the two sides cannot diverge.

**Worth a look:** the existing test helper impersonated the same user, which makes both readings agree and the defect invisible. The new tests pin the accessor to a second principal and assert that the descriptor a creator emits names the owner its validator demands, so the pair is falsifiable without a live host and a second account.

## 7. Verification rejected what the installer had just produced

`sc.exe sdshow` returns the permissions *and* the audit settings, but the whole string was compared against a permissions-only value, so service permissions always read as tampered with. Separately, `sc.exe sdset` creates a `Security` subkey with its own locked-down permissions, while the verifier insists that key inherit from its parent.

**Fix:** the comparison splits at `S:` and checks only the DACL. The SACL is deliberately unasserted — SCM owns it, DefenseClaw never sets one, and a live host had `W32Time` with no audit section at all. The subkey's explicit entries are dropped and inheritance re-enabled after the parent DACL is written; the value `sdset` wrote is untouched.

**Worth a look:** that grants standard users *read* on the subkey where `sdset` granted them nothing. I believe it is the intent — the parent ACL is written `CI`, the verifier demands inheritance, and SCM already exposes the service DACL to unelevated users — but it is a deliberate loosening, so flagging it.

## 8. The installer couldn't run on PowerShell 7

The script hardcoded the 5.1 module folder, so under 7 it could not resolve its own qualified module calls. PowerShell 7 also materializes the guardian's UTC timestamp as a `DateTime`, and casting it back to text reformats it in local culture and drops the `Z`, after which it is rejected for having no timezone. And three places read a `state` field without checking it existed, which throws under `StrictMode -Version Latest`.

**Fix:** the pinned `PSModulePath` derives from `$PSHOME`, producing the identical path on 5.1; timestamps are re-emitted as `yyyy-MM-ddTHH:mm:ssZ` in UTC; the field is probed before reading.

## 9. Errors named the wrong cause

Any config load failure was wrapped as "run `defenseclaw upgrade` first", a command that does not exist on Windows. An access-denied while inspecting a configured path was rendered as a schema error with a fabricated line and column, against a valid document. A failing managed-hook teardown target discarded its own error. And artifact hash drift, which blocks every lifecycle action through one shared gate, reported itself as a repair-and-upgrade concern even to an operator running `Uninstall` — who then found that repair refuses the same drift.

**Fix:** the real config error is surfaced unless the file genuinely declares the wrong `config_version`; filesystem failures report as filesystem failures with no source position, since no edit to the document would fix one; the per-target cause is included, with control characters folded; and the drift error names the action that hit it and the upgrade that clears it.

## 10. Smaller fixes, each of which stopped a run

- `Select-Object` was qualified as `Microsoft.PowerShell.Core\Select-Object`; it lives in `Microsoft.PowerShell.Utility`.
- Walking up to the drive root produced a bare `C:`, which means "wherever I currently am on that drive" — on-host, `Get-Item 'C:'` returned `C:\Users\Administrator`.
- The hash writer records a `codex_launcher` entry but the verification switch had no case for it, so every install ended with `installed artifact recorded in metadata is missing: codex_launcher`.
- `TEMP` is redirected into the installer's own working folder, so entries can vanish mid-sweep and the resulting error overwrote the exit code of an install that had already succeeded.
- Three registry ACL calls used `-LiteralPath`, which the registry provider mishandles. The 5.1 failure is non-terminating, so `$acl` silently became `$null` and surfaced later as a confusing null dereference.
- The documented Windows example and several fixtures used `config_version: 7` with top-level `audit_db` and `judge_bodies_db`, keys v8 rejects, so anyone following the docs got a config the gateway refuses.
- Observe mode rendered `would block this in action mode a HIGH Codex hook finding`, which cannot parse. The prefix operators alert on is unchanged.
- `cryptography`, `aiohttp`, and `js-yaml` were raised past advisories published during this work.


ihabler:
Since the last round on `15022bea`, three more defects are fixed and CI is green on `d2cde088`:

- **Uninstall was impossible on a normal fresh install.** The installer always records an ACL preimage saying `requirements.toml` did not exist beforehand, and only the owned-removal path deleted that record. Once a teardown reached a verified-absent disposition it threw on the leftover with nothing able to clear it. A preimage recording no prior file now leaves with the deployment; one recording a file that predates DefenseClaw still stops the teardown.
- **Uninstall and repair pointed at each other after any in-place binary swap.** One shared hash-drift gate reported itself as a repair/upgrade concern regardless of caller. It now names the action that hit it and the upgrade that clears it. No `-Force` bypass was added — that gate is what ties installed bytes to signed metadata.
- **Observe mode emitted a sentence that could not parse.** The prefix operators alert on is unchanged.

Worth re-testing: uninstall on a deployment where DefenseClaw created the Codex requirements file, and uninstall after an in-place binary replacement.

PR #674, opened 2026-08-06
feat(observability): per-item skill/plugin/MCP inventory + Claude Code MCP coverage

cjsanjay:
## Summary

Ship every discovered skill, plugin, and MCP server as its own
`ai_component.observed` record on the managed-enterprise pipeline, tagged with
the parent agent connector so AI Defense can correlate MCP servers, skills, and
plugins with the coding agent that hosts them. Extend the Claude Code MCP
reader so `~/.claude.json` (user- and per-project local-scope) is walked in
addition to `~/.claude/settings.json`. Split the delivery-circuit cool-down so
authentication failures no longer suppress a destination for a full 24 hours.

## Motivation

The previous managed-enterprise pipeline shipped only aggregate `mcp_inventory`
and `connector_inventory` envelopes, both v7 gateway-wrapped rollups that
carried a count but no per-item detail. Downstream AI Defense inventory
processing therefore had no way to enumerate the actual MCP servers / skills
/ plugins on the endpoint, nor correlate any of them with their parent agent
connector (codex / claudecode / cursor / ...). The Claude Code CLI stores
user-scope and local-scope MCP servers in `~/.claude.json`, which the sidecar
did not read; every MCP server the operator registered through the Claude
Desktop UI was invisible to inventory.

Separately, `FailureClassAuthentication` shared the `FailureClassUnsafeEndpoint`
24-hour immediate-open trap. A single expired token mint or a transient
identity-service hiccup could silently suppress the managed AID destination
for a full day.

## What changed

**Per-item shipping (`feat(observability)`):**

- Two new release-owned actions reserved in the managed plan --
  `managed_skill_inventory`, `managed_plugin_inventory`.
- New per-connector MCP fanout in `internal/gateway/inventory_events.go` that
  enumerates every configured connector's native MCP config (Codex TOML,
  Claude Code settings + `~/.claude.json`, Cursor JSON, Windsurf, Copilot,
  Gemini CLI, OpenHands YAML, ZeptoClaw, Hermes) and ships one record per
  `(connector, server)` pair. Coverage unions `cfg.ActiveConnectors()` with
  the built-in connector registry so managed installs (which pin a single
  guardrail connector) still scan every registered client.
- Threading of the latest AI-Discovery report snapshot into the endpoint
  inventory emitter so skills, plugins, and MCP evidence rows surfaced by the
  discovery scan flatten into per-entry records in addition to the aggregate
  rollup.
- Every per-item record carries `defenseclaw.agent.discovery.connector`
  (parent-agent slug) and `defenseclaw.inventory.item.name` (actual name).
  Component IDs use a hyphen-only prefix
  (`endpoint-mcp-entry-<sha256>`) so they satisfy the
  `defenseclaw.ai.component.id` family constraint
  (`^[A-Za-z0-9][A-Za-z0-9._:/-]*$`, no underscores).
- Managed destination routing: the `drop-managed-inventory-components` route
  now only covers agent and connector inventories (the two with a v7
  compat envelope). MCP is removed because its v7 wrapper carries only a
  count -- per-item detail must survive on the raw v8 send path.
- `managedaid/compatibility.go` accepts both `"ok"` and `"completed"` for
  `defenseclaw.ai.discovery.result`, and lets recognized v8
  discovery/inventory actions without a v7 wrapper
  (`managed_skill_inventory`, `managed_plugin_inventory`, `ai_discovery`)
  flow through unprojected. Unknown / diagnostic actions remain fail-closed.

**Claude Code MCP coverage (`feat(config)`):**

- `readMCPServersClaudeCode` now walks `~/.claude.json` in addition to
  `~/.claude/settings.json`, covering both the CLI's user-scope registry
  (top-level `mcpServers`) and per-project local-scope entries under
  `projects.<path>.mcpServers`. Skipped when `CLAUDE_CONFIG_DIR` is
  explicitly set.
- `readMCPFromClaudeSettings` now parses `url` and `type` fields so
  HTTP / SSE / streamable-http servers keep their transport identity
  instead of being silently discarded because `command` is empty.
- New exported `ReadMCPFromClaudeJSONProjects` so the per-connector
  inventory fanout can reach the local-scope block directly under a
  managed-install user home without mutating the daemon's `$HOME`.

**Bounded auth circuit (`fix(observability)`):**

- `FailureClassAuthentication` now uses a new
  `authenticationCircuitOpenDuration = 5 * time.Minute` instead of the
  shared 24-hour immediate-open. `FailureClassUnsafeEndpoint` keeps the
  24-hour trap. Both paths go through a shared
  `immediateCircuitOpenDurationFor(class)` helper so `Circuit` and
  `Dispatcher` stay in sync.

**Installer + discovery (already on the branch):**

- `feat(installer): populate ai_discovery.home_dirs from eligible-user
  enumeration` -- root/launchd sidecar now scans every real user home.
- `fix(inventory): enumerate skill/mcp child basenames in discovery signals`
  -- basenames field carries actual skill/plugin/rule names + declared MCP
  server names, not the constant parent-dir / config-file base.

## Test plan

- [x] `go build ./...` clean
- [x] `go test ./internal/observability/...` all green
- [x] `go test ./internal/config/...` all green
- [x] `go test ./internal/inventory/...` all green
- [x] `go test ./internal/gateway/` all green (updated
      `makeEndpointInventoryEmitter` signature callers)
- [x] Live end-to-end verification with a local OTEL POC pointed at
      `http://127.0.0.1:4318` (via a set of do-not-merge loopback bypass
      patches that were kept local and NOT included in this PR):
  - `webex-bridge`, `mcp-atlassian`, `cisco-platform-sdk` (Claude Desktop
    UI) land on the wire as `ai_component.observed` records with
    `defenseclaw.agent.discovery.connector = "claudecode"`.
  - `codex` connector MCP servers (`computer-use`, `node_repl`) land
    with `connector = "codex"`.
  - `.claude/settings.json` servers (`atlassian`, `poc-filesystem`,
    `poc-fs`) land with `connector = "claudecode"`.
  - Per-item skills / plugins ship with the same shape.
- [x] Pre-existing test failures in `internal/gateway/connector/` (cursor
      hook fail-open, ClaudeCode Darwin managed-preference) reproduce on
      `origin/main` and are unrelated to this change.

The changes on this PR were validated against a live sidecar in
`managed_enterprise` mode. A separate contract document
(`AI_DEFENSE_INGEST_CONTRACT.md`) has been shared with the AI Defense parser
team out-of-band so they can update
`ai-firewall-services/event_manager/defenseclaw_events` to accept the v8
per-item shape.


internal/observability/destinations/managedaid/compatibility.go
```diff
@@ -285,6 +285,17 @@ func projectManagedCompatibility(
 			return managedCompatibilityProjection{}, false, false
 		}
 	default:
+		// v8 discovery / inventory records without a v7 legacy wrapper
```
vineethsai7: [P1] This leaves managed inventory in a mixed contract: agent/connector/MCP summaries are still rewritten as schema-v7 gateway envelopes, while the new records pass through as v8. Since the target is v8-only, remove the inventory compatibility projection and deliver every generated v8 summary/item unchanged through managed egress.

internal/inventory/ai_discovery.go
```diff
@@ -1678,14 +1756,95 @@ func (s *ContinuousDiscoveryService) detectPlugins() []AISignal {
...
+	}
+	evidence := []AIEvidence{base}
+	fi, statErr := os.Stat(path)
+	if statErr == nil && fi.IsDir() {
+		if entries, err := os.ReadDir(path); err == nil {
+			for _, entry := range entries {
+				if len(evidence) >= maxEvidencePerSignal {
+					break
+				}
+				name := sanitizeBasenameValue(entry.Name())
```
vineethsai7, line 1926: [P1] A generic direct-child walk treats Codex’s `.system` container as an ordinary skill. Mirror the existing bundled-skill traversal instead: expand eligible `.system/<child>` entries, preserve trusted bundled/origin provenance, and add path-based mutation guards so `.system` and descendants can never be blocked or quarantined.

internal/gateway/inventory_events.go
```diff
@@ -146,15 +150,277 @@ func emitEndpointInventory(
...
+		return nil
+	}
+	var results [][]config.MCPServerEntry
+	tryFile := func(reader func(string) ([]config.MCPServerEntry, error), relPath string) {
+		full := filepath.Join(home, relPath)
+		if entries, err := reader(full); err == nil && len(entries) > 0 {
+			results = append(results, entries)
+		}
+	}
+	switch strings.ToLower(strings.TrimSpace(connectorName)) {
```
vineethsai7, line 385: [P1] This second connector switch is not an all-connector implementation. It omits OpenClaw/Antigravity/OpenCode/Amp and has incorrect or incomplete paths/parsers for Windsurf, Copilot, ZeptoClaw, and Hermes. Refactor the canonical connector readers to accept an explicit home/workspace rather than duplicating the matrix here.

internal/inventory/ai_discovery.go
```diff
@@ -1594,14 +1594,92 @@ func (s *ContinuousDiscoveryService) detectMCPPaths() []AISignal {
...
+// signal so a malformed config never suppresses the "endpoint has
+// MCP configured" signal itself.
+func (s *ContinuousDiscoveryService) signalFromMCPConfigPath(sig AISignature, path string) AISignal {
+	base := AIEvidence{Type: "mcp", Basename: filepath.Base(path), PathHash: hashPath(path)}
+	if s.opts.StoreRawLocalPaths {
+		base.RawPath = path
+	}
+	evidence := []AIEvidence{base}
+	for _, name := range readMCPServerNames(path) {
+		if len(evidence) >= maxEvidencePerSignal {
```
vineethsai7: [P1] Reusing the evidence limit silently caps per-item MCP inventory at 31 servers (the parent row consumes one slot), without any truncation or partial flag. Inventory should page or use a separate bound and must surface incomplete coverage.

internal/inventory/ai_signatures.json
```diff
@@ -282,6 +282,19 @@
         "Cursor"
       ],
       "local_endpoints": [],
+      "skill_paths": [
```
vineethsai7, line 285: [P1] The packaged Python catalog was not updated with this authority change, so the byte-parity CI test fails. Also, the new Cursor plugin paths conflict with the canonical connector implementation, while most supported connectors still lack skill/plugin paths. Synchronize the generated/package copy only after resolving the connector-surface parity.

vineethsai7 (changes requested):
Manual review only — no CodeRabbit used.

I’m requesting changes because the current head does not meet the stated acceptance criteria: v8-only observability end to end, complete connector coverage, and a hard guarantee that built-in skills cannot be blocked or quarantined.

## 1. Managed inventory is still mixed v7/v8

The new records are constructed with the generated v8 builders, but the managed wire path is not v8-only:

- agent, connector, and MCP summaries are projected into legacy `gatewaylog.Event` envelopes using schema version 7;
- skill/plugin summaries and MCP/skill/plugin item records pass through as raw v8;
- the managed route drops raw v8 agent/connector item records to avoid duplicating the legacy projection.

For a v8-only contract, remove the AI-discovery inventory cases from `projectManagedCompatibility`, remove `drop-managed-inventory-components`, and send the original generated v8 summary/item records through routing, redaction, and managed egress. Rewrite the managed adapter tests so they assert the canonical v8 wire body instead of deserializing `gatewaylog.Event`. If “v8 for everything” includes guardrails too, the verdict projection in the same adapter must also be removed.

The existing producer registration is already correct: these records use `gateway_event/ai_discovery`, whose identity set includes `ai.discovery.completed` and `ai_component.observed`. The `managed_*_inventory` strings are routing actions, not producer keys, and should not be added as producer-registry entries.

## 2. Built-in skill protection is not satisfied

The Go directory walker emits `~/.codex/skills/.system` itself as an ordinary `skill_entry`. The established Python discovery behavior instead expands eligible `.system/<child>` skills and marks them `bundled=True`.

The v8 projection drops that provenance completely, and existing Go watcher, REST, and Python mutation paths do not have a hard `.system` exemption. They can accept `.system` or descendants, and the quarantine helper can move the complete container.

Required before managed remediation can rely on this inventory:

- never inventory `.system` as a skill;
- expand eligible bundled children and preserve trusted origin/bundled provenance;
- hard-refuse block/disable/quarantine of `.system` and descendants at every Go and Python mutation boundary;
- treat missing provenance as non-actionable;
- preserve enforceability for a same-named user-installed skill.

## 3. Connector support is incomplete

The explicit-home MCP reader omits OpenClaw, Antigravity, OpenCode, and Amp. It also uses incomplete or incorrect paths/parsers for Windsurf, Copilot, ZeptoClaw, and Hermes. The fallback parser cannot decode OpenClaw, ZeptoClaw, OpenCode, or Amp schemas.

Skill paths currently cover only Codex, Claude Code, Hermes, and Cursor; plugin paths cover only Codex, OpenClaw, Hermes, and Cursor. OpenClaw’s catalog path is incorrect, and the added Cursor plugin paths conflict with the canonical connector implementation. Dynamically loaded connectors are added to the roster but have no inventory-reader interface.

Please make connector-native readers/surface definitions the single source of truth, parameterize them with explicit home/workspace roots, and add a parity matrix covering every registered connector and supported artifact type.

## 4. Inventory can report incomplete or contradictory data as complete

- MCP and directory enumeration silently stop after 31 item rows because the parent consumes one slot of the 32-entry evidence limit.
- Truncation, read, permission, and parse failures do not set a partial/coverage status.
- Per-home snapshots are emitted with `partial=false`.
- Same-named MCP servers across user homes collapse because identity omits home/scope.
- Same-named skills/plugins across homes get indistinguishable component IDs.
- Report-derived and direct MCP records can emit the same ID while disagreeing about `disabled`.

Inventory needs explicit coverage/truncation state and privacy-safe, scope-aware identity semantics.

## 5. The v8 telemetry specification does need updating

`schemas/telemetry/v8/operations.yaml` still documents endpoint inventory as connector/MCP/coding-agent data and has no built-in/bundled/origin field. It also does not define skill/plugin component vocabulary.

Please:

- add a trusted `inventory.item.origin` or bundled field and include it in `body.endpoint.inventory`;
- document/validate `skill` and `plugin` component types;
- update `inventory.item.name` and connector-correlation semantics;
- define the scope-aware identity rule;
- omit `agent.discovery.installed` from MCP/skill/plugin rows because it means the connector itself was observed installed;
- emit the actual discovery privacy mode rather than hardcoding `enhanced`;
- regenerate all telemetry registry/build artifacts.

`make telemetry-check` passing only proves the generated files match the unchanged YAML; it does not establish semantic completeness.

## Validation

Focused Go inventory/gateway/config/managed-destination/router tests passed locally, but they do not cover the connector matrix, truncation, built-in non-enforcement, or v8-only managed wire format.

CI is also currently failing `test_packaged_catalog_is_byte_identical_to_go_authority`: `internal/inventory/ai_signatures.json` changed but `cli/defenseclaw/inventory/ai_signatures.json` was not synchronized.

vineethsai7:
## Follow-up manual deep dive (no CodeRabbit)

Reviewed exact head `6156f3cb3103be117465ae8cea9630545f9b7070`.

I found two additional merge-blocking issues and confirmed the built-in-skill behavior with the installed scanner.

### 1. The v7 compatibility projector can reject real generated v8 guardrail records

This is more than a mixed-contract concern. Production v8 guardrail direction is `input|output|tool` and completed-record producers commonly put that normalized value in both `defenseclaw.guardrail.stage` and `defenseclaw.guardrail.direction`:

- `internal/gateway/inspect_trace_v8.go:227-235`
- `internal/gateway/api_guardrail_event_observability_v8.go:149-162`
- `internal/gateway/api_guardrail_evaluate_observability_v8.go:295-325`
- `internal/gateway/proxy_guardrail_observability_v8.go:831-846`

The managed compatibility projector copies those strings directly into the v7 `gatewaylog.Event` at `internal/observability/destinations/managedaid/compatibility.go:343-382`. The v7 schema only accepts direction `prompt|completion|tool_call` and legacy stage values. Validation therefore fails, and `adapter.go:211-215` returns `OutcomePermanentPayload` for the batch.

The golden test hides this by handcrafting `stage=final` and `direction=prompt` instead of passing a record built by the generated v8 builder (`adapter_test.go:220-255,598-638`).

This reinforces the v8-only requirement: remove the managed compatibility projection for guardrails and inventory, remove the raw-v8 agent/connector drop route, and send the canonical generated v8 records unchanged. Update the hard-coded route invariants in `destinations/factory.go` and `router/router.go`, and replace the golden with generated v8 records for input/output/tool.

### 2. Built-in Codex skills are scanned today, and false-positive enforcement is reachable

The canonical Python discovery correctly expands `~/.codex/skills/.system/<child>`, marks each child `bundled=True`, and gives an ordinary top-level same-name skill precedence. Bulk scan then collapses each row to `(name, path)`, discarding `bundled`, and `--action` applies the same quarantine/disable/block logic as an ordinary skill:

- `cli/defenseclaw/skill_discovery.py:48-90`
- `cli/defenseclaw/cmd_skill.py:2149-2209,2274-2315`
- `cli/defenseclaw/cmd_skill.py:1956-2073`

Using the installed `skill-scanner 2.0.4` with the gateway defaults (`--policy permissive --lenient`) against the current Codex built-ins:

- `plugin-creator`: **CRITICAL**, including `PROMPT_INJECTION_CONCEALMENT` and `YARA_prompt_injection_generic`;
- `skill-installer`: **MEDIUM**;
- `openai-docs`: **MEDIUM**;
- scanning the `.system` container itself exits 1 because it has no `SKILL.md`.

That last case matters because the Go watcher treats a newly created/renamed direct child `.system` as a skill. Scanner error is fail-closed and calls `enforceBlock` unconditionally, which can quarantine the whole container:

- `internal/watcher/watcher.go:455-488,748-815,988-1013`
- default skill `take_action` is true at `internal/config/config.go:3635-3637`.

Periodic rescan skips dot directories, so the same built-ins receive inconsistent treatment depending on how the filesystem event arrives.

### Recommendation: protect after scanning; do not use the existing auto-allowlist

A name-based auto-allowlist is unsafe and also loses the visibility we want:

- existing allow semantics can bypass scanning;
- a user-installed top-level skill deliberately shadows a same-name `.system` skill;
- watcher allow-list input drops stored `source_path`, so it can degrade to a name-only decision.

Instead, add connector-aware local provenance and a separate enforcement mode:

- ownership/origin: `operator|workspace|connector_bundled|defenseclaw_managed|third_party_cache|unknown`;
- verification state;
- enforcement mode: `normal|monitor_only|repair_only`.

For a locally resolved, protected connector-owned built-in: **scan it and emit all findings/errors, but never auto-quarantine, auto-disable, or auto-block it**. Remediation should be a connector update/repair. Re-resolve provenance at every mutation boundary; never trust a telemetry boolean as authorization. A same-name user/workspace skill remains ordinarily enforceable.

Please add hard refusals in the Python and Go quarantine primitives plus CLI/manual, REST, watcher scanner-error, and runtime-disable paths. Do not broadly protect plugin caches or other third-party cache roots.

### 3. The PR's new inventory still cannot carry this safety decision

The Go discovery reports `.system` itself as an ordinary skill and generic enumeration also treats containers/arbitrary direct children as components. The v8 endpoint body has no item origin, scope, verification, protection/enforcement mode, or privacy-safe leaf identity.

Please update the v8 source specs and regenerate artifacts:

- actual skill/plugin/rule component vocabulary;
- item origin/ownership, scope, verification/protection or enforcement eligibility;
- authoritative/partial/truncated coverage;
- parent connector and scope-aware identity;
- active/disabled only when observed;
- an explicit enforcement-suppression reason for protected assets.

Rules are detected by AI discovery but are not included in the managed per-entry fanout. Either add rule inventory or narrow the stated feature scope.

### Required regression coverage

1. Built-in children are scanned and findings are emitted, but critical findings and scanner errors cannot mutate/block them.
2. `.system` is never emitted or quarantined as one asset.
3. A same-name top-level user skill is not exempt.
4. Manual/REST/watcher/low-level block, disable, and quarantine reject protected paths and descendants.
5. Real generated v8 guardrail input/output/tool records survive managed export with no v7 envelope, payload HMAC, legacy gateway attributes, or forced device alias.
6. Every registered connector has table-driven MCP/skill/plugin/rule path and parser parity, including cache/container traversal and user/workspace scopes.
7. More than 31 items marks the snapshot partial/truncated instead of silently reporting complete coverage.

Validation from this pass:

- focused Go inventory/gateway/managed-adapter tests pass, but do not cover the failures above;
- existing Python bundled/system tests pass (6 tests), proving bundled skills are intentionally scanned;
- telemetry generation `--check` passes, which only proves generated artifacts match the currently stale semantic spec;
- packaged signature catalog parity still fails because the Go and Python copies differ.

vineethsai:
Did you also inlcude windows paths for skills, MCPs and plugins as well?

cjsanjay:
> Did you also inlcude windows paths for skills, MCPs and plugins as well?

No Windows changes will be a follow-up PR.

internal/observability/destinations/managedaid/compatibility.go
cjsanjay: Addressed in commit `5f567946` (`refactor(managed): v8-only inventory route, drop legacy v7 projection`) + test-update commit `b8c1fdad`.

**Code changes:**
- `internal/observability/destinations/managedaid/compatibility.go` — `projectManagedCompatibility` switch simplified to two branches: `guardrail.evaluation.completed` (v7 projection retained) + `default` (v8 passthrough via `isV8PassThroughAction`). The three AI-discovery inventory cases are gone.
- Same file — `isV8PassThroughAction` widened to include `managed_agent_inventory`, `managed_connector_inventory`, `managed_mcp_inventory`, `managed_skill_inventory`, `managed_plugin_inventory`, `ai_discovery`. Diagnostic actions intentionally excluded (fail-closed).
- `internal/config/observability_v8_managed_destination.go` — `drop-managed-inventory-components` route removed; the destination now carries two generated routes (`drop-local-inventory-diagnostics` + `all-collected-logs`).
- `internal/observability/destinations/factory.go` + `internal/observability/router/router.go` — validators updated in lockstep to `len(routes) == 2`; sensitive-profile check retargeted to `Routes[1]`.

**Test rewrites (asserting canonical v8 wire body per your ask):**
- `TestAdapterManagedCompatibilityGoldenWire` — inventory records now decoded as canonical v8 (bucket/action/body/correlation/provenance), verdict retains v7 flat-attribute shape. Golden payloads represent already-redacted state (raw_path/command_path/url-with-token/authorization stripped) since redaction runs before the adapter.
- `TestManagedCompatibilityInventoryIsV8Passthrough` (renamed) — asserts every managed inventory action × body shape returns `useProjection=false, valid=true`; diagnostic action still fail-closed.
- `TestAdapterV8InventoryPassthroughReachesNetwork` (renamed) — asserts mismatched-parallel-array carrier reaches network + wire body contains `ai.discovery` + `managed_connector_inventory` identity.

**Producer registration** untouched — records still register under `gateway_event/ai_discovery`. Confirmed the `managed_*_inventory` strings are `ProducerKey` constants used as routing actions, not producer-registry entries.

**Guardrail verdict projection** — intentionally NOT removed in this PR (the "if v8 for everything includes guardrails too" clause). It's a wire-shape break for downstream verdict consumers that warrants its own coordinated PR + approval trail. Called out in the commit's deferrals section. Happy to file a follow-up ticket if you'd like.

internal/inventory/ai_discovery.go
cjsanjay: Addressed in commit `7969ed75` (`fix(inventory): protect .system bundled skills at every mutation boundary`).

**Walker changes** (`internal/inventory/ai_discovery.go`):
- `signalFromDirectoryChildren` for `detector=="skill"` never emits `.system` as a `skill_entry`. Recurses one level via new `appendBundledSkillChildren` helper — each `.system/<child>` emits with `Origin="bundled"`, `Bundled=true`. Sibling user skills continue to emit with `Origin="user"`. Non-skill detectors (plugin/rule) treat `.system` literally — the reserved-container contract is skill-specific.
- `AIEvidence` extended with `Origin` (`"user"` | `"bundled"`) + `Bundled bool` fields (kept in sync). Missing `Origin` is treated as non-actionable by all mutation paths — the fail-safe rule you asked for.

**Hard mutation guards at every Go + Python file-mutation boundary:**
- New `internal/enforce/bundled_skill.go` — `IsBundledSkillPath` (component-wise `.system` check on cleaned path, guards against substring false-positives like `hello.system.md`) + `IsBundledSkillContainerName` + `ErrBundledSkill` sentinel.
- `internal/enforce/plugin_enforcer.go` — `Quarantine` refuses `ErrBundledSkill` before any stat / copy touches the source tree. No partial move can land.
- `internal/enforce/quarantine_asset.go` — `NewAssetQuarantinePlan` (skill target) refuses `ErrBundledSkill` after `pathWithinRoots` normalises the source path.
- `cli/defenseclaw/enforce/skill_enforcer.py` — Python `SkillEnforcer.quarantine` resolves symlinks then raises `BundledSkillRefusedError`. Check runs before any hash/copy so bundled content is never staged.

**Same-named user skill parity** preserved — sibling `hello` next to `.system/hello` continues to emit distinctly and remains enforceable (verified in `TestSignalFromDirectoryChildrenExpandsBundledSkillContainer`).

**11 test cases** added across `internal/enforce/bundled_skill_test.go` + `internal/inventory/bundled_skill_walker_test.go`: table-driven path check (positive + substring-negative), refusal + source-tree preservation, quarantine-store no partial copy, walker recursion parity, non-skill detector isolation, symlink guard.

**Deferred** — `PolicyEngine.Block/Quarantine/Disable` guard by source-path lookup needs the audit store's `GetSourcePath` surface which doesn't exist yet; the file-mutation boundary is already guarded. Called out in the commit message.

internal/gateway/inventory_events.go
cjsanjay: Addressed in commit `08808403` (`fix(inventory): connector-surface parity + Python catalog sync`).

**Direct-home MCP switch (`readMCPServersUnderHome` in `internal/gateway/inventory_events.go`):**
- Added: `openclaw`, `antigravity`, `opencode`, `amp`.
- Fixed parsers: `hermes` now uses `ReadMCPFromYAMLPath` with the native `mcp.servers` key (was silently returning nothing on YAML); `amp` uses `ReadMCPFromClaudeSettings` (matches the top-level `mcpServers` shape); `zeptoclaw` reads both `.zeptoclaw/config.json` (native) and legacy `.zeptoclaw/mcp.json`.
- `hasNativeMCPReader`: duplicate `openclaw` removed; every supported connector now returns true.

**Catalog fixes (`internal/inventory/ai_signatures.json`):**
- `openclaw`: `plugin_paths` corrected — was `~/.openclaw/plugins`, now `~/.openclaw/extensions` (canonical location per `connector/openclaw.go` — the fetch-interceptor plugin lives at `~/.openclaw/extensions/defenseclaw/`).
- `cursor`: speculative `plugin_paths` **removed** — resolves the conflict you flagged. Cursor doesn't have a user-plugin surface; VS Code extensions are tracked separately via `editor_extension`.
- `skill_paths` added for `windsurf`, `geminicli`, `copilot`, `antigravity`, `openhands` (previously coverage was Codex / Claude Code / Hermes / Cursor only).
- `plugin_paths` added for `antigravity`, `opencode`, `amp` (previously coverage was Codex / OpenClaw / Hermes / Cursor).

**Parity matrix test** — new file `internal/inventory/connector_parity_test.go`:
- `TestSupportedConnectorParityMatrix` — every `supported_connector` must declare at least one of `mcp_paths / skill_paths / plugin_paths`. Failures name the exact connector/artifact gap so a regression points at one row. `omnigent` is exempt (custom policy bridge, no user-facing surfaces) with a documented reason inline.
- `TestSupportedConnectorMCPOnEveryConnector` — `mcp_paths` required on every supported_connector except `mcpExempt` (`omnigent`).
- `TestSupportedConnectorCountMatchesRegistry` — pins the count at 14 so silent additions trip the test.

**Deferred** — the full `InventoryReader` interface refactor deleting `readMCPServersUnderHome` entirely (and the parser dispatch) is a larger cross-package refactor with its own review surface; the catalog is now the single source of truth for path declarations. The `disabled` normaliser between report-derived and direct MCP readers belongs with that refactor. Both called out in the commit's deferrals.

internal/inventory/ai_discovery.go
cjsanjay: Addressed in commit `15ed8243` (`feat(inventory): coverage/truncation flags on AISignal + MCP cap fix`).

**MCP cap fix (`internal/inventory/ai_discovery.go` — `signalFromMCPConfigPath`):**
- New explicit constant `maxMCPServerRowsPerSignal = maxEvidencePerSignal - 1` so the parent evidence row's slot reservation is grep-able intent.
- Cap-hit now stamps `signal.Partial = true`, `signal.CoverageReason = CoverageReasonCapExceeded`.
- New `readMCPServerNamesWithErr` shim so the parse-error branch propagates too — malformed MCP config emits `Partial=true, CoverageReason=parse_error` with the parent `"mcp"` evidence row still present (operator sees "endpoint has MCP configured" even when the parser fails).

**Generalized coverage semantics** — same commit adds two additive JSON fields to `AISignal` (both `omitempty`, no wire break):
- `Partial bool` — snapshot is not complete; consumers that promise authoritative MUST fail-close on `true`.
- `CoverageReason string` — enumerated over `cap_exceeded / read_error / permission_denied / parse_error`. Constants exported.

`signalFromDirectoryChildren` (skill/plugin/rule walker) sets both on stat/readdir permission-denied, generic read error, and per-signal cap hit. `appendBundledSkillChildren` (from the `.system` protection commit) returns `(partial, reason)` so bundled read errors propagate up to the outer signal.

**Wire vocabulary** — new attributes documented in `schemas/telemetry/v8/operations.yaml` (commit `b283da90`):
- `defenseclaw.inventory.partial` (boolean).
- `defenseclaw.inventory.coverage_reason` (enum with the four values above).

**6 test cases** across `internal/inventory/coverage_test.go` + `mcp_cap_test.go`:
- MCP cap exceeded on 36-server config → `Partial=true, reason=cap_exceeded, mcp_server rows == maxEvidencePerSignal-1` (parent reserved).
- Malformed JSON → `Partial=true, reason=parse_error`, parent row preserved, zero `mcp_server` rows.
- Healthy 3-server config → `Partial=false, CoverageReason=""`.
- Directory walker cap-exceeded / permission-denied / healthy cases.

**Paging** — chose truncate-and-flag over per-connector paging for v1 (documented in `PR674_VINEET_REVIEW_PLAN.md` open question #4). Happy to revisit if you'd rather have paging for MCP specifically.

internal/inventory/ai_signatures.json
cjsanjay: Both issues in this thread addressed in commit `08808403` (`fix(inventory): connector-surface parity + Python catalog sync`).

**Cursor plugin path conflict:**
- The speculative `plugin_paths` I'd added on Cursor is **removed**. Cursor doesn't expose a user-plugin surface — `~/.cursor/plugins/local/` is Cursor's internal state, not a user-installed plugin dir. VS Code extensions (the closest real analog) are already tracked separately via `editor_extension`. Cursor's signature now carries `skill_paths` + `mcp_paths` + `rule_paths` only, matching the canonical connector implementation.

**Missing skill/plugin paths on the other supported connectors** — filled the catalog gaps:
- `skill_paths` added: `windsurf`, `geminicli`, `copilot`, `antigravity`, `openhands`.
- `plugin_paths` added: `antigravity`, `opencode`, `amp`.
- Plus the OpenClaw path fix: `plugin_paths: ~/.openclaw/plugins` → `~/.openclaw/extensions` (canonical location per `connector/openclaw.go`).

Coverage matrix now enforced by `TestSupportedConnectorParityMatrix` (new file `internal/inventory/connector_parity_test.go`) — every `supported_connector` must declare at least one of `mcp_paths / skill_paths / plugin_paths`; the failure message names the exact connector/artifact gap.

**Python catalog byte-parity:**
- `cli/defenseclaw/inventory/ai_signatures.json` synced byte-for-byte from the Go authority in the same commit. `test_packaged_catalog_is_byte_identical_to_go_authority` passes locally; all 8 Python tests in `tests/test_ai_signatures.py` green. CI is now green on this file — was previously failing per your inline flag.

Order-of-operations note: the sync landed **after** the connector-surface parity fixes were applied to the Go authority in the same commit, so nothing had to be re-synced in a follow-up.

PR #677, opened 2026-08-06
cherry-pick: port release-26.7.3 installer fixes (#676 + #685) to main

cjsanjay:
Companion to #676 (release-defenseclaw-enterprise-26.7.3). Ports the customer-facing half of that fix to `main`.

## Summary
- `render_targets_manifest` now skips a target row when `discover_agent_version` returns empty for that user's home. Prior behavior emitted `agent_version: ""` and the guardian churned every tick against a CLI that wasn't installed — filling `hook_guardian_state.json` with permanent "agent_version empty" failures for that (user, connector). Late installs are still picked up by the next `hook-enumerator` re-render.
- Updated `install.sh` zero-target `die()` to name the new failure mode ("none of the requested connectors are installed for any eligible user") in addition to the existing "every connector unsupported" wording.

## Why this is a partial port of #676
The release-26.7.3 branch introduced a multi-install-picker refactor in `discover_agent_version` (`_pick_highest_supported "${MIN_..._VERSION}" "${versions[@]}"`). Under macOS bash 3.2 + `set -u` that array expansion is an "unbound variable" error when the array is empty — which killed the installer mid-render on hosts missing one connector. `main` uses the simpler first-hit-wins loop with no `versions` array and no picker, so the crash side of #676 **does not apply** here — only the "don't emit a row for something we can't find" policy needs to come across. `test_agent_version_multi_install.sh` from the release-branch PR is likewise skipped: the file does not exist on `main` because the picker it tests doesn't exist on `main`.

## Test plan
- [x] `packaging/macos/tests/run_tests.sh` — 148 pass under real bash 3.2.57 on macOS
- [x] `absent connectors are skipped (partial-box render)` renders a manifest with only cursor's row when codex/claudecode aren't installed
- [x] `all connectors absent yields zero rows` confirms schema-valid empty manifest so the installer's `MANIFEST_TARGETS == 0` preflight fires the operator-visible die() (not silent success)
- [x] Prior tests made hermetic against real Codex/Claude/Cursor installs on the dev box via a `_reset_discover_stub` helper (previously the fake `/Users/alice`, `/Users/bob` homes only happened to produce rows because the old code emitted them with empty `agent_version`; the new policy would have dropped all rows on any CI runner)

## Related
- Release branch PR: #676


cjsanjay:
Added a second cherry-pick to this PR: **#685 — remove `/usr/bin/python3` dependency from macOS install/uninstall lifecycle**.

Rationale: QA regression on stock macOS hosts without Xcode Command Line Tools. `/usr/bin/python3` is a stub there that passes `[ -x ]` but fails to launch, crashing every install/uninstall step that shelled out to it. PR #685 replaced each Python here-doc with pure awk/bash/Go equivalents and landed on the release branch.

**Commit:** [`2ed3dd0b`](../commit/2ed3dd0b)

**Main-branch reconciliations applied during the port:**

- **Hybrid scrubber for the `amp` connector.** Main added a signed-backup restore mechanism for Amp (PR #640) that the Go scrubber this port introduces doesn't cover. Rather than regress amp uninstall or block on a lengthy Go port of SHA-256 + base64 + JSON manifest crypto, `uninstall.sh` split-routes: `codex/claudecode/cursor` → Go binary (python3-free per the QA fix); `amp` → `scrub_agent_configs.py` (Python retained just for this one connector). This keeps main's amp functionality intact while closing the QA regression for the three connectors that account for essentially all customer installs.

- **`_read_codex_version_as_user` rewrite.** Main added a new python3-based 5-second-timeout probe (`sudo -n -u USER codex --version`) after the release branch was cut. That helper is on the install-time path (`discover_agent_version`'s codex branch), so leaving it python3-dependent would leak the same QA regression this port closes. Rewritten in pure bash with a background-PID + sleep-loop + KILL as the 5-second budget.

- **`_read_json_version` extension.** Main added an optional `expected_name` argument that verifies `package.json .name` before returning `.version` (used by the `@ampcode/cli` identity check). Preserved as a two-field awk pass — read name, compare, then read version — so the caller-facing contract survives the pure-awk rewrite.

- **Root PreRunE refactor.** Main refactored the inline `PersistentPreRunE` into a top-level `rootPersistentPreRunE`. Re-applied #685's `defenseclaw.skip-daemon-bootstrap` annotation carve-out so `enterprise hooks scrub` doesn't try to load config.yaml on partial-install hosts.

- **Managed env-key set.** Main's Claude Code connector picked up seven new OTel keys (`OTEL_TRACES_EXPORTER` + per-signal `metrics/logs/traces` variants). Added to `claudeManagedEnvKeys` so the scrubber strips them on uninstall — the sync-guard test enforces this both directions.

- **`TestMain` composition.** Main added a `TestMain` in `connector_cmd_test.go` for the Codex app-server fixture. Folded #685's scrub-helper exit path into that same `TestMain` (Go permits one `TestMain` per package).

- **Dep bumps already on main.** `pyproject.toml` (aiohttp / cryptography / rich), `uv.lock`, and `extensions/defenseclaw/package-lock.json` (js-yaml) already have the versions #685 bumped to. Main's values preserved where they were newer.

**Test tally on this branch:**
- `go build ./...` clean.
- `go test ./internal/cli/ -run 'TestScrub|TestExitCodeFor'`: green (22 scrub + exit-code tests including sync-guard reverse coverage).
- Shell test suite: 146/148. The two failing shell tests fail identically on `origin/main` HEAD (macOS symlink-detection around `/var` → `/private/var`) — pre-existing, unrelated to this cherry-pick.
- `pip-audit` + `npm audit --omit=dev`: clean.

cjsanjay:
@coderabbitai review

Round-2 findings addressed in commit \`02f6c7c5\`:

- **rc 5** for missing \`--file\` (was rc 3, which is documented as unsupported-connector). Distinct code + updated help text.
- **Stale registry consumer line** dropped: registry entry now points at bare \`packaging/macos/uninstall.sh\` rather than a line number that drifted across commits.
- **\`_read_codex_version_as_user\`** fails closed on mktemp failure (dropped the predictable \`/tmp/…\$\$\` fallback; a colocated user could have pre-symlinked it under a root-run install).
- **\`LC_ALL=C\`** pinned on the JSON-parser awk invocation so \`sprintf("%c", cp)\` and \`substr\` operate on raw octets. Added \`t_read_json_field_decodes_unicode_escape_bytes\` — exercises ASCII / BMP / surrogate-pair \\uXXXX round-trips under both \`en_US.UTF-8\` and \`C\`.
- **Bundle README** now qualifies the "no Python needed" claim: Python 3 is only required when \`--purge\` runs against a host with the \`amp\` connector installed (its scrubber uses stdlib crypto for SHA-256 verification).

Local checks green: \`gofmt\`, \`go build ./...\`, scrub Go tests, envvar + docs-CLI sync guards. Please re-review.

cjsanjay:
@coderabbitai review

Two new commits address the previously-open review threads:
- `4b7b06cd` addresses the RC 5 for empty --connector thread on `internal/cli/enterprise_hooks_scrub.go:118`
- `127e36a5` addresses the Heavy-lift discovery-error propagation thread on `packaging/macos/lib/installer_lib.sh:101`

Both include tests. Please re-review the fresh commits.

cjsanjay:
@coderabbitai review

Merged latest `origin/main` into the branch to bring in 20+ commits landed since my last push. Only one file conflicted (`bundles/llm/model_catalog.json`, Groq model list) — resolved by taking main's superset. Also dropped `llama-3.3-70b-versatile` and `llama-3.1-8b-instant` which passed their LiteLLM deprecation date yesterday.

All 92 CI checks are green post-merge. Requesting a fresh review since the earlier APPROVED signal was invalidated by the merge commit.

cjsanjay:
@coderabbitai review

New commit `1dc69444` addresses all eight open findings from the previous review pass:

**Major**
- `enterprise_hooks_scrub.go:292` — dropped redundant `hooks != nil` guards (S1031 lint)
- `enterprise_hooks_scrub.go:773` — `scrubCodexFile` now handles nested `[[hooks.<event>]]` / `[[hooks.<event>.hooks]]` shape (`tomlHooksSectionRE` + `stayInHooksHierarchy`). Two new tests pin strip + preserve behaviour.
- `enterprise_hooks_scrub.go:411` — root uid boundary check in `writeConfigAtomic`: when euid=0, reject symlink hops that cross the initial owner's uid.

**Minor / Trivial**
- `enterprise_hooks_scrub.go:481` — `tmp.Chown(uid,gid)` on the open fd (was path-based `os.Chown`)
- `enterprise_hooks_scrub.go:535` — corrected the ensure_ascii comment
- `enterprise_hooks_scrub_test.go` — `setScrubFlags` now saves all five package flag vars; new `TestScrubEmitsJSONSummaryOnSuccess` pins the `{ok,connector,file,changed}` `--json` contract
- `installer_lib.sh` — `move_legacy_aside` reads positional args defensively under `set -u`
- `uninstall.sh` — SCRUB_PY (amp branch) validated via `_scrub_bin_trusted`

All 92 CI checks are green on the merged branch. Please re-review.

cjsanjay:
@coderabbitai full review

Commit `0aa93b6a` closes both findings from your 10:50 comment:

1. **Mixed nested Codex hooks** — `scrubCodexFile` now derives the event name from each `[[hooks.<event>]]` entry point (`tomlHooksEventRE`) and builds a per-event predicate that stops at the sibling event boundary. New `TestScrubCodex_MixedNestedHooksKeepsUserEventLosesDCEvent` pins the fix.

2. **marshalSortedIndent comment** — corrected. Go's `SetEscapeHTML(false)` emits raw UTF-8; Python default `ensure_ascii=True` emits escapes.

All 92 CI checks green. Zero open review threads. Please re-review to clear the CHANGES_REQUESTED gate.

cjsanjay:
@coderabbitai full review

Nudging again — my earlier `full review` comment (17:27 UTC yesterday) landed during the GitHub incident and I don't see a fresh review from you after commit `0aa93b6a`.

Current state:
- HEAD is at `0aa93b6a`
- All 92 CI checks are green
- Zero open review threads (confirmed via GraphQL `reviewThreads` query)
- Both findings from your 10:50 UTC comment (mixed nested Codex hooks + `marshalSortedIndent` ensure_ascii comment) are addressed in `0aa93b6a` with a new regression test (`TestScrubCodex_MixedNestedHooksKeepsUserEventLosesDCEvent`)

The `CHANGES_REQUESTED` gate from your 09:14 review is blocking merge. Please re-scan and clear it if you have no new findings.

cjsanjay:
@coderabbitai full review

Your 04:34 UTC comment flagged one remaining issue on `packaging/macos/lib/installer_lib.sh`:

> Line 459 accepts a comma without requiring another object member. Lines 512-524 accept any non-empty unquoted token as a scalar. As a result, invalid documents such as these can return `1.2.3` with exit code `0`:
>
>     {"version":"1.2.3","other":wat}
>     {"version":"1.2.3",}

This is fixed in commit **`0d25131c`** (pushed *after* your 04:34 comment):

- The awk state machine now tracks `after_member` and `saw_comma`. A comma before any member, a double comma, or a comma immediately before `}` all exit 2 (trailing-comma detection).
- The non-target scalar-skip loop now validates the captured run against the JSON scalar grammar: exactly `true` / `false` / `null`, or an RFC 8259 number (`-?(0|[1-9][0-9]*)(\.[0-9]+)?([eE][+-]?[0-9]+)?`). Anything else exits 2.

Verified locally with the exact payloads from your comment:

    {"version":"1.2.3","other":wat}     → rc 2, empty stdout ✅
    {"version":"1.2.3",}                  → rc 2, empty stdout ✅
    {"version":"1.2.3","count":42,"active":true}  → rc 0, "1.2.3" ✅ (over-strict regression guard)

Five new tests in `packaging/macos/tests/test_discover_agent_version.sh`:
- `_read_json_field rc 2 on garbage non-target scalar`
- `_read_json_field rc 2 on trailing comma`
- `_read_json_field rc 2 on leading/double comma`
- `_read_json_field rc 0 on mixed valid scalars` (over-strict guard)
- `_probe_json_version records garbage/trailing-comma` (end-to-end into DC_DISCOVERY_ERRORS_LOG)

All 24 discover-agent-version tests pass on bash 3.2.57. Zero open review threads on GraphQL `reviewThreads`. Please re-scan HEAD (`0d25131c`) and clear the CHANGES_REQUESTED gate.

sanjayab23:
Addressed the two remaining CodeRabbit findings (no prior `Addressed` marker) in `65fe202d`:

- **[ID 3802850705](https://github.com/cisco-ai-defense/defenseclaw/pull/677#discussion_r3802850705) — `internal/cli/enterprise_hooks_scrub.go`**: `newStayInHooksEvent` now accepts only nested continuations (`[hooks.<event>.<sub>]` / `[[hooks.<event>.<sub>]]`) and stops at a repeated root `[[hooks.<event>]]` header. Prevents a DC-owned array element from scrubbing an adjacent user-owned element of the same event. Regression: `TestScrubCodex_RepeatedRootEventKeepsUserElement`.

- **[ID 3802850710](https://github.com/cisco-ai-defense/defenseclaw/pull/677#discussion_r3802850710) — `packaging/macos/lib/installer_lib.sh`**: `_read_json_field` awk parser now rejects two adjacent members without a comma between them (`{"a":"1" "b":"2"}`) with rc 2, keeping `_probe_json_version`'s `malformed-json` discovery-error path engaged. Regressions: `t_read_json_field_rc_2_on_missing_comma_between_members`, `t_probe_json_version_records_missing_comma`.

Branch is also now up to date with `main` (merged `1a1869dd` — provider frequency/presence penalty fix — via `d16745d7`).

Test results on this HEAD:
- `packaging/macos/tests/run_tests.sh`: 166 passed / 2 failed (the 2 failures are pre-existing on `main`, unrelated to this PR).
- `go test ./internal/cli/... -run TestScrubCodex`: all pass.

sanjayab23:
@coderabbitai review

PR #697, opened 2026-08-11
fix(installer): authenticate sandbox helper before execution

### LINKED ISSUE #408: security: verify fallback sandbox installer before execution
## Summary

PR #400 pinned the fallback sandbox installer URL to the selected DefenseClaw release, which avoids tracking `main`, but the fallback path still executes the downloaded script without an integrity check when `scripts/install-openshell-sandbox.sh` is not present next to `scripts/install.sh`.

## Severity

High: installer trust boundary. A network or upstream compromise on the fallback script path can execute code during install when the adjacent sandbox installer file is absent.

## Affected files

- `scripts/install.sh` around the sandbox installer fallback

## Reproduction / context

- Related merged PR: #400
- Merged commit: `9e713cccf04dc03a7e4e349f669c3a042ac36d35`
- Existing PR review context: CodeRabbit requested a checksum/signature-backed download before invoking the fallback script.
- No duplicate open or closed issue was found for `install-openshell-sandbox` checksum/signature tracking.

## Recommended fix

Publish `install-openshell-sandbox.sh` as a release artifact with a checksum or signature, update `scripts/install.sh` to download to a temporary file, verify the artifact, then invoke the verified local copy. Add static installer tests that assert the fallback command verifies the selected release artifact before execution.

vineethsai7:
## Problem

When `install.sh --sandbox` could not find an adjacent `install-openshell-sandbox.sh`, it streamed a pinned Git tag from `raw.githubusercontent.com` directly into Bash. A tag pin provides version selection, but it does not authenticate the bytes before execution. A network, release-account, or mutable-reference compromise could therefore cross the installer trust boundary.

Fixes #408.

## Current Situation

The primary installer already authenticates modern release policy through the Sigstore-signed `checksums.txt` bundle, but the sandbox helper bypassed that boundary:

```mermaid
flowchart LR
    A["Pinned raw GitHub URL"] --> B["curl response"]
    B --> C["bash (executed before authentication)"]
```

The helper also was not part of the release candidate's reviewed immutable installer set, so it had no signed release-asset digest to verify.

### Publish the helper as a reviewed release asset

Starting with release 0.8.11, the release candidate includes `install-openshell-sandbox.sh` in the immutable payload set. Candidate assembly validates its bounded regular-file snapshot, executable mode, completeness marker, exact bytes, and LF representation. The existing release pipeline then includes its SHA-256 digest in the Sigstore-signed checksum bundle.

Historical releases remain historically exact: 0.8.8–0.8.10 still require only the installer assets they actually published.

### Authenticate before execution

The primary installer now downloads/copies the helper into its private release-policy staging directory, verifies it against the already authenticated checksum bundle, protects its mode, re-hashes the bytes immediately before execution, and only then invokes Bash.

```mermaid
flowchart LR
    A["Release workflow OIDC identity"] --> B["Sigstore-signed checksums.txt"]
    B --> C["install-openshell-sandbox.sh SHA-256"]
    D["Immutable release asset"] --> E["Private policy staging"]
    C --> F["Checksum verification"]
    E --> F
    F --> G["Immediate pre-exec re-hash"]
    G --> H["bash authenticated local file"]
```

Missing, tampered, historically unavailable, or pre-execution-mutated helpers fail closed without executing sandbox-installer bytes. Unsupported sandbox releases are rejected immediately after authenticated policy loading, before gateway, CLI, plugin, OpenClaw, or connector installation can mutate the host.

### Prove both security and compatibility

A shell-level regression harness exercises the real download/copy, checksum, digest-binding, and execution functions. It proves a mismatched artifact never executes and a matching artifact still does.

## What Changed (Where & How)

| Path | Change |
| --- | --- |
| `scripts/install.sh` | Replaced raw `curl | bash` and unverified adjacent execution with signed-release download, checksum verification, private staging, and pre-exec digest binding. |
| `scripts/install-openshell-sandbox.sh` | Added the reviewed completeness marker used by candidate validation. |
| `scripts/release_candidate.py` | Added the helper to the immutable installer asset set from 0.8.11 onward without rewriting historical asset contracts. |
| `.gitattributes` | Made helper bytes stable across release runners with LF checkout. |
| `cli/tests/test_install_script_static.py` | Added ordering assertions plus tampered/authenticated execution controls through the real shell functions. |
| `cli/tests/test_release_candidate.py` | Added version-gated asset, staging, checksum-set, and cross-platform byte-contract coverage. |

## Breaking Changes

A newer `install.sh` invoked with `--sandbox` while explicitly selecting DefenseClaw 0.8.10 or older now fails closed because those immutable releases did not publish an authenticated sandbox-helper asset. Normal installation without `--sandbox` is unchanged. Release 0.8.11 and newer publish the authenticated helper.

## Open Issues / Follow-ups

None.

## Test Plan

- `PYTHONPATH=. uv run --frozen pytest -q cli/tests/test_install_script_static.py cli/tests/test_release_candidate.py cli/tests/test_release_workflow_staged.py`
  - **265 passed**
- `PYTHONPATH=. uv run --frozen pytest -q cli/tests/test_install_docs_static.py cli/tests/test_release_channel.py`
  - **160 passed**
- `uv run --frozen ruff check scripts/release_candidate.py cli/tests/test_install_script_static.py cli/tests/test_release_candidate.py`
  - **Passed**
- `bash -n scripts/install.sh && bash -n scripts/install-openshell-sandbox.sh`
  - **Passed**
- `shellcheck -e SC2012,SC2034,SC2059,SC2086,SC2155 scripts/install.sh scripts/install-openshell-sandbox.sh`
  - **Passed** (the exclusions are the repository's existing diagnostics outside this change)
- `git diff --check`
  - **Passed**

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -4,6 +4,7 @@ scripts/upgrade.sh text eol=lf
 scripts/upgrade.ps1 text eol=lf
 scripts/defenseclaw-rescue.sh text eol=lf
 scripts/defenseclaw-rescue.ps1 text eol=lf
+scripts/install-openshell-sandbox.sh text eol=lf
 scripts/install.sh text eol=lf
 scripts/install.ps1 text eol=lf
 cli/defenseclaw/install_publish.py text eol=lf
--- a/cli/tests/test_install_script_static.py
+++ b/cli/tests/test_install_script_static.py
@@ -133,6 +133,19 @@ def test_sandbox_installer_is_authenticated_before_execution() -> None:
     assert download < authenticate < bind_digest < recheck < execute
 
 
+def test_unsupported_sandbox_release_fails_before_installation_work() -> None:
+    text = INSTALL_SH.read_text(encoding="utf-8")
+
+    release_policy = text.index("load_release_policy\n", text.index("main()"))
+    early_version_gate = text.index(
+        'version_gte "${RELEASE_VERSION}" "${SANDBOX_INSTALLER_ASSET_START_VERSION}"',
+        release_policy,
+    )
+    first_install = text.index("install_gateway\n", release_policy)
+
+    assert release_policy < early_version_gate < first_install
+
+
 def test_sandbox_installer_rejects_tampered_bytes_and_executes_authenticated_bytes(
     tmp_path: Path,
 ) -> None:
--- a/cli/tests/test_install_script_static.py
+++ b/cli/tests/test_install_script_static.py
@@ -13,8 +13,12 @@ from __future__ import annotations
 import hashlib
 import os
 import re
+import shlex
+import shutil
 import subprocess
 from pathlib import Path
+
+import pytest
 
 try:
     import tomllib
@@ -133,6 +137,9 @@ def test_sandbox_installer_rejects_tampered_bytes_and_executes_authenticated_byt
     tmp_path: Path,
 ) -> None:
     text = INSTALL_SH.read_text(encoding="utf-8")
+    bash = shutil.which("bash")
+    if bash is None:
+        pytest.skip("Bash is unavailable on this platform")
 
     def shell_function(name: str) -> str:
         match = re.search(rf"(?ms)^{re.escape(name)}\(\) \{{\n.*?^\}}\n", text)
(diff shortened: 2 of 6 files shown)


vineethsai7:
Addressed the late-gate review in `d05b0ff20`: unsupported sandbox releases now fail immediately after authenticated release-policy loading, before gateway/CLI/plugin/OpenClaw installation can mutate the host. A static ordering regression pins the gate before the first install operation; focused file is 10/10 green.

@coderabbitai review

Issue #642, opened 2026-07-30
Bind Doctor authenticated probes to the verified gateway peer


vineethsai7:
## Problem

Doctor verifies the managed gateway PID, executable, start identity, data home, and listener ownership before sending its bearer token. The HTTP probe then opens a separate TCP connection. A local endpoint replacement between listener inspection and connect can therefore receive the credential before the authenticated runtime metadata check rejects it.

The current safeguards remain valuable: proxy bypass, redirects disabled, exact listener ownership, strong PID generation identity, and authenticated PID/data-home metadata. They do not bind the connected peer itself to the inspected process generation.

## Desired outcome

- Bind credential-bearing local requests to the verified gateway peer on macOS, Linux, and Windows.
- Prefer a kernel-backed peer identity, a scoped challenge bound to the inspected process generation, or a transport that cannot be replaced between inspection and authentication.
- Do not expose the long-lived gateway bearer to an unbound replacement listener.
- Preserve support for the configured local TCP API or provide a compatible migration path.
- Add deterministic replacement-race regressions for all supported platforms.

## Context

Tracked as a structural hardening follow-up to the Doctor auto-repair overhaul. The current change fails closed after the response when authenticated runtime PID or data home differs, but it cannot eliminate the inspect-to-connect interval without a gateway protocol or transport change.

PR #701, opened 2026-08-11
fix(setup): bind action-mode restart readiness to one generation

### LINKED ISSUE #681: Switching claude-code connector to --mode action races the gateway restart, causing a fail-closed hook error on the very next tool call
## Summary

Running `defenseclaw setup claude-code --mode action --yes` restarts the gateway as part of the same command. The connector's own PostToolUse hook fired immediately after (from Claude Code, on the CLI invocation that ran the setup command itself) and failed with:

```
defenseclaw: claude-code hook error: invalid or missing action in gateway response
```

Because the connector's default `fail-mode` is `closed` (confirmed via `defenseclaw doctor` → `Mode [claudecode] - observe; fail-mode=closed; provenance=claude-env`), this blocked Claude Code's tool-call continuation immediately after enabling enforcement — i.e., turning on enforcement can transiently lock out the very session that turned it on.

## Steps to reproduce

1. Install DefenseClaw 0.8.10 on Windows, connector = Claude Code, mode = observe (default quickstart flow).
2. From inside a running Claude Code session, run:
   ```
   defenseclaw setup claude-code --mode action --yes
   ```
3. Observe the setup command's own PostToolUse hook (fired by Claude Code right after the command completes) return the error above and stop continuation.

## Gateway log excerpt

```
[guardrail] direct-upstream mode: claudecode policy_mode=action enforcement=true — proxy port intentionally not bound
[sidecar] context cancelled, waiting for subsystems to stop ...
╔══════════════════════════════════════════════╗
║       DefenseClaw Gateway Sidecar            ║
╚══════════════════════════════════════════════╝
...
[sidecar-api] listening on 127.0.0.1:18970
```

The hook call appears to land in the window between "context cancelled" and the new sidecar API coming back up on 127.0.0.1:18970, so the HTTP response the hook client parses is empty/invalid rather than a real allow/deny verdict.

## Expected

`setup ... --mode action` (and any command that restarts the gateway) should not return control to the caller — or should not let a hook call race — until the new sidecar API is confirmed listening and healthy. Given `fail-mode=closed` is the default, a restart race that produces a malformed response currently reads as "block everything" rather than "gateway briefly unavailable."

## Workaround

`defenseclaw setup claude-code --mode observe --yes` immediately un-wedges it. Re-enabling action mode again afterward (once the gateway is confirmed up via `defenseclaw doctor` / `defenseclaw-gateway status`) may avoid the race, but that's a manual workaround, not a fix.

## Environment

- DefenseClaw 0.8.10, Windows 11 x64
- Claude Code 2.1.187
- Connector: claudecode, hook contract claudecode-hooks-v1

vineethsai7:
Base: `main`. This is an independent, targeted fix for #681.

## Problem

`defenseclaw setup claude-code --mode action --yes` restarts the gateway, then returns into a Claude Code session whose own `PostToolUse` hook runs immediately. The existing readiness checks could combine evidence from two different gateway generations:

1. the retiring gateway could still answer `GET /health`;
2. the replacement gateway could already publish fresh connector marker files;
3. setup could return before the replacement API had actually bound the hook endpoint.

With Claude Code's fail mode set to `closed`, that short window turns an invalid or empty hook response into a block immediately after enabling enforcement.

## Current Situation

The setup path already waits for:

- `api.state == running` from `/health`; and
- fresh `active_connector.json` plus `hook_contract_lock.json` files.

Those observations were individually useful but not generation-bound. A successful health response from the old process and marker files from the new process could satisfy the two phases as an ABA race.

### Capture the generation boundary before lifecycle mutation

Before starting or restarting the gateway, setup now captures:

- the exact stable bytes of `gateway.pid`;
- the current health document's `started_at`, when an existing configured generation is reachable; and
- a local wall-clock boundary for fresh-start or unavailable-old-health cases.

PID reads use the repository's bounded, no-follow regular-file reader.

### Require one coherent replacement

The replacement readiness loop now accepts a health response only when:

- the PID record has changed;
- the current PID still identifies a DefenseClaw gateway process;
- `health.started_at` differs from the retiring generation (or is newer than the start boundary when no old health was available);
- `api.state` is `running`; and
- every requested hook connector is present in `health.connectors` with state `running`.

Connector marker freshness is still required. After those markers become ready, setup rechecks the generation-bound API and connector health before returning.

```mermaid
flowchart LR
    A["Capture old PID bytes + started_at"] --> B["Start/restart gateway"]
    B --> C{"Replacement PID record changed?"}
    C -- "No" --> C
    C -- "Yes" --> D{"Health started_at is replacement?"}
    D -- "No: retiring API" --> D
    D -- "Yes" --> E{"Connector markers fresh?"}
    E -- "No" --> E
    E -- "Yes" --> F{"Same replacement API reports Claude running?"}
    F -- "No" --> F
    F -- "Yes" --> G["Return from setup; next hook is safe"]
```

This same generation check is also used after a lifecycle-launcher timeout, so status reconciliation cannot accidentally accept the retiring API.

Fixes #681.

## What Changed (Where & How)

| Path | Change |
|---|---|
| `cli/defenseclaw/commands/cmd_setup.py` | Captures pre-restart generation evidence, factors one bounded health read, rejects old/PID-mismatched health, requires requested connectors to be running, and rechecks after marker readiness. |
| `cli/tests/test_guardrail.py` | Adds the deterministic old-API/new-marker race regression and updates lifecycle/readiness contract coverage. |
| `cli/tests/test_doctor_security_boundaries.py` | Keeps the Doctor lifecycle seam compatible with generation-aware readiness arguments. |

## Breaking Changes

- A setup-triggered gateway restart now fails closed if it cannot prove that the API response and connector runtime belong to the replacement generation.
- On a slow restart, setup may remain in the existing bounded readiness window instead of returning early on a stale `/health` response.
- No configuration, API, hook payload, or command-line flag changes.

## Open Issues / Follow-ups

- #642 tracks the stronger, separate goal of binding authenticated Doctor probes to the verified listener peer. This PR does not claim to solve that trust-boundary work; it closes the concrete lifecycle generation race in #681.

## Test Plan

Observed locally on macOS (the deterministic race test is platform-independent and also runs in the Windows Python suite):

- `uv run pytest -q cli/tests/test_guardrail.py cli/tests/test_doctor_security_boundaries.py cli/tests/test_cmd_setup_multi_connector.py`
  - **217 passed, 1 skipped, 6 subtests passed**
- `uv run pytest -q cli/tests/test_cmd_setup*.py cli/tests/test_setup*.py`
  - **447 passed, 2 skipped, 42 subtests passed**
- `uv run ruff check cli/defenseclaw/commands/cmd_setup.py cli/tests/test_guardrail.py cli/tests/test_doctor_security_boundaries.py`
  - **passed**
- `python3 -m py_compile cli/defenseclaw/commands/cmd_setup.py cli/tests/test_guardrail.py cli/tests/test_doctor_security_boundaries.py`
  - **passed**
- `git diff --check`
  - **passed**


vineethsai7:
Addressed all three review items in `1e55d204b`: boundary timestamp parsing now runs only for the no-prior-health branch and rejects timezone-naive values; deterministic older-than-boundary and naive-timestamp tests were added; and the restart-services test now asserts the exact `previous_generation` object is forwarded. Focused review suite: 172 passed, 1 skipped.

@coderabbitai review

PR #706, opened 2026-08-11
fix(installer): don't die on zero-target hook-guardian manifest (AIFW-31486)

cjsanjay:
## Summary

- Fixes [AIFW-31486](https://cisco-sbg.atlassian.net/browse/AIFW-31486) — DefenseClaw macOS install fails on fresh customer boxes when no supported connector CLI (Codex / ClaudeCode / Cursor) is installed yet.
- Replaces the zero-target `die` in `packaging/macos/install.sh` with a `warn` + proceed so the hook-guardian + hook-enumerator LaunchDaemons still bootstrap. The enumerator's 5-min tick re-renders `targets.yaml` and the guardian wires hooks the moment a connector appears — no operator action required.
- Adds `t_install_survives_zero_target_manifest` as a regression guard so this can't silently regress again.

## Root cause

Customer install.log:

```
[install] rendering initial hook-guardian manifest -> /opt/cisco/secureclient/defenseclaw/hook-guardian/targets.yaml
[install] ERROR: rendered hook-guardian manifest has zero targets despite 1 eligible user(s) and connectors=codex,cursor,claudecode. ...
[postinstall] WARNING: DefenseClaw install failed with exit code 1; continuing AVC install
```

The AVC-shipped `.pkg` runs postinstall before the user has installed Codex / ClaudeCode / Cursor, so the initial cross product renders zero rows. Old code `die`d, which:
- exited the installer with code 1,
- was caught by AVC's outer postinstall as "install failed",
- left the box with no hook-guardian daemon running,
- so hooks would never wire even after a connector was installed later.

## Fix

- `packaging/macos/install.sh` — replace the `die` with a two-line `warn` and continue to bootstrap both LaunchDaemons. Safety verified: `internal/enterprisehooks/manifest.go`'s `LoadManifest` explicitly accepts an empty targets doc (`TestLoadManifestAcceptsEmptyDocument`), and `render_targets_manifest` already emits a schema-valid empty manifest (`version: 1` + `targets:`).
- `packaging/macos/tests/test_packaging_assets.sh` — new `t_install_survives_zero_target_manifest` asserts (a) no `die` remains on the zero-target path and (b) the explicit `warn` string is present so operators grepping install.log can still tell why hooks aren't wired yet.
- `packaging/macos/tests/test_render_targets_manifest.sh` — refresh a stale comment that described the old die-on-zero-targets contract.

**Scope note:** the zero-users case (`install.sh:1071`) is intentionally untouched — that path is more likely a real enumeration/perms failure, and the ticket scoped the fix to zero connectors.

## Test plan

- [x] `bash packaging/macos/tests/run_tests.sh` — 139 passed, 0 failed (was 138 before; +1 new regression guard).
- [x] `bash -n packaging/macos/install.sh` — clean.
- [ ] CI on the release branch is green.
- [ ] On a fresh Mac with no Codex/ClaudeCode/Cursor installed, `sudo ./install.sh` completes successfully, both LaunchDaemons are bootstrapped, and after installing Codex the hook-guardian wires its hook within ~5 min without a re-run.

### DIFF
--- a/packaging/macos/install.sh
+++ b/packaging/macos/install.sh
@@ -1184,15 +1184,23 @@ if [[ "${SKIP_CONNECTOR}" != "true" ]]; then
     || die "could not reserve a private manifest staging file"
   INSTALL_TEMP_FILES+=("${MANIFEST_TMP}")
   render_targets_manifest "${SUPPORT_DIR}" "${CONNECTOR}" "${USER_LINES}" > "${MANIFEST_TMP}"
-  # Belt-and-suspenders: catch the case where user_lines was non-empty
-  # AND the connector CSV was non-empty but the rendered cross product
-  # still produced zero targets (e.g. every connector was filtered out
-  # as unsupported). Without this check the guardian would happily
-  # reconcile a "0 targets, all ok" manifest and the daemon would look
-  # green while enforcing nothing.
+  # A user_lines-non-empty × connector-non-empty cross product that
+  # still resolves to zero targets means either (a) every requested
+  # connector is unsupported (not in codex/claudecode/cursor) or
+  # (b) no eligible user has any of the requested connector CLIs
+  # installed yet (AIFW-31486: the AVC-shipped .pkg lands on boxes
+  # where Codex/ClaudeCode/Cursor haven't been installed yet).
+  #
+  # We do NOT fail the install here — bootstrapping the daemons with
+  # an empty manifest is still useful: the hook-enumerator LaunchDaemon
+  # re-renders targets.yaml on a 5-min tick, so as soon as a user
+  # installs a supported connector the guardian picks it up and wires
+  # hooks without any operator action. Failing the install would leave
+  # the customer with no reconciler running at all.
   MANIFEST_TARGETS="$(grep -c '^  - user:' "${MANIFEST_TMP}" || true)"
-  if [[ "${MANIFEST_TARGETS}" == "0" ]] && [[ -n "${USER_LINES}" ]] && [[ "${ALLOW_EMPTY_USERS}" != "true" ]]; then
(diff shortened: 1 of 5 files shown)


cjsanjay:
Follow-up commit 2650518f addresses CodeRabbit's review:

**Finding 1 (Major) — partially addressed.** Added `classify_zero_target_reason` in `installer_lib.sh` and wired it into `install.sh:1207` so the warn now distinguishes the two operationally-different zero-target modes:
- `all-unsupported` → operator must rerun with a supported `--connector` value; the enumerator's tick will NOT fix this on its own.
- `none-installed` → this is the AIFW-31486 default customer case; the 5-min enumerator tick picks it up automatically as connectors appear.

Deferred: threading a real error status through `discover_agent_version → render_targets_manifest → install.sh` so unreadable/malformed metadata takes a fatal path. That is a cross-layer refactor and is out of scope for a release-branch cherry-pick — the current change does not regress the pre-fix behavior (the old `die` was equally blind to "why zero targets"). Documented in the helper's docstring so the follow-up owner has the trail.

**Finding 2 (nit) — addressed.** Strengthened `t_install_survives_zero_target_manifest` from a source scan into a control-flow guard: it now walks the `if [[ ${MANIFEST_TARGETS} == "0" ]]` block and asserts the block contains no `die()`. Mutation-verified — injecting a `die()` inside that block flips this test red locally. Legitimate defensive `die()`s guarding the manifest file (symlink refusal, atomic-move failure) live OUTSIDE the block and are untouched.

Also fixed a bug from the first commit: the regression test function was defined but never registered with `run_case` — so it silently never ran. Registered and now passing.

Test count: 139 → 143 locally.

cjsanjay:
@coderabbitai review

PR #753, opened 2026-08-18
feat(windows): consolidated enterprise Setup + CMID bootstrap

cjsanjay:
## Summary

Consolidates the Windows enterprise work — the machine-wide CMID bootstrap from #707 and the self-contained enterprise Setup from #748 — into a single PR against `main` so all Windows changes land together.

- Machine-wide `defenseclaw enterprise windows` lifecycle (install / upgrade / repair / reconcile / verify / status / uninstall) with `requireAdministrator`, protected ProgramData ACLs, and CMID-authenticated gateway (#707 work, now landed on `windows-avc-machine-installer`).
- Separate elevated `DefenseClawSetup-Enterprise-x64.exe` bootstrap that embeds commit- and SHA-256-bound `defenseclaw-gateway.exe`, `defenseclaw-hook.exe`, `defenseclaw.exe`, `install-enterprise.ps1`, and `DefenseClawEnterprise.psm1`.
- `DefenseClawSetup-x64.exe` (the per-user product) stays strictly OSS/asInvoker and rejects any attempt to relabel it as managed-enterprise.
- Native Windows builder and public, fork-safe GitHub Actions workflow validating the enterprise Setup contracts without private-repo access or secrets.

## Relationship to existing PRs

- Supersedes #748 as a target-`main` consolidation. #748 remains open for reference until this PR merges.
- Includes #707's changes because they've already landed on the `windows-avc-machine-installer` base branch.

## Security and release contract

- PE manifest uses `requireAdministrator` for the enterprise Setup; the ordinary Setup remains `asInvoker`.
- Production builds require Cisco Authenticode credentials and an allowlisted HTTPS timestamp service.
- Unsigned builds require the existing exact disposable certification scope and explicit `--allow-unsigned`.
- Source-commit mismatch has no bypass in the enterprise builder.
- Success is `0`; incomplete/rolled-back lifecycle is `1603`.
- PR CI has read-only repository permissions and does not use `AI_COMMON_READ_TOKEN`, another secret, or private `cisco-aispg/ai-common` access.

## Validation

- `go test ./cmd/defenseclaw-enterprise-setup ./cmd/defenseclaw-setup ./internal/windowsresources ./internal/tools/windowsresources`
- `go vet` over the same set
- Windows amd64 cross-compilation of the Windows-only tests
- PowerShell AST parse, workflow YAML parse, Bash syntax check, Makefile dry-run

## Test plan

- [ ] Windows Enterprise Setup workflow green on this PR
[...]

PR #767, opened 2026-08-20
impl: Windows managed_enterprise parity — merge integration branch (5 workstreams, 10 PRs)

sanjayab23:
## Summary

Merges the `windows-enterprise-integration` long-lived branch back into `main`. Closes out the majority of `docs/WINDOWS-MANAGED-ENTERPRISE-PARITY-PLAN.md` — five of six workstreams landed as ten PRs across three weeks, every substantive concern from CodeRabbit review addressed and resolved before merge.

## What lands

Ten commits total, ten pull requests. Each PR merged into the integration branch after CI green + CodeRabbit review.

### Parity-plan implementation (5 PRs / 5 workstreams)

| Workstream | Spec | PR | CR final review | CR threads | Merged |
|---|---|---|---|---|---|
| **E** — Deterministic-build primitives for AVC | [`001-windows-deterministic-build`](https://github.com/cisco-ai-defense/defenseclaw/pull/762) | [#762](https://github.com/cisco-ai-defense/defenseclaw/pull/762) | ✅ APPROVED (3 rounds) | 13 / 13 resolved | 2026-08-19 |
| **A** — AVC-driven Windows Setup packaging (build kit + assemble.sh) | [`002-windows-avc-packaging`](https://github.com/cisco-ai-defense/defenseclaw/pull/763) | [#763](https://github.com/cisco-ai-defense/defenseclaw/pull/763) | ✅ APPROVED (3 rounds) | 26 / 26 resolved | 2026-08-19 |
| **B** — Deferred `config.yaml` / `targets.yaml` delivery | [`003-windows-deferred-config`](https://github.com/cisco-ai-defense/defenseclaw/pull/764) | [#764](https://github.com/cisco-ai-defense/defenseclaw/pull/764) | ✅ APPROVED (4 rounds) | 11 / 11 resolved | 2026-08-19 |
| **C** — Windows UI IPC (AVC gRPC surface) — initial unauthenticated cut | [`004-windows-ui-ipc`](https://github.com/cisco-ai-defense/defenseclaw/pull/765) | [#765](https://github.com/cisco-ai-defense/defenseclaw/pull/765) | ⚠️ CHANGES_REQUESTED then 100% threads resolved (see note) | 15 / 15 resolved | 2026-08-19 |
| **D** — Hook enumerator service (`DefenseClawHookEnumerator`) | [`005-windows-per-user-hook-lifecycle`](https://github.com/cisco-ai-defense/defenseclaw/pull/766) | [#766](https://github.com/cisco-ai-defense/defenseclaw/pull/766) | ✅ APPROVED (5 rounds) | 13 / 13 resolved | 2026-08-20 |

**Note on #765:** CodeRabbit posted three review rounds; each round's findings were addressed via targeted fixes and every thread was resolved via GraphQL `resolveReviewThread`. CR never re-issued a final APPROVED label after the last round (it left the review state at CHANGES_REQUESTED from an earlier round), but every substantive concern was closed. The thread-resolution count (15/15) is the operative signal of resolution.

### Windows foundation (5 PRs — pre-parity-plan setup)

These landed BEFORE the parity plan began; they split the original PR #753 (Windows enterprise scaffolding) into reviewable slices so CR could parse each chunk against the existing repo.

| PR | Slice | CR final review | CR threads | Merged |
|---|---|---|---|---|
| [#756](https://github.com/cisco-ai-defense/defenseclaw/pull/756) | Slice 1/4: gateway + config foundation | ✅ APPROVED (5 rounds) | 24 / 24 resolved | 2026-08-18 |
| [#757](https://github.com/cisco-ai-defense/defenseclaw/pull/757) | Slice 2/4: cli + enterprisehooks + managed | ⚠️ COMMENTED (2 rounds) | 24 / 26 resolved | 2026-08-19 |
| [#758](https://github.com/cisco-ai-defense/defenseclaw/pull/758) | Slice 3/4: cmd + packaging + scripts | ✅ APPROVED (3 rounds) | 20 / 20 resolved | 2026-08-19 |
| [#759](https://github.com/cisco-ai-defense/defenseclaw/pull/759) | Slice 4/4: python cli shim + docs + top-level | ✅ APPROVED (4 rounds) | 15 / 15 resolved | 2026-08-19 |
| [#760](https://github.com/cisco-ai-defense/defenseclaw/pull/760) | chore: retire PR #753's replaced script (rename cleanup) | — no CR review (trivial chore) | 0 / 0 | 2026-08-19 |

**Note on #757:** two unresolved CR threads at merge time; both were "Maintainability / Heavy lift" refactor suggestions (duplicated protected-journal read/publish logic in `windows_managed_hooks_lifecycle.go`, and `Verify` duplicating `Install`'s target-resolution prologue in `installer.go`). Neither is a correctness or security concern; both are refactor debt tracked for a follow-up cleanup PR. Merged with reviewer awareness.

**Note on #760:** trivial chore commit that deleted a superseded script. CodeRabbit's review policy skipped it (no substantive code to review). Merged with reviewer awareness.

## Review-quality summary

- **10 PRs total**, **9** received CodeRabbit review; **1** was a trivial chore (skipped correctly).
- **187 CR threads posted across the 9 reviewed PRs**; **185 (98.9%)** resolved before merge; **2 remaining** are refactor-debt suggestions in #757.
- **Zero unresolved security, correctness, or data-integrity threads** at merge time across any PR.
- **Zero CR "critical" findings** left unaddressed. Every critical finding on every PR was addressed with a code fix + thread resolution.
- All 10 PRs merged with CI green (or with known Windows-native flakiness carried forward — see "Known non-blockers" below).

## Not included (deliberate deferrals)

- **Workstream F** — Uninstall parity (per-user targeted purge on Windows). Scoped OUT after review with the packaging team confirmed the operator workflow is `defenseclaw enterprise windows uninstall --purge` (machine-scoped, matches `sudo ./uninstall.sh --purge -y` on macOS). Multi-user Windows deployment scenarios (Terminal Server / Azure Virtual Desktop) that F would target are outside the current deployment profile. The parity-plan document (untracked, archived under `defenseclaw-debug/windows-parity-windows/`) records the rationale.
- **Workstream C §4.4 follow-up — Windows UDS peer-auth mechanism.** Spec 004 landed the initial unauthenticated UDS + the compile-time GA gate (`//go:build ga` refuses a release-candidate build while `UnixPeerUnauthenticated` is reachable in the code path). The GA gate is in place; the actual auth mechanism (DACL pinning / bearer token / PID+WinVerifyTrust / out-of-band token) waits on the packaging team's choice. **Blocks GA, not beta.**
- **Documentation debt** — Two parity-plan sections (A6, B4) called for updates to `docs/WINDOWS-MACHINE-INSTALLER-INTERFACE.md`. Not blocking any release; recorded as follow-up work.

## Known non-blockers

- **Three Windows-native CI checks failed on PR #766's tip commit** (`23509ef1`): `Windows x64 Go suite`, `Windows complete Python suite / shard 3`, and the `Windows Native Required` aggregator. These will re-run on this merge PR and may reproduce; investigate case-by-case at merge time.

## Test plan

- [ ] CI re-runs on this merge PR (Go / Python / Windows-native suites).
- [ ] Confirm no regression on any main-branch job that doesn't exercise the integration surface.
- [ ] After merge, retire the `windows-enterprise-integration` long-lived branch (`git push origin --delete windows-enterprise-integration`).

## References

- `docs/WINDOWS-MANAGED-ENTERPRISE-PARITY-PLAN.md` (archived under `defenseclaw-debug/windows-parity-windows/`) — parent roadmap.
- Per-PR specs 001–005 (archived alongside the plan) — design + requirements + tasks + plan for each workstream.
- All CR review threads accessible on the individual PRs linked above.


sanjayab23:
# Code-review punch list — remaining findings

The `/code-review` slash-command ran 10 finder agents (Angles A–J) against this branch vs `main`. Consolidated below, deduped across agents, ranked by severity. **Peer-auth findings deliberately excluded** — those track parity plan §4.4 (blocks GA, not beta, waiting on packaging-team mechanism choice) and are handled outside this PR.

## Progress (as of commit d8e3c3d0)

- ✅ **CI fix** (`1f35d530`) — 3 stale `removed_services -eq 2` assertions in `enterprise-uninstall-transaction-smoke.ps1` bumped to `-eq 3` for the spec 005 D1 3-service model (gateway + guardian + enumerator).
- ✅ **Tier 1 T1.1** (`8d6c59f6`) — macOS/Linux `managedEnterpriseHTTPClient` restored (was returning `errManagedGatewayPeerUnverified` unconditionally, breaking every Codex/Claude hook call on macOS managed_enterprise).
- ✅ **Tier 1 T1.2** — verified false positive. `internal/managed/cloudreg/provider.go:47-50` documents `Config.LibPath` as optional; the private macOS-side factory registered via the build-tag overlay handles library discovery internally.
- ✅ **Tier 3 (all 9)** (`d8e3c3d0`) — T3.1 handle-based envConfig open, T3.2 CMID reparse rejection, T3.3 pre-mkdir + post-CreateTemp validate, T3.5 clear "env not set" error, T3.6 ancestor-mask rationale documented, T3.8 case-fold guardian target key, T3.9 error message includes writer SID. T3.4 documented explicit-DACL follow-up (parent-DACL inheritance is already trusted). T3.7 (route through managed writer) documented + writer-injectable seam added; behavioural swap blocked by hookwiring test suite that needs test-injectable trust seam first.
- ✅ **Tier 4 (all 5)** (`d8e3c3d0`) — T4.1 new `DEFENSECLAW_UNIX_SERVICE_ACCOUNT` env override, T4.2 explicit-drop docstring, T4.3 fail-closed non-root env trust, T4.4 parent-only trust on missing codex requirements file, T4.5 fail-closed on unrecognized `Sys()` shape.
- ✅ **Tier 5 (all 9)** (`d8e3c3d0`) — T5.1 documented enumerator writer/reader follow-up, T5.2 cached-provider refresh + setInspectionAvailability, T5.3 explicit `cloudreg.Registered()` recheck on hot-reload, T5.4 deep-copy of nested slices + typed maps, T5.5 symmetric NTP-tolerance window, T5.6 accept version==0 legacy files, T5.7 cold-start ACL error pointer, T5.8 symmetry invariant documented, T5.9 backward-compat advisory prefix.

## Remaining — Tier 2 (Cisco AI Defense endpoint data parity — the explicit user ask)

### T2.1 · 🔴 Windows AI-discovery uses SYSTEM's profile, not per-user

Windows scans use `s.opts.HomeDir` (the daemon's own home = `C:\Windows\System32\config\systemprofile`) instead of iterating `homesToScan()`. macOS iterates every logged-in user's home. Consequence: `SignalEditorExtension`, `SignalShellHistoryMatch`, `SignalLocalModel`, and `SignalDesktopApp` systematically undercount on Windows — the Cisco AI Defense endpoint receives dramatically less inventory from Windows than macOS.

- `internal/inventory/ai_discovery.go:2211` — `detectEditorExtensions`
- `internal/inventory/ai_discovery.go:2889` — `detectShellHistory`
- `internal/inventory/ai_model_files.go:948` — `modelFileScanRootsWithErrors`
- `internal/inventory/ai_discovery_platform_windows.go:207` — `windowsRegistryApplicationNames` (reads `HKCU` under SYSTEM's hive)

**Fix shape:** Windows needs a `homesToScan()` equivalent that walks `HKLM\SOFTWARE\Microsoft\Windows NT\CurrentVersion\ProfileList` (spec 005's enumerator already does this — extract + reuse) and impersonates each user via existing `WithWindowsEnterpriseTargetImpersonation` for `HKCU` + `AppData` reads. Est. 400–800 LOC.

### T2.2 · 🔴 `MergeAIDefenseEndpoint` + `AIDefenseAllowlistHost` have NO production caller

Documented feature was never implemented. `AllowedHostsProvider` interface exists (`connector.go:611, 626`) but nothing folds connectors into `FirewallConfig` at boot. Operators setting `guardrail.cisco_ai_defense.endpoint` for EU / preview tenants have inspection traffic silently dropped by the egress firewall on both platforms.

- `internal/firewall/aidefense.go:44` — `MergeAIDefenseEndpoint` dead
- `internal/gateway/connector/openclaw.go:131-135` — `AllowedHosts()` returns default only
- Missing: sidecar-bootstrap-time helper that iterates active connectors + folds `AllowedHostsProvider.AllowedHosts()` into `cfg.Firewall`, then applies `MergeAIDefenseEndpoint(cfg.CiscoAIDefense.Endpoint)`.

**Fix shape:** New assembler function called from sidecar boot before firewall rules compile. Est. 200–400 LOC.

### T2.3 · 🟠 `AIStateStore.Save` env-vs-config drift (5 call sites)

`managed.PinnedDeploymentMode()` reads `DEFENSECLAW_DEPLOYMENT_MODE` from env; sidecar reads `cfg.DeploymentMode` from disk. If the env is unset (subprocess re-exec, SCM env-inheritance bug, admin-run CLI on managed host), Save takes the non-managed branch and calls `safefile.WritePrivate` on an Administrators-owned directory → fails "not owned by caller" → discovery telemetry stops updating → macOS/Windows diverge on what reaches Cisco AI Defense.

- `internal/inventory/ai_discovery.go:3893` — primary
- `internal/gateway/client.go:631, 684`
- `internal/gateway/firstboot.go:134`
- `internal/gateway/daemon.go:1026`

**Fix shape:** Thread `cfg.DeploymentMode` (or a resolved "managed?" bool) through each call site instead of re-reading env. Est. 150–300 LOC.

## Remaining — Tier 6 (reuse / simplification)

Tier 6 tracked as a separate follow-up PR per split recommendation.

| # | Finding | Location | Notes |
|---|---|---|---|
| T6.1 | `sidIsNTService` duplicated verbatim | `internal/managed/trust_windows.go:325` + `internal/ipc/acl_windows.go:263` | extract to shared `internal/winsecurity` |
| T6.2 | `publishServiceRuntimeFile` reimplements weaker `safefile.Write` | `internal/managed/runtime_dir.go:77` | drop reparse rejection + mode-0600 clamp missing here |
| T6.3 | Hand-rolled `sortVersionsDescending` + `compareVersions` | `internal/managed/cmid_library.go:75, 87` | use `sort.Slice` + `golang.org/x/mod/semver` |
| T6.4 | `updateConfiguration` 4-attempt epoch/CAS retry over-engineered | `internal/gateway/health.go:437` | next tick self-corrects |
| T6.5 | `WriteServiceRuntimeFile` walks ancestor ACL chain twice | `internal/managed/runtime_dir.go:60` | per-token-rotation cost |
| T6.6 | `ancestor bool` threaded through 5 `trust_windows.go` signatures | `internal/managed/trust_windows.go:145` | pass a mask parameter |
| T6.7 | `inspectionMu` + 2 fields + 2 accessors for one bool+string | `internal/gateway/sidecar.go:161` | replace with `atomic.Pointer` |
| T6.8 | Restore-teardown funcs share ~120 LOC | `internal/enterprisehooks/managed_policy_windows.go:473, 564` | collapse via identity-validator func |
| T6.9 | `payloadManifest` validation duplicated verbatim | `cmd/defenseclaw-setup/main.go:186` vs `cmd/defenseclaw-enterprise-setup/main.go:78` | enterprise variant already dropped per-file SHA verify, silent drift risk |
| T6.10 | `err != io.EOF` bare compare in trailing-JSON check | `internal/gateway/sidecar.go:3769` | use `errors.Is(err, io.EOF)` |
| T6.11 | `versionComponent` asymmetry (-1 for parse-error vs parsed negatives) | `internal/managed/cmid_library.go:95, 105` |  |
| T6.12 | `os.Unsetenv` without `t.Cleanup` — test-order-dependent flake | `internal/cli/windows_enterprise_service_test.go:1651` | use `t.Setenv("", "")` pattern |
| T6.13 | Per-ACE `windows.StringToSid` for TrustedInstaller constant on every ACL walk | `internal/managed/trust_windows.go:384` | hoist to package-level `sync.Once`-cached SID |
| T6.14 | Efficiency: registry walk / manifest reload / config reload lacks mtime short-circuit | `internal/managed/runtime_dir.go:60`, `env_config_windows.go:78`, `enumerator_windows.go:180` |  |

## Overall accounting

- **10 finder agents ran** — angles A (line-by-line diff), B (removed-behavior audit), C (cross-file caller/callee), D (Go pitfalls), E (wrapper/proxy correctness), F (cross-platform parity), G (reuse hunter), H (simplification), I (efficiency), J (altitude/conventions).
- **~72 findings raised** across the 10 agents; **~35 unique after dedup**.
- **~15 peer-auth-related findings excluded** per parity plan §4.4.
- **~35 findings tracked here total.**

### Status roll-up

- **Landed in this PR: 25 findings.**
  - 3 in prior session (T1.1 verified fix, T1.2 verified false positive, CI fix).
  - 9 Tier 3 (all Windows security).
  - 5 Tier 4 (all cross-platform trust).
  - 8 Tier 5 (all health/observability except T5.1 doc-only).
- **Deferred as follow-up: 3 findings** (T3.7 needs hookwiring test-infra refactor, T3.4 needs full DACL builder, T5.1 needs cross-process state-file plumbing) — each with explicit follow-up notes in the code.
- **Open — Tier 2 (3 findings):** endpoint-data parity, largest LOC, tracked as separate PR.
- **Open — Tier 6 (14 findings):** reuse/simplification, tracked as separate PR.

**Est. remaining scope:** ~1,750–3,600 LOC across two follow-up PRs (Tier 2 + Tier 6).

vineethsai7:
PR review comments:

1. Windows IPC silently ignores the peer-auth config knobs
internal/ipc/peerauth_windows.go:78 (newCodesignValidatingListener), :56 (extractPeerIdentity)

The deferred-auth posture itself is documented and gated (authposture_gagate.go) — that part is clear. What isn't documented as intentional is that require_unix_peer, require_signing_metadata, and the team/signing/bundle allowlists are still accepted from config and then ignored on Windows, with no start-time refusal.

An operator sets require_unix_peer: true plus a signing-ID allowlist on a managed_enterprise Windows host. codesignStateLabel returns deferred_windows, one warning line is logged, and the server starts. Every process running as any Authenticated User can then connect to defenseclaw_ipc.sock and call GetHealth / GetStatsSnapshot / WatchNotifications. Nothing in config validation or the health surface says the requirement was dropped.

Suggested fix that doesn't wait on the peer-auth follow-up spec: refuse to start on Windows when any of those knobs are non-empty, rather than accepting and discarding them.

2. configuration.state can never report waiting_for_config
internal/cli/sidecar.go:129

The only production call is SetDaemonConfigLoaded(true), and it runs after loadGatewayConfigV8 already succeeded. The configuration pointer is nil until that first call, so it is never non-nil while daemonConfigLoaded is false — making the !h.daemonConfigLoaded branch of recomputeConfigurationLocked and the CONFIGURATION_STATE_WAITING_FOR_CONFIG proto value dead in production.

On a --deferred-config install, rootPersistentPreRunE blocks in waitForConfigV8Managed before runSidecar ever constructs the health object. So during the window the state exists to describe, the IPC server isn't listening at all and Secure Client gets a connect failure — not waiting_for_config. The "waiting for configuration" UI state that spec 003 REQ-19/20 exists to drive can never be rendered.

Only the unit tests call SetDaemonConfigLoaded(false), which is why CI is green.

3. setup trusted-paths add --force permanently breaks defenseclaw init
cli/defenseclaw/commands/cmd_init.py:855, cli/defenseclaw/commands/cmd_setup.py:1980

Two commands disagree about whether a prefix that fails validation may be trusted:

trusted_paths_add honors --force, and _set_config_trusted_bin_prefixes persists entry = resolved or str(raw).strip() into ai_discovery.trusted_binary_prefixes.
_validated_preinit_trusted_binary_prefixes then raises ClickException for any persisted prefix that fails the same validation. run_first_run raises ValueError on the same input.
Repro: defenseclaw setup trusted-paths add --force D:\staging\bin on a group-writable path, then defenseclaw init → "pre-init trusted binary prefix is no longer safe" before any work happens, on every subsequent run. There is no CLI recovery path; the operator has to hand-edit config.yaml, and the error names neither the file nor the fix.

Either --force shouldn't persist an entry that init will later reject, or init should quarantine and report the bad entry rather than abort.

4. inspection_available still lags a dropped provider auth
internal/gateway/sidecar.go:2035

The T5.2 comment says the added per-call Refresh means "/health cannot lag reality", and describes exactly the bug it intends to fix. But ensureCMIDProvider is only reached from newManagedInspector (line 2000) and the boot guardrail (line 3613) — not from the per-inspection path. The Inspector built at boot is reused for every tool call.

So: gateway boots, publishes inspection_available=true. Hours later the identity library stops refreshing (agent restart, token expiry). No config reload occurs, so ensureCMIDProvider is never called again. Every inspection fails upstream while /health and the guardrail detail keep reporting inspection_available=true with no inspection_error — the stale-health condition the change was written to eliminate.

5. Guardian authorization dedup key varies by identifier granularity
internal/gateway/sidecar.go:3858 (managedGuardianTargetKey)

The key is built from whichever of SID, User, UserHome is present first, so the same target described at two granularities produces two distinct keys and passes the duplicate check.

A file listing claudecode twice for one user — once with sid, once with only user_home — yields claudecode\0sid\0S-1-5-21-… and claudecode\0home\0c:\users\alice. Neither the duplicate check nor the targets map fires, so TargetCount=2, SuccessCount=2, len(ProtectedTargets)=2 all agree, the strict completeness assertion passes, and managedGuardianCoversConnectors returns covered=true for a machine where one profile was actually protected. guardian_verified is published true on partial coverage.

Note the manifest loader has the same shape (manifestTargetKey), but validateManifestPlatformTarget requires both SID and user_home on Windows, so it always keys by SID there. The authorization file has no equivalent requirement.

6. DisallowUnknownFields breaks guardian forward compatibility
internal/gateway/sidecar.go:3795

The decoder rejects unknown fields, while the version check immediately below deliberately accepts older version==0 files. Skew is handled in one direction only.

Roll the guardian to a spec that adds one field to protected_targets while the gateway on the same host is still on this build: decoder.Decode fails, managedGuardianCoversConnectors returns (false, "parse hook guardian authorization: …"), and the guardrail publishes guardian_verified=false on a correctly-protected host. Guardian-ahead-of-gateway is the ordinary direction when the guardian ships in the AVC payload.

7. Config wait aborts when the parent directory doesn't exist yet — plausible, please confirm
internal/cli/config_v8_wait.go:59

fsw.Add(parentDir) returns an error if the directory isn't there, and waitForConfigV8Managed propagates it immediately — before the loop, so the 30s poll ticker that exists to cover fsnotify gaps is never reached.

If UCB creates the directory and config.yaml as one drop, the daemon starts first, gets ErrNotExist, enters the wait, and fsw.Add fails with ENOENT. The intended 24-hour bounded wait becomes an immediate crash-restart loop. Flagged as plausible rather than confirmed: if the installer pre-creates the parent, this can't trigger — I couldn't confirm that from install-enterprise.ps1. An os.MkdirAll on the parent before fsw.Add, or treating an Add failure as "fall through to polling", would make it robust either way.

8. Teardown discards the failure cause unless --json
internal/cli/windows_managed_hooks_teardown.go:119

RunE returns a fixed "managed-hook teardown %s failed"; only the --json branch emits report.Error. A non-JSON teardown-managed-hooks prepare that fails deep inside — say ValidateTrustedRuntimeDir rejecting the state root, or the journal reporting an incomplete captured transaction that requires rollback — surfaces to the operator and the MSI log as that one string.

The difference between "rollback is required" and "noncanonical protected ownership path" is exactly what determines the next recovery step. This is the same command whose exit codes the PR carefully plumbs to 1603 for deployment systems to act on, so the detail matters here more than usual.

9. Documented socket-path override cannot work on Windows
internal/ipc/paths.go:71

The comment says operators who need to move the socket in production must set cfg.Managed.SocketPath. On Windows, validateWindowsSocketPathOverride accepts only a parent equal to <TrustedProgramData>\Cisco\Cisco Secure Client\DefenseClaw\ipc, so any real override fails the bind and the managed subsystem goes to StateError. The only accepted override is byte-identical to what rule 3 already computes — the knob reads as supported but is inert. Worth either scoping the comment to non-Windows or saying plainly that the path is fixed on Windows.

10. Manifest identity re-checks precede the read they guard
internal/enterprisehooks/manifest.go:74 — altitude

LoadManifest layers Lstat → Open → Stat → Lstat → ReadAll, but the final Lstat happens before io.ReadAll, so nothing detects a swap in the window that matters. The ~40 added lines provide only the bound that expected.Size() plus the LimitReader already give.

stableManagedArtifactDigest in connector_state.go (same PR) solves this correctly — hash twice, compare size/mtime/identity before and after. The repo now has two patterns for one invariant with no signal that this is the weaker copy. Worth extracting the rigorous one and calling it from both.

11. Guardian JSON buffer copied twice before decoding
internal/gateway/sidecar.go:3794 — efficiency

json.NewDecoder(strings.NewReader(string(data))) converts []byte → string → reader, copying up to managedGuardianAuthorizationMaxBytes (4 MiB) for no benefit. bytes.NewReader(data) is the direct form; the same function already reads the bytes with io.ReadAll a few lines above.

12. Redundant symlink test alongside IsRegular on an Lstat result
internal/gateway/connector/connector_state.go:1178 — simplification

expected.Mode()&os.ModeSymlink != 0 || !expected.Mode().IsRegular() — IsRegular() is mode&ModeType == 0, so a symlink can never satisfy it and the ModeSymlink clause can never decide the branch. Same pair recurs in manifest.go LoadManifest, ReadWindowsClaudeManagedPolicyTargets, readWindowsManagedHooksTeardownJournal, and writeWindowsManagedHooksTeardownJournal. Each site reads as though two hazards are screened, so the next person to add a case has to re-derive that half of every guard is inert.

13. Socket parent basename check subsumed by the trusted-root check
internal/ipc/server_windows.go:157 — simplification

Check 3 (filepath.Base(parent) != "ipc") can only fire for inputs check 4 would reject anyway, since check 4 requires the full parent to equal a trusted root ending in ipc. The function documents four enforced rules where three are load-bearing, which invites someone to weaken check 4 believing check 3 still anchors the parent.

14. Teardown overwrites the surviving-reference count it just set
internal/cli/windows_managed_hooks_teardown.go:341 — simplification

report.SurvivingOwnedPathReferences = 0 on the success tail clobbers what prepare and verify each stored a few lines earlier. The field is only ever observable via fail(), so threading surviving through three call layers and two return signatures buys nothing on the success path.

vineethsai (changes requested):
See latest pr comment

sanjayab23:
Addressed the 14-item review in three commits on top of PR #777's merge (`21483087`):

- **`fe4c4e24`** — cleanups + hardening (items 5, 6, 9, 10, 11, 12, 13, 14)
- **`206035f3`** — daemon boot / wait semantics (items 2, 7)
- **`5cf59c19`** — quarantine bad trusted-path prefixes (item 3)

Per-item disposition:

| # | Status | Notes |
|---|---|---|
| 1 | **Deferred** | Windows peer-auth intentionally left unimplemented in the initial cut; parity plan §4.4 owns the follow-up. No config-time refusal added — that would be inconsistent with the documented deferred-auth posture. |
| 2 | **Documented** | The `waiting_for_config` branch in `recomputeConfigurationLocked` is unreachable in production today (`daemonConfigLoaded` is flipped to `true` before `SidecarHealth` is exposed). Left in place as reserved for a future hot-reload path; removing `CONFIGURATION_STATE_WAITING_FOR_CONFIG` would be a proto wire-compat change to AVC UI clients. Comment names the wiring point. |
| 3 | **Fixed** | `setup trusted-paths add --force` no longer permanently breaks `defenseclaw init`. Both `_validated_preinit_trusted_binary_prefixes` (`cmd_init.py`) and `run_first_run` (`bootstrap.py`) now quarantine invalid persisted entries with a visible warning that names the exact `setup trusted-paths remove <path>` recovery command, then continue on the remaining valid prefixes. |
| 4 | **Already addressed** | Commit `658109b5` (before this PR's audit) already added a per-inspection `Refresh` in `ensureCMIDProvider` and calls `setInspectionAvailability` with the current outcome on both the cached and fresh paths. `/health.inspection_available` no longer lags a dropped provider auth. |
| 5 | **Fixed** | `managedGuardianTargetKey` now requires SID on Windows (returns `""` on missing SID → caller's existing `key == ""` branch rejects the row). Same-target dupes at two granularities can no longer both pass the dedup check. Non-Windows behavior unchanged. |
| 6 | **Fixed** | Dropped `DisallowUnknownFields` on the guardian authorization decoder — completes the symmetry with the already-lenient `version == 0` acceptance. Guardian-ahead-of-gateway (guardian in AVC payload) now parses cleanly. Comment explains the rationale. |
| 7 | **Fixed** | `waitForConfigV8Managed` no longer aborts when `fsw.Add(parentDir)` returns ENOENT / permission errors. Falls through to the poll ticker with a diagnostic line; the 24-hour bounded wait is preserved. |
| 8 | **Fixed** | Non-JSON teardown `RunE` now wraps the underlying error into the returned string so MSI logs and the CLI both surface the real cause — no more collapsing distinct failure classes into `"managed-hook teardown <action> failed"`. |
| 9 | **Fixed** | Scoped the `ResolveSocketPath` doc: on Windows `cfg.Managed.SocketPath` only accepts a byte-identical override; the knob is effectively non-configurable there. Linux/darwin path is unchanged. |
| 10 | **Fixed** | `LoadManifest` now re-Lstats + verifies `SameFile` + size AFTER `ReadAll`, closing the swap window. Same pattern as `stableManagedArtifactDigest` in `connector_state.go`. |
| 11 | **Fixed** | `json.NewDecoder(bytes.NewReader(data))` — drops the `[]byte → string → *strings.Reader` copy pair on the 4 MiB read buffer per authorization check. |
| 12 | **Partially fixed** | Redundant `os.ModeSymlink != 0` clauses removed at the four sites Vineeth named (manifest.go, connector_state.go's stableManagedArtifactDigest, managed_policy_windows.go's Claude target read, windows_managed_hooks_teardown.go's journal read+write). Systemic sweep across ~40 additional call sites is out of scope for this PR; will land as a follow-up cleanup. |
| 13 | **Documented** | Check 3 (`filepath.Base(parent) != "ipc"`) is subsumed by check 4 in production, but check 4 is short-circuited by `allowUnsafeSocketOverrideForTest` — check 3 is the anchor under the test hook. Added a doc comment naming this so a future reader doesn't weaken check 4 believing check 3 is defense-in-depth. |
| 14 | **Fixed** | Removed `report.SurvivingOwnedPathReferences = 0` on the success tail of `runWindowsManagedHooksTeardown` — the prepare/verify/finalize phases already record the current count and clobbering it on success loses the observation trail without any consumer benefit. |

Cross-compile clean for `windows/amd64`, `go vet ./...` clean. Native darwin test suite passes on all affected packages; the pre-existing `TestBindConnectorLifecycleConfigHomeOverridesAmbientAndRestoresIt` / `TestResolvePolicyPaths*` darwin `/var`-symlink flakes reproduce on the unmodified `HEAD` and are unrelated to these changes.

@vineethsai7 — ready for another pass.

vineethsai7:
Final Review:

#1 — Windows IPC peer-auth knobs still silently ignored. peerauth_windows.go, server.go, and authposture_gagate.go are byte-identical to the reviewed revision, and no commit message mentions it. require_unix_peer / require_signing_metadata / the allowlists are still accepted from config and discarded, with no start-time refusal.

Worth noting: the new internal/managed/cmidbroker from #777 does exactly this peer authentication — GetNamedPipeClientProcessId, verifyPipeClientTokenSID against the gateway service SID, requireLocalSystem. AF_UNIX on Windows genuinely has no peer-credential API, so the deferred posture itself is defensible — but refusing to start when those knobs are set is independent of that and is a few lines.

#4 — inspection_available lag remains. ensureCMIDProvider was rewritten in 658109b56, but for the late-arriving-library case. It's still reached only from newManagedInspector (2008) and the boot guardrail (3747), neither per-inspection, so health still only updates at inspector construction. The rewrite is an improvement — keeping the cached provider instead of nilling it lets the lane self-heal — but the staleness window is unchanged and the "/health cannot lag reality" comment at line 2110 still overstates what the code does.

### New findings possibily Introduced by the fixes

internal/ipc/paths.go:73 — the new comment cites validateManagedSocketPathOverride, which doesn't exist anywhere in the tree, and states linux/darwin validates the override. server_unix.go does no override validation at all. The old comment was incomplete about Windows; this one asserts a control that isn't implemented, which is worse for anyone auditing socket trust.

cli/defenseclaw/bootstrap.py:691 — the quarantine comments and the operator hint say the persisted list is left untouched, but line 691 (and 806) assigns the pruned list before finalize_first_run_config saves it. The entry is already gone by the time we tell the operator to remove it; running the suggested command reports not-found. The self-healing behavior is right — just the message that's inverted.

cli/defenseclaw/commands/cmd_init.py:874 — {entry!r} renders a Python repr inside a suggested shell command, so D:\staging\bin comes out as 'D:\\staging\\bin'. Same !r in the bootstrap.py warning. Bare path (or shlex.quote) for the command portion.

cli/defenseclaw/bootstrap.py:679 — the stated reason for using warnings over stderr is "so pytest's caplog ... sees them", but caplog captures logging records, not warnings.warn. Worth confirming the new tests assert via recwarn/pytest.warns and aren't passing vacuously.

internal/enterprisehooks/impersonation_windows.go:207 — windowsEnterpriseTokenAdvisoryLastEmit is never pruned. Bounded by distinct SIDs, so it's a slow leak rather than a hot one, but it's a permanent map in a service designed to run indefinitely; an RDS/Citrix host cycling profiles accumulates an entry per elevated user seen.

proto/.../secureclient.proto:55 — mirror the new health.go note here. The proto still tells UI implementers that waiting_for_config is something they'll receive, and someone will build and test that state against a value the daemon never sends.

sanjayab23:
Reopened item #1 in `7f284471`.

You're right that the config-honesty gap is orthogonal to the peer-auth deferral itself. On Windows the AF_UNIX socket genuinely has no peer-credential API — the deferred posture is defensible. But silently accepting `managed.allowed_team_ids` / `allowed_signing_ids` / `allowed_bundle_ids` from config and dropping them on the floor is not defensible: an operator setting an allowlist expecting hardening ends up with a socket whose only access boundary is the file DACL, and there's no runtime knob (managed_enterprise config is admin-locked) to correct the misread.

Fix now refuses to load a managed_enterprise config on Windows if any of those three allowlists are non-empty:

```
config: managed.allowed_team_ids, managed.allowed_signing_ids,
managed.allowed_bundle_ids cannot be set on Windows in the initial-cut
managed IPC peer-auth posture (spec 004): the AF_UNIX socket has no
peer-credential API on Windows, so any allowlist here would be silently
discarded. Remove these keys from config.yaml; the socket DACL is the
enforcement boundary until parity plan §4.4 lands the Windows peer-auth
mechanism.
```

The Windows install pipeline doesn't render these keys today (verified by grep against `packaging/windows/`), so a correctly-installed Windows deployment can never trigger this — the rejection is defense-in-depth for anyone hand-editing config.yaml. Non-Windows behavior unchanged; unmanaged builds unchanged.

Native test at `TestManagedEnterpriseWindowsPeerAuthKnobsAreRejected` exercises the validator on all platforms via `runtime.GOOS` so the happy path and the rejection path are both pinned.

Correctly your original scoping — "a few lines" — the diff is ~30 LOC of validator plus the test.

The CMID-broker peer-auth from #777 stands as the correct answer for that different transport; the AF_UNIX socket keeps its DACL-only posture until §4.4.

sanjayab23:
Round-3 follow-ups landed in `55df246d`. Per-item:

**#4 inspection_available (real fix, not just a comment)** — 658109b1 improved the late-arriving-library case by keeping the cached provider on Refresh error but you're right the staleness window is unchanged and the "cannot lag reality" comment overstated it. Fixed properly: `CiscoDefenseClawInspectClient` now carries an availability observer; every per-request `Token()` outcome fires it (nil on success so a self-healing lane clears the flag, wrapped err on failure). `newManagedInspector` binds `s.setInspectionAvailability` as the observer. Together with the construction-time signal, /health tracks reality within one inspection latency of a lane state change. Corrected the misleading `ensureCMIDProvider` header comment to name the construction-time-vs-per-request split honestly.

**paths.go:73 (fictional validator)** — `validateManagedSocketPathOverride` doesn't exist; only the Windows one does. Reworded to describe reality: linux/darwin honor `cfg.Managed.SocketPath` verbatim with start-time ownership+mode checks as the trust boundary; Windows fixes the path via `validateWindowsSocketPathOverride`. No implied cross-platform validator that isn't there.

**bootstrap.py:691 (message inversion)** — you're right. The pruned list IS assigned back and persisted by `finalize_first_run_config`, so telling the operator to `setup trusted-paths remove <entry>` after the fact is inverted (the entry is already gone). Reworded to say the list has been pruned on this run and no follow-up remove is required. Removed the stale operator-hint command on the bootstrap path.

**cmd_init.py:874 (`{entry!r}` inside shell command)** — `_validated_preinit_trusted_binary_prefixes` only filters the returned tuple for in-memory init work; the persisted list truly IS untouched there, so the "run setup trusted-paths remove" hint stays. But the entry is now wrapped via `shlex.quote` so the suggested command is cut-and-paste valid in a shell instead of rendering a Python repr.

**bootstrap.py:679 (caplog vs warnings)** — you're right, caplog captures logging records, not `warnings.warn` output. Corrected the justification comment to name the right assertion primitives (`pytest.warns` / `recwarn`). Existing behavior is unchanged; no tests were added to exercise this yet (flagged for the follow-up sweep).

**impersonation_windows.go:207 (unbounded map)** — capped `windowsEnterpriseTokenAdvisoryLastEmit` at 4096 entries with an oldest-first eviction on admission. Fine for single-user hosts (never trips); RDS/Citrix hosts churning through 4K+ profiles evict the oldest entry each time a new principal is seen, which just means the evicted entry re-emits on next observation — that's the intended posture anyway.

**secureclient.proto:55 (mirror health.go note)** — added a reserved-value note explaining `CONFIGURATION_STATE_WAITING_FOR_CONFIG` is not currently produced on the initial-boot path (the daemon doesn't have a health surface constructed during the actual wait), that it's kept for wire-compat, and that UI implementers should treat it as observable-but-rare and not hard-fail on receiving it. Keeps the proto in sync with the daemon reality.

Cross-compile clean for windows/amd64; native tests pass on affected packages.

PR #798, opened 2026-08-29
Fix/claude code hook discovery prestage

rucpande:
• ## Summary

  Fixes AIFW-32990, where managed enterprise Claude Code hooks were omitted when Claude Code was installed through Claude Desktop and version
  discovery returned empty.

  ## Changes

  - Discover Claude Code versions from Claude Desktop’s embedded bundle:

    `~/Library/Application Support/Claude/claude-code/<version>/claude.app/Contents/MacOS/claude`

  - Pre-stage per-user DefenseClaw hook artifacts when version discovery fails, without modifying native agent settings.
  - Use connector-specific runtime presence as evidence for retaining an unversioned target.
  - Keep unverified hook-contract installation gated in `action` mode by `DEFENSECLAW_ALLOW_HOOK_CONTRACT_DRIFT=1`.
  - Log enabled connectors omitted from `targets.yaml`, including the discovery and presence reason.
  - Add regression coverage for embedded discovery, pre-staging, presence fallback, and action-mode contract gating.

  ## Safety

  Unversioned targets remain subject to the existing hook-contract validation. In `action` mode, installation fails closed unless the explicit  exploratory drift override is enabled.

  AIFW-32991 and AI-discovery-based revalidation are intentionally excluded from this PR.

  ## Validation

  - Focused Go tests passed, including `internal/gateway` contract-gate tests.
  - macOS installer shell tests: 30/30 passed.
  - `bash -n` syntax validation passed.
  - `git diff --check` passed.

### DIFF
--- a/internal/cli/enterprise_hooks.go
+++ b/internal/cli/enterprise_hooks.go
@@ -40,20 +40,28 @@ import (
 )
 
 var (
-	enterpriseHookConnector     string
-	enterpriseHookUser          string
-	enterpriseHookUserHome      string
-	enterpriseHookUID           int
-	enterpriseHookGID           int
-	enterpriseHookDataDir       string
-	enterpriseHookAPIAddr       string
-	enterpriseHookProxyAddr     string
-	enterpriseHookAgentVersion  string
-	enterpriseHookManifest      string
-	enterpriseHookJSON          bool
-	enterpriseHookWatchInterval time.Duration
-	enterpriseHookWatchDebounce time.Duration
-	enterpriseHookWatchSettle   time.Duration
+	enterpriseHookConnector         string
+	enterpriseHookUser              string
+	enterpriseHookUserHome          string
+	enterpriseHookUID               int
+	enterpriseHookGID               int
+	enterpriseHookDataDir           string
+	enterpriseHookAPIAddr           string
+	enterpriseHookProxyAddr         string
+	enterpriseHookAgentVersion      string
+	enterpriseHookManifest          string
+	enterpriseHookJSON              bool
+	enterpriseHookWatchInterval     time.Duration
+	enterpriseHookWatchDebounce     time.Duration
+	enterpriseHookWatchSettle       time.Duration
+	enterpriseHookPrestageConnector string
+	enterpriseHookPrestageUser      string
+	enterpriseHookPrestageUserHome  string
+	enterpriseHookPrestageUID       int
+	enterpriseHookPrestageGID       int
+	enterpriseHookPrestageDataDir   string
+	enterpriseHookPrestageAPIAddr   string
+	enterpriseHookPrestageJSON      bool
 )
 
 const defaultEnterpriseHookManifest = "/etc/defenseclaw/hook-guardian/targets.yaml"
@@ -86,6 +94,17 @@ a systemd timer/MDM guardian. First-time installs require the agent's native
 hook config file to already exist, so broad process discovery cannot create a
 new app profile from scratch.`,
 	RunE: runEnterpriseHooksInstall,
+}
+
+var enterpriseHooksPrestageCmd = &cobra.Command{
+	Use:   "prestage",
+	Short: "Materialize hook artifacts without modifying agent configuration",
+	Long: `Materialize a connector's per-user DefenseClaw hook artifacts without
+editing the connector's native configuration. This is used by managed
+enterprise discovery when the connector is configured but its agent version
+has not yet been resolved. A later enterprise hooks reconcile activates the
+native hook configuration after a valid target is emitted.`,
+	RunE: runEnterpriseHooksPrestage,
 }
 
 var enterpriseHooksReconcileCmd = &cobra.Command{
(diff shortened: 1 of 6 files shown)


sanjayab23:
Hi @rucpande — coordinating on the customer's fix. This PR contains three logically-separate changes, and after digging into the customer's install shape (jlunde on `JLUNDE-M-2WW2`, DART bundle `0827_0914`) I've landed one of the three onto the smaller release-branch fix in #785, and want to walk through the reasoning on the other two so we can decide the disposition of this PR together.

## What I took from here into #785

**Change 1 — `_claude_desktop_embedded_version_from_home` (essential fix)**

This is the change that actually unblocks jlunde. His `~/.local/bin/claude` symlink resolves to `~/Library/Application Support/Claude/claude-code/2.1.219/claude.app/Contents/MacOS/claude`, and the existing `_native_claudecode_version_from_bin` walks the symlink chain (one hop) then stops at basename `claude`/`MacOS` — both non-semver — and returns empty. Your bundle-directory probe is the correct fix.

I ported it verbatim into #785 with commit [`da9df7119`](https://github.com/cisco-ai-defense/defenseclaw/pull/785/commits/da9df7119), including the `t_claudecode_via_claude_desktop_embedded_bundle` test (using jlunde's actual `2.1.219` so the fixture traces cleanly to the customer bundle). You're credited via `Co-Authored-By` and both the function docstring and the test docstring cite `PR #798` and `AIFW-32990`.

## Changes not taken, with justification

**Change 2 — `connector_present_for_user` presence fallback**

This one already exists on #785 with virtually identical semantics — we independently converged on the same signals (`~/.claude.json`, `~/.claude/sessions`, `~/.claude/projects`, `~/.claude/session-env` for claudecode; `~/.codex/log|history.jsonl|auth.json` for codex; `~/.cursor/extensions|argv.json` for cursor). The presence-fallback in #785 has also picked up a CodeRabbit-driven refinement: the header docstring explicitly names the *blocked* signals (the dirs `prepare_userspace_for` creates — `~/.claude`, `~/.claude/settings.json`, `~/.codex`, `~/.codex/config.toml`, `~/.cursor`, `~/.cursor/hooks.json`) so a future reviewer catches drift by inspection, plus two regression tests:

- `t_defenseclaw_prepared_state_does_not_trigger_fallback` — mirrors `prepare_{claudecode,codex,cursor}_userspace` one-to-one on a tmpdir home, stubs discovery empty, asserts zero rows emitted and each helper returns 1.
- `t_legacy_prepared_dir_then_cli_use_triggers_fallback` — legacy prepared `~/.claude/settings.json` plus real CLI `~/.claude/sessions` → fallback fires. Proves the tightened signals catch mixed real-world state (which is jlunde's).

Since it's the same design and it's already merged-into-testing on #785, taking it again would be a merge conflict without adding coverage. Independent convergence corroborates that the design is right.

**Change 3 — `enterprise hooks prestage` subcommand**

I'd like to defer this one, and I want to be direct about the reasoning because it's a substantive scope call, not a rubber-stamp. Traced through the actual behavior — [`prestage.go`](https://github.com/cisco-ai-defense/defenseclaw/pull/798/files) itself documents that PreStage:

> deliberately does not create a hook contract lock, native config backup, or native hook entry. The presence of files under ~/.defenseclaw is therefore not treated as evidence that the agent is actively protected.

So prestage writes `~/.defenseclaw/hooks/*.sh` but does **not** edit `~/.claude/settings.json` / `~/.cursor/hooks.json` / `~/.codex/config.toml`. That native-config edit is what makes hooks actually fire — it's the entry in the agent's config that points at those scripts. Nothing under `~/.defenseclaw/hooks/` alone triggers anything.

Tracing the timelines:

- **Customer's actual scenario, after Change 1 lands**: discovery succeeds → row emitted with real version → `Install()` runs the full four-step flow (`bootstrapMissingHookConfig` → `validateHookContract` passes → `conn.Setup()` writes hook scripts *and* edits native config → contract lock saved) → hooks live in seconds. The prestage branch never fires because version discovery no longer returns empty for jlunde. Same total time either way.

- **Discovery-still-empty scenario (a channel we don't yet probe)**: presence fallback fires (from Change 2, already on #785), row emitted with `agent_version: ""`. In action mode `validateHookContract` refuses `Unversioned` unless `DEFENSECLAW_ALLOW_HOOK_CONTRACT_DRIFT=1`. Guardian records `ok:false` in `protected_targets.json` — a diagnostic per-target failure, much better than a silent drop. **With prestage: scripts get pre-laid at `~/.defenseclaw/hooks/`, but the native config still isn't edited (Setup never runs). Same "hooks not live" outcome.** Prestage's early script-write is unused because the native config never references them.

- **Install-then-launch race (user installs Claude Desktop after DefenseClaw)**: enumerator's next 5-min tick picks up the new `~/Library/Application Support/Claude/claude-code/<VER>/` tree (thanks to Change 1) → next guardian reconcile runs `Install()` → hook scripts written idempotently as part of `conn.Setup()` → hooks live. Prestage would have laid the same scripts earlier, but `Setup()` overwrites them with identical content when it runs. No time saved on the critical path.

In every case the bottleneck is the native-config edit, which requires a valid contract, which requires a version. Prestage writes files that don't gate that.

Prestage would legitimately matter if the design became "operator manually references pre-known script paths from the agent config file" — that's an architecture change (decoupling script layout from native-config edit) that isn't proposed anywhere today. If that decoupling is desired later, the cleaner shape is a per-connector `PrestageArtifacts(home, uid, gid)` interface method with an audited `event_type: prestage` — not a separate CLI subcommand. I've captured that in the plan doc below.

The other concern with prestage on a release branch: it adds new Go surface (117-line `prestage.go`, new CLI subcommand, new bash-invokes-Go fork-exec per empty-version tuple per enumerator tick). Release branches on this repo have consistently taken minimum-diff `fix(...)` commits (see #706, #745, #771 as recent precedents). Landing a new subcommand is a departure from that cadence without a proven-necessary benefit for the customer this cycle.

## The larger design context

I've written a detailed migration plan for eliminating the bash renderer entirely — the whole `discover_agent_version` / `render_targets_manifest` / `prepare_userspace_for` layer moves into Go alongside each connector's implementation. It's staged as three phased PRs against `main`:

- Phase 1: Introduce Go `AgentVersionDiscoverer` / `UserspacePreparer` / `PresenceProvider` interfaces, port every bash probe (including your desktop-embedded one) into the per-connector `.go` files, add `defenseclaw-gateway enterprise hooks enumerate` subcommand. Ship behind a feature flag.
- Phase 2: Flip the default; keep bash as a runtime escape hatch.
- Phase 3: Delete bash + PowerShell renderers. `installer_lib.sh` drops from 1899 → ~800 LOC.

Draft at [`docs/PLAN-consolidate-hook-enumerator.md`](https://github.com/cisco-ai-defense/defenseclaw/blob/plan/consolidate-hook-enumerator/docs/PLAN-consolidate-hook-enumerator.md) on the `plan/consolidate-hook-enumerator` branch. Appendix A explicitly addresses prestage — if the pre-warming story is worth doing, it belongs in that Phase 1 interface as a method on the connector, not as a subcommand.

## Suggested paths forward for this PR

1. **Close #798 in favor of #785** — the customer fix is complete once #785 merges with the desktop-embedded probe folded in. Reopen the prestage piece separately against `main` if/when there's a concrete pre-warming consumer.
2. **Rebase #798 onto `main` and scope it to just the prestage subcommand** — keeping your authorship on that piece, but taking it out of the release-branch scope. Would need a real design justification for what consumes the pre-laid scripts.
3. **Leave #798 open for now** while #785 goes through review; we can close it after #785 merges to avoid duplicate release-branch traffic.

Happy to jump on a quick call if any of the tradeoffs above are worth talking through. Thanks for the fast turnaround on the desktop-embedded probe — that's the crown-jewel fix for the customer and would have taken me another cycle to write independently.

/cc @cjsanjay

PR #821, opened 2026-09-02
Fix Copilot Windows hook enforcement

ihabler:
Follow-up to #655 after it merged. This PR completes the Copilot Windows native hook path by preserving JSON stdin/stdout through the PowerShell adapter, handling Windows PowerShell 5.1 BOM/CRLF behavior, accepting Copilot's documented JSON-string toolArgs, and routing native PowerShell commands through strict host-shell policy evaluation.

Validated with the official Copilot CLI v1.0.82 (real PermissionRequest denial), 20 focused Python doctor tests, Copilot connector Go tests, and focused gateway decode/mapping/full-chain tests. Scope is intentionally Copilot-only; the other connectors landed through #655.


vineethsai7:
@coderabbitai review

The three original review threads are addressed on this branch: Doctor/Windows CI now require a single `$timeoutMS = 25000` assignment and exactly one `$hook` binding, and the PowerShell harness runs a bounded large-payload concurrent stdin probe. The later Go-helper comment is obsolete because that concurrent Go test was removed after it overloaded the native suite.

PR #833, opened 2026-09-03
feat(identity): attribute hook events and MCP inventory to the invoking end user

ihabler:
> **Stack note:** This PR is stacked on #864 and #865. Its base is `stack/pr865-dns-name-limits`, a repository-owned mirror of the reviewed #865 head (`063556bd1`), because GitHub cannot select the fork-owned #865 head branch as another PR's base.

## Problem

Under a managed enterprise install, the gateway runs as a service account. It cannot infer which interactive user invoked an agent hook, so hook events and per-profile MCP inventory can otherwise be unattributed or incorrectly attributed to the service principal.

The previous revision also treated identifier shape as proof of provenance, mixed identity fields from different sources, omitted `user.id_kind` from several v8 telemetry paths, trusted agent-controlled environment variables in OmniGent, and left the Codex notify bridge without identity headers.

## Current Situation

Identity-bearing telemetry needs a stable OS identifier and an explicit namespace without allowing an arbitrary numeric payload value to masquerade as a POSIX uid. Managed mode must also avoid falling back to the gateway service account.

The seven open review threads covered exactly those trust and propagation gaps. This revision fixes each one and updates the stale Windows Devin acceptance contract to the direct native launcher emitted by production.

### Provenance-aware identity resolution

- Classify an ID as `posix_uid` or `windows_sid` only after trusted OS/transport provenance is established.
- Keep ID and name paired to one source; a partial header is never completed from an agent-controlled payload.
- Resolve the effective POSIX uid and its matching account name, including setuid-safe behavior.
- Resolve OmniGent identity through the OS account database rather than `USER`/`LOGNAME`.
- Sanitize an inventory owner ID before deriving its kind.

### End-to-end v8 propagation

```text
OS identity -> fixed DefenseClaw headers -> loopback gateway context
                                            |-> hook/model/tool v8 events
                                            |-> HILT + router approvals
                                            |-> network-egress v8 events
Profile owner lookup ------------------------|-> MCP inventory v8 events
```

The Codex notify bridge now reads identity from the fixed `/usr/bin/id` utility and transmits the same DefenseClaw headers as the ordinary hook. `user.id_kind` is preserved through gateway context, egress, HILT, router approval, and inventory builders; untrusted values remain deliberately unclassified.

### Connector and acceptance-contract corrections

- Avoid marshalling every connector payload when only Cursor can provide hook email.
- Validate and invoke Devin's direct, POSIX-quoted native Windows hook command instead of the retired encoded PowerShell launcher.

## What Changed (Where & How)

| Area | Paths | Change |
| --- | --- | --- |
| Identity resolution | `internal/gateway/user_identity.go`, `internal/useridentity/` | Adds provenance-aware classification, source pairing, and effective-user lookup. |
| Transport identity | connector hooks and bridges under `internal/gateway/connector/` | Sends OS-derived ID/name without trusting agent environment. |
| Telemetry propagation | `internal/gateway/*observability_v8.go`, `internal/audit/network_egress*` | Carries `defenseclaw.user.id_kind` through all affected v8 families. |
| Inventory | `internal/gateway/inventory_events.go` | Re-derives kind only after owner-ID sanitization. |
| Windows acceptance | `scripts/live-connector-e2e/*windows.ps1` | Aligns Devin checks with the direct native command. |
| Regression tests | gateway, connector, audit, and identity tests | Pins all seven review findings and the Devin contract. |

## Breaking Changes

No configuration or API break. The v8 telemetry schema gains additive user-attribution fields. Untrusted numeric application IDs that were previously mislabeled `posix_uid` now correctly carry no `id_kind`.

## Open Issues / Follow-ups

None.

## Test Plan

| Command | Observed result |
| --- | --- |
| `make telemetry-check` | Pass after regenerating the stacked telemetry artifacts. |
| `make check` | Pass, including schema, v8 hard-cut, dashboard, provider-coverage, catalog, and upgrade-manifest gates. |
| `go test ./internal/gateway -run 'TestResolveHookUserIdentity\|TestResolveHTTPUserIdentity\|Test.*HILT.*V8\|Test.*RouterApproval.*'` | Pass. |
| `go test ./internal/gateway/connector -run 'TestCodexNotifyBridge\|TestDevinWindowsHookCommand\|TestOmnigent'` | Pass. |
| `go test ./internal/audit -run 'TestLogger_LogNetworkEgress'` | Pass. |
| `go test ./internal/useridentity` | Pass. |
| `uv run --frozen ruff check internal/gateway/connector/hooks/omnigent-policy.py` | Pass. |
| Windows `GOOS=windows GOARCH=amd64 CGO_ENABLED=0` compile tests for the changed Go packages | Pass. |
| GitHub Actions on `4dbddeb08` | 101 passed, 8 intentionally skipped, 0 failed, 0 pending. |


internal/gateway/connector/hookwiring_test.go
```diff
@@ -1401,7 +1413,7 @@ func main() {
 	}
 	copilotAdapter, err := renderTemplate(string(copilotTemplate), templateData{
 		HookBinaryPS:         strings.ReplaceAll(helper, "'", "''"),
-		CopilotHookTimeoutMS: copilotWindowsHookAdapterTimeoutMS,
+		CopilotHookTimeoutMS: copilotHookAdapterResultPropagationTimeoutMS,
```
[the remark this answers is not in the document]
ihabler: Addressed in cc6774e9. The Windows-only regression renders the production `copilotWindowsHookAdapterTimeoutMS` budget, runs a delayed child, asserts the documented fail-open timeout response, and verifies that the timed-out child process is no longer running.

docs/design/identity-fabric-telemetry-branch-testing.md
```diff
@@ -0,0 +1,440 @@
...
+below for why that distinction is the whole test.
+
+## Step 2 — A live gateway with a file sink
+
+There is no environment variable that turns on file output in v8; it is a
+config destination. Use an isolated home so nothing touches your real install.
+
+```bash
+go build -o /private/tmp/dc-idfab/defenseclaw ./cmd/defenseclaw
+mkdir -p /private/tmp/dc-idfab/home /private/tmp/dc-idfab/confighome
```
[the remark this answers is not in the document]
ihabler: Fixed — `mkdir -p` now runs before `go build` in that setup block, so the output directory exists on a clean machine.

internal/gateway/hook_model_observability_v8.go
```diff
@@ -423,7 +423,9 @@ func applyHookModelV8AgentFacts(
 	input.DefenseClawRequestID = hookModelV8OptionalID(meta.RequestID)
 	input.DefenseClawTurnID = hookModelV8OptionalID(meta.TurnID)
 	input.UserID = hookModelV8OptionalID(meta.UserID)
+	input.DefenseClawUserIDKind = v8UserIDKind(meta.UserIDKind)
 	input.DefenseClawUserName = hookModelV8OptionalID(meta.UserName)
```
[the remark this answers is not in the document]
ihabler: Not applying this one — we tried it, and it had to be reverted to keep CI green.

`hookModelV8OptionalText` does let display names with spaces past the gateway-side filter, but `defenseclaw.user.name` *does* carry an identifier pattern in the v8 registry: `^[A-Za-z0-9][A-Za-z0-9._:/-]*$`, 256 bytes. With the relaxed filter in place, the Windows fallback account name `MACHINE\runneradmin` reached the builder, `BuildLogModelRequest` returned an error, and the emitter failed closed — dropping the **entire audit record** rather than just the attribute. That took out several tests in the Windows x64 Go suite.

Keeping `hookModelV8OptionalID` means a non-conforming account name costs only `defenseclaw.user.name`, while the record and its `user.id` survive. Normalizing the Windows `DOMAIN\user` form so the name survives instead of being dropped is worth doing, but as a separate change against the registry pattern rather than by relaxing the client-side filter alone.

internal/gateway/llm_event_emit.go
```diff
@@ -587,8 +588,10 @@ func proxyLLMEventMeta(p *GuardrailProxy, r *http.Request, req *ChatRequest, pro
 		AgentID:        firstNonEmpty(env.AgentID, p.agentIDForRequest()),
 		AgentName:      firstNonEmpty(env.AgentName, p.agentNameForRequest(r.Header.Get("X-Agent-Name"))),
 		AgentType:      p.connectorName(),
-		UserID:         userID,
-		UserName:       userName,
+		UserID:         user.ID,
+		UserIDKind:     user.IDKind,
+		UserName:       user.Name,
+		UserEmail:      user.Email,
```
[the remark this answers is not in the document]
ihabler: The premise doesn't hold here: `resolveHTTPUserIdentity` never returns an email. It ends in `newLLMEventUser(userID, userName)`, which sets `ID`, `IDKind`, and `Name` and leaves `Email` at its zero value. Adding `UserEmail: user.Email` to the Codex notify constructor would copy an empty string and imply a capability that isn't there.

The absence is also deliberate. Hook-path email is payload-only, via `EmailFromPayloadForConnector`, which today resolves for Cursor alone. Codex keeps its address in `auth.json` under the user profile, and we stopped reading profile files from any request-influenced path: the user id on a hook request is asserted by a local hook process and is not authenticated, so a request could otherwise select whose credential file gets opened. Codex email attribution instead comes through the `ai.discovery` inventory path, where the profile owner is established by stat'ing an operator-configured directory rather than by anything in the request.

schemas/telemetry/v8/genai.yaml
```diff
@@ -732,6 +732,41 @@ attributes:
...
+  owner: defenseclaw
+  field_class: identifier
+  sensitivity: sensitive
+  cardinality: high
+  normalization:
+    id: identifier-v1
+    overrides:
+      max_utf8_bytes: 254
+      pattern: '^[A-Za-z0-9][A-Za-z0-9._%+-]*@[A-Za-z0-9][A-Za-z0-9.-]*\.[A-Za-z]{2,}$'
+    notes: preserve only an explicitly reported end-user email; never derive it from a display name, a git identity, or the service principal running the sidecar
```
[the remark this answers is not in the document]
ihabler: The gate is in place. `ai_discovery.include_user_email` defaults to `false` and is wired through both `NewSidecar` and `applyConfigReload`, and both emission sites consult it — `hookUserEmail` in `internal/gateway/user_identity.go` and the inventory owner resolution in `internal/gateway/inventory_events.go`. No address is emitted unless an operator turns it on deliberately.

The attribute is declared `field_class: identifier` with `sensitivity: sensitive` and an email-shaped normalization pattern, so the redaction engine classifies it correctly when it is enabled.

Privacy sign-off itself is an organizational step that can't be completed inside the PR; noting it here so it isn't lost, but the shipped default is off.

internal/gateway/connector/hook_user_identity_test.go
```diff
@@ -0,0 +1,252 @@
...
+// count from the same allowlist keeps the assertion exact on the accounts CI
+// and developers actually run as, which is where the silent-failure mode this
+// test exists to catch would show up.
+func identityHeaderLineCountForHost(t *testing.T) int {
+	t.Helper()
+	out, err := exec.Command("id", "-un").Output()
+	if err != nil {
+		t.Fatalf("id -un: %v", err)
+	}
+	if !hookAccountNamePattern.MatchString(strings.TrimSpace(string(out))) {
```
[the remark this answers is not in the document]
ihabler: Both fixed in dc11308f.

`hostAccountName` now trims only the command's own line ending, so an account whose name carries a leading or trailing space is classified the way the shell helper actually treats it — refused — instead of being predicted to produce a name header it never emits.

The name assertion compares the whole header against `id -un` rather than matching a prefix, so a header naming the wrong account now fails. `TestIdentityHeadersSurviveTheSystemShellsReader` makes the same full comparison.

internal/gateway/connector/hook_user_identity_test.go
```diff
@@ -0,0 +1,252 @@
...
+	if len(lines) != wantLines {
+		t.Fatalf("helper emitted %d lines, want %d (a -H and a value per header): %q", len(lines), wantLines, lines)
+	}
+	for i, expected := range want {
+		if lines[i] != expected {
+			t.Fatalf("line %d = %q, want %q (all lines: %q)", i, lines[i], expected, lines)
+		}
+	}
+	if wantLines == identityHeaderLinesWithName &&
+		!strings.HasPrefix(lines[3], "X-DefenseClaw-User-Name: ") {
```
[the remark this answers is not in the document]
ihabler: Both fixed in dc11308f.

`hostAccountName` now trims only the command's own line ending, so an account whose name carries a leading or trailing space is classified the way the shell helper actually treats it — refused — instead of being predicted to produce a name header it never emits.

The name assertion compares the whole header against `id -un` rather than matching a prefix, so a header naming the wrong account now fails. `TestIdentityHeadersSurviveTheSystemShellsReader` makes the same full comparison.

internal/gateway/inventory_events_test.go
```diff
@@ -125,6 +127,13 @@ func withManagedEnterprise(t *testing.T, on bool) {
 	t.Cleanup(func() { SetManagedEnterpriseActive(previous) })
 }
 
+func withUserEmailCollection(t *testing.T, on bool) {
+	t.Helper()
+	previous := UserEmailCollectionEnabled()
+	SetUserEmailCollectionEnabled(on)
+	t.Cleanup(func() { SetUserEmailCollectionEnabled(previous) })
+}
```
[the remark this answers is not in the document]
ihabler: No longer duplicated — `withUserEmailCollection` is defined once, at `internal/gateway/inventory_events_test.go:130`. The copy in `llm_event_emit_test.go` is gone and the package compiles; the Go suite is green on the current head.

internal/gateway/inventory_events.go
```diff
@@ -340,13 +371,64 @@ func perConnectorMCPEntriesForOS(cfg *config.Config, reg *connector.Registry, go
...
+		return llmEventUser{}
+	}
+	// This read is safe where the hook path's is not: home comes from the
+	// operator-configured profile roots and its owner was just resolved by
+	// stat'ing that directory, so no request influences whose file is opened.
+	// It is still gated, because collecting the address at all is a privacy
+	// decision independent of whether it can be attributed correctly.
+	if UserEmailCollectionEnabled() {
+		if email, err := useridentity.EmailForConnector(connectorName, home, nil); err == nil {
+			owner.Email = email
```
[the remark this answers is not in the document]
ihabler: The gate is in place. `ai_discovery.include_user_email` defaults to `false` and is wired through both `NewSidecar` and `applyConfigReload`, and both emission sites consult it — `hookUserEmail` in `internal/gateway/user_identity.go` and the inventory owner resolution in `internal/gateway/inventory_events.go`. No address is emitted unless an operator turns it on deliberately.

The attribute is declared `field_class: identifier` with `sensitivity: sensitive` and an email-shaped normalization pattern, so the redaction engine classifies it correctly when it is enabled.

Privacy sign-off itself is an organizational step that can't be completed inside the PR; noting it here so it isn't lost, but the shipped default is off.

schemas/config/v8/defenseclaw-config.schema.json
```diff
@@ -1051,6 +1051,11 @@
         "include_network_domains": {
           "type": "boolean"
         },
+        "include_user_email": {
+          "type": "boolean",
+          "default": false,
+          "description": "Opt in to attaching the end-user email address a connector reports to identity telemetry. Off by default: unlike the uid or SID it accompanies, the address identifies a person across systems and is emitted as plaintext."
+        },
```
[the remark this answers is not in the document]
ihabler: The gate is in place. `ai_discovery.include_user_email` defaults to `false` and is wired through both `NewSidecar` and `applyConfigReload`, and both emission sites consult it — `hookUserEmail` in `internal/gateway/user_identity.go` and the inventory owner resolution in `internal/gateway/inventory_events.go`. No address is emitted unless an operator turns it on deliberately.

The attribute is declared `field_class: identifier` with `sensitivity: sensitive` and an email-shaped normalization pattern, so the redaction engine classifies it correctly when it is enabled.

Privacy sign-off itself is an organizational step that can't be completed inside the PR; noting it here so it isn't lost, but the shipped default is off.

internal/gateway/user_identity.go
```diff
@@ -0,0 +1,243 @@
...
+	if userID == "" && userName == "" {
+		userID, userName = localProcessUser()
+	}
+	return llmEventUser{
+		ID: userID,
+		// Classified rather than asserted. A payload can carry an arbitrary
+		// string here — a login name, an opaque token — and labelling that
+		// posix_uid would put a value into the v8 enum that consumers then
+		// join on as a uid.
+		IDKind: useridentity.KindForID(userID),
```
vineethsai7: **`defenseclaw.user.id_kind` is asserted for untrusted numeric ids.**

`KindForID` runs over identifiers that came from an untrusted request, so an application-level numeric user id gets labelled `posix_uid`.

An OpenAI-style client POSTs `{"model":..., "user":"1001"}` through the guardrail proxy. `resolveHTTPUserIdentity` (lines 99-113) lifts `body.User="1001"`, `KindForID` matches `isPOSIXUID`, and the record ships `user.id=1001` + `defenseclaw.user.id_kind=posix_uid`. A consumer joining on `(id_kind, user.id)` then merges that application tenant with the real POSIX uid 1001 on the endpoint — a different person.

Same path via `X-User-Id`/`X-User` on the proxy and OTLP-ingest routes, and via `user_id`/`user`/`actor`/`login` in an agent-controlled hook payload (`userFieldsFromHookPayload`).

The comment right here says this is exactly what must not happen, and `TestKindForIDOnlyLabelsWhatItCanPlace` calls out the risk — but the guard only rejects non-numeric shapes, and a numeric app id is indistinguishable from a uid. Suggest asserting the kind only for ids the gateway resolved itself (hook identity headers, `useridentity` OS lookups) and leaving it absent for header/body/payload-supplied values.

internal/useridentity/identity_unix.go
```diff
@@ -0,0 +1,78 @@
...
+// survives setuid hook impersonation, plus the account name when it resolves.
+//
+// Geteuid is preferred over user.Current because user.Current consults HOME
+// and the passwd database, and a hardened hook deliberately rewrites HOME.
+func currentIdentity() Identity {
+	out := Identity{
+		ID:     strconv.Itoa(os.Geteuid()),
+		IDKind: KindPOSIXUID,
+	}
+	if current, err := user.Current(); err == nil && current != nil {
```
vineethsai7: **euid is paired with the *real* uid's account name.**

`ID` comes from `os.Geteuid()` above, but `Name` comes from `user.Current()`, which Go resolves from `syscall.Getuid()` — the real uid.

With a setuid hook running at euid=501 and real uid 0, `ID="501"` but `user.Current()` returns `root`, so `hookexec` sends `X-DefenseClaw-User-Id: 501` alongside `X-DefenseClaw-User-Name: root`. `resolveHookUser` ranks the two headers as a pair specifically to avoid "a record naming one account and identifying another" — but the pair is already inconsistent at the source, and this file's own doc comment cites setuid hook impersonation as the motivating case.

The `user.LookupId(out.ID)` fallback below only fires when `Name` is empty, so it never corrects a successful-but-wrong lookup. Looking up `out.ID` first and falling back to `user.Current()` fixes it, and matches the shell helper (`id -un` reports the *effective* name).

internal/gateway/connector/hooks/omnigent-policy.py
```diff
@@ -338,6 +343,35 @@ def _trace_headers() -> dict[str, str]:
...
+    headers: dict[str, str] = {}
+    try:
+        # os.getuid is absent on Windows, where no POSIX uid exists.
+        uid = os.getuid()  # type: ignore[attr-defined]
+    except AttributeError:
+        uid = -1
+    if uid >= 0:
+        headers["X-DefenseClaw-User-Id"] = str(uid)
+    try:
+        name = getpass.getuser()
```
vineethsai7: **`getpass.getuser()` reads agent-controlled environment.**

It resolves `LOGNAME`/`USER`/`LNAME`/`USERNAME` from the environment before consulting `pwd`. This module runs inside the Omnigent agent's own Python process, which controls that environment: with `USER=admin` exported, `os.getuid()` still returns the real 501 but this returns `admin` (which passes the allowlist), so the gateway records `user.id=501` + `defenseclaw.user.name=admin`.

The docstring says this "mirrors the account-name allowlist the POSIX hooks apply", but `_hardening.sh` uses `id -un` and `hookexec` uses the OS lookup — neither is env-derived. `pwd.getpwuid(os.getuid()).pw_name` gives the same value as the other transports.

internal/gateway/user_identity.go
```diff
@@ -0,0 +1,243 @@
...
+// and hook tokens are scoped per connector rather than per user. A local
+// account could therefore name another user's uid or SID and have the gateway
+// — running as a service account under a managed install — open that profile's
+// credential file and stamp its owner's address onto the caller's own event.
+//
+// The per-user inventory pass reads the same files without that exposure,
+// because there the profile comes from the operator-configured roots and its
+// owner is resolved by stat'ing the directory. That is where a file-backed
+// address is attributed; see inventoryHomeOwner.
+func hookUserEmail(connector string, payload map[string]interface{}) string {
```
vineethsai7, line 215: **Whole hook payload is re-marshalled on every meta build.**

`json.Marshal(payload)` runs on every `hookLLMEventMeta` call even though only `cursor` can ever yield an address.

With `include_user_email=true`, one Claude Code `PostToolUse` delivery carrying a large tool result calls `hookLLMEventMeta` from `stampHookEnvelopeIdentity` (`hook_telemetry.go:355`), `hookDecisionMeta` (`agent_hook.go:622`), `enrichAgentHookSpan` (`agent_hook.go:1402`) and `emitClaudeCodeHookLLMEvent` (`llm_event_emit.go:802`) — four full marshals on the synchronous path that gates the tool call, each so `EmailFromPayloadForConnector` can immediately return `ErrNoEmail` for `claudecode`.

`stampHookEnvelopeIdentity` then discards the result except `env.AgentPhase`/`Sequence`/`LifecycleID` — the exact "wasted work on a path that has nowhere to put the result" that `resolveHookUser`'s doc comment exists to avoid. Normalising the connector before marshalling (or only marshalling for `cursor`) removes it.

internal/gateway/otel_ingest.go
```diff
@@ -1025,23 +1025,24 @@ func (a *APIServer) emitCodexNotifyTurnCompleteLLMEvents(ctx context.Context, r
 	if provider == "unknown" {
 		provider = "codex"
 	}
-	userID, userName := userFromHTTPRequest(r, nil)
+	user := resolveHTTPUserIdentity(r, nil)
```
vineethsai7, line 1028: **Codex notify bridge sends no identity headers, so this always resolves nothing under a managed install.**

This reads the user purely from request headers, but the generated `notify-bridge.sh` (`internal/gateway/connector/codex.go:2307-2385`) sets only `X-DefenseClaw-Client` and the bearer token.

`resolveHTTPUserIdentity` finds nothing in headers, is passed a nil body, and `localProcessUser()` returns empty because `ManagedEnterpriseActive()` is true. The prompt and response records for that turn ship with no `user.id`, while `codex-hook.sh` records for the same `session_id` carry the real uid — so a per-user roll-up of turns silently under-counts.

Same omission class the new `TestIdentityHeadersSurviveTheSystemShellsReader` was written to catch, but its glob (`hooks/*-hook.sh`) doesn't reach the generated notify bridge.

internal/gateway/api.go
```diff
@@ -4100,8 +4100,7 @@ func (a *APIServer) handleNetworkEgressIngest(w http.ResponseWriter, r *http.Req
 	evt.Connector = firstNonEmpty(env.Connector, evt.Connector)
 	evt.AgentID = firstNonEmpty(identity.AgentID, env.AgentID, evt.AgentID)
 	evt.ToolID = firstNonEmpty(env.ToolID, evt.ToolID)
-	userID, _ := userFromHTTPRequest(r, nil)
-	evt.UserID = firstNonEmpty(userID, evt.UserID)
+	evt.UserID = firstNonEmpty(resolveHTTPUserIdentity(r, nil).ID, evt.UserID)
```
vineethsai7: **`.IDKind` is computed and discarded on this line.**

`resolveHTTPUserIdentity` returns the classified kind, but only `.ID` is taken — so `egress_observability_v8.go:88` and `:112` emit `user.id` with no `defenseclaw.user.id_kind`.

Same gap in `hilt_observability_v8.go:150` and `router_approval_observability_v8.go:152`, `:342`, `:380`. A consumer following the schema note ("closed namespace so a consumer can join `user.id` without inferring the platform from its shape") can't join these against the hook and inventory records that *do* carry `id_kind`, and is pushed back to shape-inference for exactly the records the enum was added to disambiguate.

internal/gateway/inventory_events.go
```diff
@@ -340,13 +371,64 @@ func perConnectorMCPEntriesForOS(cfg *config.Config, reg *connector.Registry, go
...
+// account they are signed into the connector as.
+//
+// The sidecar cannot use its own process identity here. It walks every profile
+// root on the endpoint, and under a managed install it runs as a service
+// account, so "the current user" is either the wrong user or no user at all.
+func inventoryHomeOwner(connectorName, home string) llmEventUser {
+	identity := useridentity.ForHome(home)
+	owner := llmEventUser{
+		ID:     sanitizeLLMEventUser(identity.ID),
+		IDKind: identity.IDKind,
```
vineethsai7: **`IDKind` is asserted rather than re-derived from the value actually emitted.**

`ID` above goes through `sanitizeLLMEventUser`, but `IDKind` is copied raw — unlike `newLLMEventUser` (`user_identity.go:127`), which classifies the sanitized value.

`sanitizeLLMEventUser` truncates at `maxLLMEventUserLength` and rewrites anything `needsRequestIDClean` flags. If `ForHome` ever returns an ID that sanitizing alters — a long or unusual `ProfileList` subkey on Windows — the row ships the mutated ID alongside `IDKind=windows_sid` even though `KindForID` no longer classifies it that way. Passing the sanitized value back through `useridentity.KindForID` keeps the two consistent by construction.

vineethsai7 (commented):
## Review: v8 identity / observability contract

Reviewed the full diff (`77d215a84...51b52011e`, 70 files), focused on the v8 observability contract, the hook envelope, and correlation. I could not build or run tests in my environment (Go module proxy unreachable), so this is from reading the code and tracing call paths — the findings should be confirmed against CI.

Eight findings. Seven are attached inline; the one below sits outside the diff hunks so it has no anchor.

### Lifecycle merge recovers `user.id` but not `id_kind` / `email` — `internal/gateway/llm_event_emit.go:1744`

`mergeHookSessionLifecycle` (line 1744) and `reconcileHookParent` (line 1816) recover `UserID`/`UserName` from the retained hook-session snapshot but never recover `UserIDKind` or `UserEmail`.

A `SubagentStart`/`Stop` delivery arrives with no identity headers and no payload user fields (managed install, so `localProcessUser()` returns empty). `meta.UserID` and `meta.UserIDKind` are both empty. The merge fills `UserID` from `snapshot.meta.UserID` = `S-1-5-21-...` but leaves `UserIDKind` empty, so `v8UserIDKind` returns `Absent` and the record ships `user.id` with no `defenseclaw.user.id_kind` — breaking the "id_kind travels next to user.id" invariant declared in `schemas/telemetry/v8/genai.yaml`. `UserEmail` is dropped the same way even though the snapshot holds it and collection is enabled.

---

## What checks out

The core plumbing is sound:

- `ctx` is threaded through every `hookLLMEventMeta` / `stampHookEnvelopeIdentity` / `hookDecisionMeta` call site
- The header-to-context promotion in `CorrelationMiddleware` is loopback-gated, and `registry` is never nil in production (`api.go:947`)
- `AgentRegistry.resolve` never populates `UserID`, so `resolveHookUser`'s "headers only" claim genuinely holds
- The pair-ranking rationale (never merging a header id with a payload name) is right
- `_hardening.sh` v6→v7 rides the existing monotonic `parseHookSchemaVersion` gate and the call sites are `declare -F`-guarded, so an older on-disk helper degrades instead of failing the hook
- The `mapfile`→read-loop change is correct for macOS bash 3.2, and the glob-based hook test (10 `*-hook.sh` files) closes the hand-maintained-list gap; the `.ps1` adapters are covered because they invoke the native binary through `hookexec`
- Email collection is off by default and gated at both emission sites; the payload-only hook path genuinely prevents a request-named uid from driving a cross-profile credential read
- The Devin change is a clean identity swap — old `windowsDevinBashHookCommand` is byte-identical to new `legacyWindowsDevinEncodedPowerShellHookCommandForBinary` and vice versa — so existing wirings migrate rather than orphan

## Minor / non-blocking

`HomeForID` (`internal/useridentity/identity.go:92`) plus both platform `homeForID` implementations have no non-test caller. The doc comment says it "closes the loop for the gateway", but the hook path deliberately cannot use it. Worth deleting or wiring up — not a bug.

internal/gateway/inventory_events.go
vineethsai7: Fixed in 65b8cccee. Inventory sanitizes the owner ID first and derives IDKind from the exact emitted value, keeping the pair consistent by construction.

internal/gateway/otel_ingest.go
vineethsai7: Fixed in 65b8cccee. The generated Codex notify bridge reads uid and name from the fixed /usr/bin/id utility, validates them, and sends the DefenseClaw identity headers. The transport-security regression test verifies both headers survive the bridge.

internal/gateway/user_identity.go
vineethsai7: Fixed in 65b8cccee. Numeric IDs from generic headers, request bodies, and hook payloads are now explicitly untrusted and carry no id_kind; only OS-derived DefenseClaw transport identities are classified. Regression coverage includes an application-level numeric ID.

internal/gateway/api.go
vineethsai7: Fixed in 65b8cccee. Trusted UserIDKind now flows through agent context, network-egress v8, HILT, and router approval span, log, and metric builders. The legacy v7 gateway-event schema remains unchanged.

internal/gateway/user_identity.go
vineethsai7: Fixed in 65b8cccee. hookUserEmail now returns before marshaling for every connector except Cursor, eliminating repeated large-payload marshals on connectors that cannot supply email.

internal/useridentity/identity_unix.go
vineethsai7: Fixed in 65b8cccee. Unix identity now resolves the account name from the effective uid first and only uses user.Current when its uid matches, preventing mixed setuid identity pairs.

internal/gateway/connector/hooks/omnigent-policy.py
vineethsai7: Fixed in 65b8cccee. OmniGent now resolves the account through pwd.getpwuid and os.getuid, never USER or LOGNAME. A regression test injects forged environment values and verifies they are ignored.

vineethsai7 (approved):
Reviewed the identity-fabric telemetry changes at 4dbddeb08 on top of #864/#865. All open findings were addressed, provenance and v8 propagation remain fail-closed, the Windows acceptance contracts are aligned with production, and the full current-head CI is green with no merge conflicts.

PR #843, opened 2026-09-04
fix(actionfacts): model SOCKS-preproxy plus HTTP-proxy observer chains

vineethsai7:
Stacked on top of #842 (`fix/curl-socks-upload-observers`). Review/merge that PR first; this PR's base is the stack parent, not `main`.

Fixes #773.

## Problem

DefenseClaw left HTTP(S) curl transfers that use `--preproxy SOCKS` plus `--proxy HTTP(S)` non-authoritative. The two-hop route was already modeled for FTP, but HTTP(S) could not pair origin authority, CONNECT/SNI, credentials, or plaintext request bytes with the hop that actually observes them.

## Current Situation

PR #755 closed direct HTTP(S) proxy, direct SOCKS, CONNECT, SNI, and plaintext SOCKS observer cases. `staticCurlProxyDestination` still rejects `lastProxy && lastPreproxy`, so classify marked `--preproxy`/`--proxy` as `unsupported_construct` and `--proxy-user` could not stay on the main HTTP(S) proxy. FTP already owns this chain with separate credential namespaces; HTTP(S) did not.

## Solution

Parse the exact `--preproxy SOCKS` plus `--proxy HTTP(S)` route and materialize both physical peers.

- SOCKS4a/5h transmit the main-proxy hostname; locally resolving SOCKS4/5 do not.
- When the main proxy is HTTP, SOCKS also observes plaintext origin Host/SNI, headers, and inline bodies.
- When the main proxy is HTTPS, those origin bytes stay inside TLS to the main proxy.
- `--proxy-user` binds to the main HTTP(S) proxy; only preproxy URL credentials bind to SOCKS.
- `--noproxy`, mixed `--next` groups, config/env-derived routes, HTTP preproxies, SOCKS4+IPv6 main, and pre-connect setup failures stay detection-only.
- Same-rule gateway checks correlate each value only with the exact external `NetworkConnect` observer.

```
curl --preproxy socks5h://preproxy --proxy http://main --header "X-Key: SECRET" http://origin/
        │
        ├─ SOCKS NetworkConnect preproxy
        │    sees main hostname, origin Host/header/body
        └─ HTTP NetworkConnect main
             sees --proxy-user, origin hostname, header/body
```

## What Changed (Where & How)

| File | Summary |
| --- | --- |
| `internal/actionfacts/curl_http_proxy_chain.go` | Two-hop route, dual `NetworkConnect` facts, preproxy hostname/request projectors |
[...]

Issue #770, opened 2026-08-21
Bind curl trusted-action authority to executable capability facts


vineethsai7:
## Problem

ActionFacts can prove static curl argv semantics, but it has no fact describing the resolved curl executable's version or compiled protocol/feature set. Some curl options are accepted only when a build includes a protocol or feature (for example an overwritten `--proto-default file`, positive compression/TLS/GSSAPI flags, or other compile-gated protocol aliases).

Without an executable-capability proof, promoting those forms to trusted-action authority can false-block a command that aborts during curl option parsing on a reduced build. Treating them as detection-only is safe, but leaves action-mode coverage on full-feature builds.

## Proposed outcome

Introduce a code-owned, authenticated curl capability boundary that can be consumed by ActionFacts without ambient filesystem/process probing in the synchronous decision path. It should bind at least:

- the resolved/trusted executable identity;
- curl semantic version;
- compiled protocol list;
- feature flags relevant to option-parse reachability;
- connector/session provenance and invalidation rules.

Then allow protocol/feature-dependent curl request-component proofs only when the exact required capabilities are present. Preserve detection-only behavior when identity or capabilities are unavailable, stale, ambiguous, or mismatched.

## Acceptance criteria

- Capability evidence cannot be inferred solely from GOOS, dialect, basename, or PATH spelling.
- Exact supported and unsupported build fixtures cover overwritten `--proto-default` values and positive feature-gated flags.
- Same-rule trusted-action proof stays destination/component bound.
- Capability loss or mismatch cannot mint authority.
- No synchronous unbounded process execution or filesystem probing is added to the policy path.

## Context

Follow-up from PR #755. That PR intentionally keeps build-dependent Telnet padding forms detection-only and authorizes only the build-independent Telnet subset.

vineethsai7:
[...]

PR #841, opened 2026-09-04
fix(actionfacts): prove sequential curl transfer prefixes without global authority

vineethsai7:
Stacked on top of #840 (`fix/windows-rotation-phase1`). Review/merge that PR first; this PR's base is the stack parent, not `main`.

Fixes #769.

## Problem

ActionFacts has one global parse-authority bit. In sequential curl mode, an earlier `--next` group can send bytes before a later group hits a lazy setup failure (missing form/upload file). A rule-local "ignore partial" exception would also authorize shell short-circuits, parallel mode, ancestor redirects, and eager later-group parse failures.

## Current Situation

The FTP control projector already walked sequential groups and kept earlier login bytes when a later lazy setup failed, but that walk was FTP-private. Other projectors cannot reuse it, and a caller-supplied boolean could mint the same exception. Global `Facts.Authoritative()` remains `StatusComplete` only; ordinary partial facts stay advisory.

## Solution

Add a code-owned `curlSequentialPrefixProof`. Only `proveCurlSequentialTransferPrefix` can attach the nonce; the zero value and numeric-only structs do not cover any group.

- Unconditional curl envelopes only (no wrappers, redirects, or incomplete argv).
- Eager later-group failures (`--header @file`, empty `--next`, invalid `--continue-at`) close the whole proof.
- `--parallel` requires every queued group to be setup-valid before any prefix is proved.
- Sequential lazy failures stop the prefix at the last setup-valid group; earlier groups remain covered.
- The FTP control projector consumes the proof. `StaticCurlFTPControlRequestComponentsForFacts` still requires global authority, so short-circuit, pipeline, and later-interface uncertainty stay detection-only.

```
curl group0 FTP login  --next  group1 lazy --form @/missing
        │                         │
        └─ prefix.covers(0)       └─ group1 not covered
           FTP components emit       Facts stay complete
           ForFacts may block

false && curl ...          → StatusPartial → ForFacts empty
curl ... --parallel --next → later lazy    → proof empty
curl ... --next --header @ → eager fail    → proof empty
```

## What Changed (Where & How)

| File | Summary |
| --- | --- |
| `internal/actionfacts/curl_sequential_prefix.go` | Parser-owned prefix proof and constructor |
| `internal/actionfacts/curl_sequential_prefix_test.go` | Sequential/parallel/eager/short-circuit/redirect pairing |
| `internal/actionfacts/web_transfer_parsed_classify.go` | FTP control projector consumes the proof |
[...]

PR #844, opened 2026-09-04
fix(actionfacts): bind plaintext HTTP-after-CONNECT metadata to its proxy

### LINKED ISSUE #774: Bind plaintext HTTP-after-CONNECT metadata to its proxy observer
## Problem

For an explicit HTTP(S) proxy with curl `--proxytunnel` and a plaintext `http://` origin, the proxy receives CONNECT and then observes the plaintext HTTP request inside that tunnel. DefenseClaw already binds the CONNECT hostname and exact inline body to the proxy, but origin path/query, origin authentication, and ordinary HTTP headers remain bound only to the origin network fact. A local origin reached through an external proxy can therefore leave those exact proxy-observed values detection-only.

PR #755 closes destination-hostname authority and direct/forward-proxy/SOCKS observer cases. Plaintext HTTP-after-CONNECT needs a dedicated observer projection so origin components are not copied onto a proxy fact when the post-CONNECT stream is encrypted or otherwise unproved.

## Required behavior

- Rebind already exact origin HTTP path/query/request fields, ordinary headers, and origin authentication to the explicit proxy used by that exact `--proxytunnel` target.
- Apply this only to plaintext `http://` traffic after CONNECT; keep HTTPS origin request bytes excluded absent a separately proved interception boundary.
- Preserve final `--noproxy`, `--next` group isolation, `--proxy-header` versus ordinary-header ownership, Host override behavior, and exact proxy scheme/host/port binding.
- Keep config/env-derived routes, unsupported chains, local proxies, setup failures, wrappers, redirects, and uncertain native argv detection-only.

## Tests

- HTTP and HTTPS proxy peers with plaintext HTTP origins under `--proxytunnel`.
- Path/query, ordinary header, origin auth, custom Host, and inline body observer cases.
- HTTPS origin, matching `--noproxy`, local proxy, mixed transfer groups, separate proxy-header, and first-wire failure negatives.
- Same-rule gateway checks proving each value correlates only with the exact external proxy `NetworkConnect` fact.

vineethsai7:
Stacked on top of #843 (`fix/curl-http-proxy-chain`). Review/merge that PR first; this PR's base is the stack parent, not `main`.

Fixes #774.

## Problem

For an explicit HTTP(S) proxy with curl `--proxytunnel` and a plaintext `http://` origin, the proxy receives CONNECT and then observes the plaintext HTTP request inside that tunnel. CONNECT hostname and inline bodies were already proxy-bound, but path, query, ordinary headers, and origin authentication stayed on the origin fact. A local origin through an external proxy therefore left those exact proxy-observed values detection-only.

## Current Situation

PR #755 binds CONNECT hostnames and inline proxy bodies. `staticCurlHTTPProxyTransmittedMetadata` returns early on `--proxytunnel` before path/query/auth are copied. Ordinary `--header` values are also omitted from the proxy when a separate `--proxy-header` list is present, even though those headers still travel in plaintext after CONNECT.

## Solution

Rebind already-proved origin HTTP path, query, ordinary headers, and origin authentication onto the exact HTTP(S) proxy used by that `--proxytunnel` target.

- Applies only to plaintext `http://` traffic after CONNECT.
- HTTPS origin request bytes stay excluded.
- `--proxy-header` versus ordinary-header ownership is preserved; CONNECT-phase headers are not duplicated.
- Matching `--noproxy`, mixed `--next` groups, local proxies, header-file setup failures, and config/env-derived routes stay detection-only.

```
curl -p --proxy http://proxy.example --header "X-Key: SECRET" http://127.0.0.1/path
        │
        ├─ origin NetworkDownload 127.0.0.1 (loopback)
        └─ proxy NetworkConnect proxy.example
             path/header/auth bind here → action-mode block
```

## What Changed (Where & How)

| File | Summary |
| --- | --- |
| `internal/actionfacts/curl_http_after_connect.go` | After-CONNECT projector and unique component merge |
| `internal/actionfacts/curl_http_after_connect_test.go` | HTTP/HTTPS proxy, path/query/header/auth, Host, noproxy, mixed-group |
| `internal/actionfacts/web_transfer_parsed_classify.go` | Append after-CONNECT components to proxy metadata |
| `internal/actionfacts/curl_proxy_metadata_test.go` | Tunneled query and CONNECT Host now include after-CONNECT path |
| `internal/gateway/trusted_action_context_disposition_test.go` | Same-rule enforce/audit pairs for the proxy observer |

## Breaking Changes

Authoritative action-mode now blocks `--proxytunnel` HTTP origins that send proved secrets through an external proxy after CONNECT. Previously those commands were detection-only when the origin itself was local.

## Open Issues / Follow-ups

- #770 - curl executable capability facts for HTTPS-proxy hostname and build-gated flags

## Test Plan

```
go test ./internal/actionfacts/ -count=1 -timeout 180s \
  -run 'TestStaticCurlHTTPAfterCONNECT|TestStaticCurlProxyTransmittedMetadata$'
```

Observed: `ok`

```
go test ./internal/gateway/ -count=1 -timeout 180s \
  -run 'TestTrustedActionRequestMetadataRiskPairs|TestTrustedActionCurlProxyMetadataEgress'
```

Observed: `ok`

### DIFF
--- a/internal/actionfacts/curl_http_after_connect.go
+++ b/internal/actionfacts/curl_http_after_connect.go
@@ -1,0 +1,115 @@
+// Copyright 2026 Cisco Systems, Inc. and its affiliates
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package actionfacts
+
+// staticCurlHTTPAfterCONNECTRequestComponents rebinds already-proved origin
+// HTTP path, query, ordinary headers, and origin authentication onto the
+// explicit HTTP(S) proxy that observes those bytes after CONNECT. HTTPS
+// origin request bytes stay excluded: they are encrypted inside the tunnel.
+func staticCurlHTTPAfterCONNECTRequestComponents(
+	command CommandFact,
+) []TransmittedRequestComponent {
+	proxy, parsed, ok := staticCurlProxyDestination(command)
+	if !ok || proxy.Scheme != "http" && proxy.Scheme != "https" ||
+		len(parsed.Targets) == 0 {
+		return nil
+	}
+	group := parsed.Targets[0].Group
+	if !curlProxyTunnelEnabled(parsed, group) ||
+		!staticCurlHostnameFirstWireSetupValid(command, parsed, group) {
+		return nil
+	}
+
+	var proxiedHTTP []NetworkFact
+	for _, target := range parsed.Targets {
+		if target.Group != group || !curlTargetUsesExplicitProxy(parsed, target) {
+			continue
+		}
+		targetFact, valid := webTargetFact(
+			command.ID, target.Value, NetworkDownload,
+		)
+		if valid && targetFact.Scheme == "http" {
+			proxiedHTTP = append(proxiedHTTP, targetFact)
+		}
+	}
+	if len(proxiedHTTP) == 0 {
+		return nil
+	}
+
+	origin := StaticCurlTransmittedMetadata(command)
+	component := func(value string) TransmittedRequestComponent {
+		return TransmittedRequestComponent{
+			Value: value, Scheme: proxy.Scheme,
+			Host: proxy.Host, Port: proxy.Port,
+		}
+	}
+	var components []TransmittedRequestComponent
+	appendComponent := func(value string) {
+		if value == "" {
+			return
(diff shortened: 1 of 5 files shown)


vineethsai7:
@coderabbitai review

vineethsai7:
@coderabbitai review

vineethsai7:
@coderabbitai review

vineethsai7:
The fresh CodeRabbit notes on this head are already the stacked contract:

- Two-hop after-CONNECT already falls back to `staticCurlHTTPProxyChainDestination`.
- HTTPS after-CONNECT is closed on this PR (`proxy.Scheme == "https"` returns nil). #845 then reopens that lane only when `https-proxy` is attested, and keeps the risk-pair audit-only without that fact.
- Closing all HTTPS-proxy request components on this PR would drop `--proxy-user` for common curl builds. That credential lane stays open; origin hostname/path/after-CONNECT stay capability-bound on #845.

Please re-review this head against that split rather than shrinking HTTPS proxy-user enforcement here.

vineethsai7:
Superseded by #841. This stack was consolidated into three PRs; these commits now ride inside **group 3 (actionfacts on #853)** on #841, rebuilt and verified on macOS, Linux and Windows against a `main` baseline. Review threads here remain the record for those commits.

PR #849, opened 2026-09-04
fix(openclaw): prove interceptor rewrite and fail doctor on silent bypass

vineethsai7:
Stacked on top of #848 (`fix/windows-managed-rotation-phase2`). Review/merge that PR first; this PR's base is the stack parent, not `main`.

Fixes #473.

## Problem

OpenClaw 2026.6.x can answer agent turns while `:4000` looks healthy. Direct curls to the guardrail proxy block, but `openclaw agent` / Dashboard sessions do not increment `INCOMING REQUEST` when logs cannot grow, and doctor treated a liveliness 200 as proof of agent-path enforcement.

## Current Situation

The plugin already patches `globalThis.fetch`, Node `http`/`https.request`, and the undici global dispatcher. Those undici `intercept` events were rejected by the proxy egress validator (`known|shape|passthrough` / `allow|block` only). Doctor and setup still treated OpenClaw as `STATUS_NOT_GATED` with no transport advisory and no self-test.

Live verification on OpenClaw 2026.6.8 already showed agent traffic hopping `:4000`; a stale log count is not evidence of bypass. The remaining hole is that operators cannot distinguish "proxy is up" from "interceptor rewrote an LLM URL onto the proxy."

## Solution

Keep the transport patches. Add a sentinel self-test that rewrites `https://api.openai.com/v1/chat/completions` onto the local proxy **without leaving the host**, publish `branch=selftest` on `POST /v1/events/egress`, and expose `/health.interception`. Doctor fails the OpenClaw interception row when the self-test misses. Setup and doctor warn for OpenClaw ≥2026.6.8.

```
OpenClaw LLM call
        │
        ├─ fetch / https.request / http.request / undici dispatcher
        │         │
        │         └─ rewrite to 127.0.0.1:4000 + X-DC-Target-URL
        │
        └─ plugin self-test (stubbed original fetch)
                  │
                  ├─ dest == http://127.0.0.1:{port}/v1/chat/completions
                  └─ POST /v1/events/egress branch=selftest
                            │
                            └─ /health.interception.verified
                                      │
                                      └─ defenseclaw doctor
```

## What Changed (Where & How)

| File | Summary |
| --- | --- |
| `extensions/defenseclaw/src/fetch-interceptor.ts` | Layer banner, per-host intercept notes, sentinel self-test |
| `extensions/defenseclaw/src/index.ts` | Re-verify layers on `llm-interceptor` service start |
[...]

PR #848, opened 2026-09-04
feat(windows): add native managed rotation prepare/commit/rollback

vineethsai7:
Stacked on top of #847 (`fix/doctor-inspect-to-use-bind`). Review/merge that PR first; this PR's base is the stack parent, not `main`.

Implements #736 Phase 2: a Windows-native prepare/commit/rollback participant on the shipped `DefenseClawHookGuardian` LocalSystem service. Does not close #736 or #704; disposable-host certification is still required.

## Problem

#736 Phase 1 recorded that Windows managed-enterprise rotation must not use the Linux/macOS guardian journal. Without a native participant, `setup rotate-token` could not rotate scoped hook credentials on Windows without claiming false POSIX parity at a privileged credential boundary.

## Current Situation

Phase 1 (#840) fail-closed Windows `managed_enterprise` before stop(A) or credential mutation. Linux/macOS already have versioned prepare/commit/rollback plus #733 current attestations. The shipped LocalSystem guardian installs and repairs hooks but was not a rotation participant. Closed draft PR #658 is not a product contract.

## Solution

Windows now joins the #735 coordinator through a distinct native adapter.

### Participant

`enterprise hooks rotate-prepare|commit|rollback` dispatch to a Windows journal (`windows-rotation-transaction.json`, schema `windows-guardian-rotation.v1`). The POSIX `rotation-transaction.json` path is refused. Public JSON stays the #735 contract: `ok`, `action`, `operation_id`, `generation`, `phase`, and target count.

### Preflight and identity

Every enabled target must carry a parseable SID, a canonical profile, a certified Claude/Codex connector, an active interactive session, expected non-secret B fingerprints, and current #733 attestations. Cursor remains spec 006 and is not a rotation target. Mutations require LocalSystem. Per-user writes impersonate the exact active non-elevated token for the manifest-pinned SID.

### Path invariants

Token artifacts are opened with `FILE_FLAG_OPEN_REPARSE_POINT`. Reparse points, directories, and `NumberOfLinks != 1` fail closed. Capture includes volume serial, file index, canonical `GetFinalPathNameByHandle` path, owner SID, and protected-DACL. B is re-read through the same handle proof before prepare returns or commit retires rollback material.

### Transaction

Prepare snapshots exact A, publishes B under impersonation, then writes generation-bound current attestations. Partial failure restores exact A. Commit is idempotent and operation/generation-bound after B auth. Rollback restores A bytes or absence, owner SID, protected DACL, and volume identity.

## Architecture Diagram

```
[...]

PR #847, opened 2026-09-04
fix(doctor): bind credential probes and lifecycle exec to inspected objects

vineethsai7:
Stacked on top of #846 (`fix/archive-artifact-lineage`). Review/merge that PR first; this PR's base is the stack parent, not `main`.

Fixes #642.
Fixes #643.

## Problem

Doctor verifies the managed gateway generation, then sends the long-lived bearer on a later TCP connection and later execs the lifecycle binary by pathname. A local replacement in either interval can receive the credential or run instead of the inspected object. Post-response PID/home checks and pathname custody walks cannot close those races.

## Current Situation

Authenticated `/status` probes already disable redirects, bypass proxies, and require listener plus PID/start-identity/home trust. The urllib connect still writes `Authorization` before that generation is proven to own the established socket.

Gateway restart and the Windows Cursor runtime probe already reject links, reparses, weak ACLs, and PATH overrides. `subprocess.run([path, ...])` still opens a new object after those checks.

## Solution

Bind both intervals to the inspected object, then fail closed.

- After TCP connect and before request bytes, map the ESTABLISHED peer to a PID and start identity. Loopback bearer probes without that bind never send the token.
- Linux lifecycle spawn execs the held inode through `/proc/self/fd/<n>`. Darwin and Windows re-validate path identity against the held object and refuse a replacement.
- The Windows Cursor probe holds the system PowerShell and adapter objects across the live round trip.

```
inspect PID + start identity + listener
        │
        ├─ TCP connect ── peer bind ──▶ send bearer / refuse
        │
        └─ open executable ── hold inode/handle ──▶ spawn held object / refuse
```

## What Changed (Where & How)

| File | Summary |
| --- | --- |
| `cli/defenseclaw/doctor_peer.py` | ESTABLISHED peer PID/start-identity bind for Linux, macOS, and Windows |
| `cli/defenseclaw/doctor_exec.py` | Handle-bound lifecycle spawn and documented installation custody model |
| `cli/defenseclaw/commands/cmd_doctor.py` | Peer-bound gateway bearer probes; Cursor probe holds PowerShell and adapter |
| `cli/defenseclaw/commands/cmd_setup.py` | Restart/status/stop spawn from the held executable object |
| `cli/tests/test_doctor_peer.py` | Replacement-race coverage: bearer is never sent to an unbound peer |
| `cli/tests/test_doctor_exec.py` | Linux held-inode exec; Darwin/Windows fail-closed path swap |

## Breaking Changes

[...]

PR #840, opened 2026-09-04
docs(windows): record managed rotation adapter Phase 1 decision

vineethsai7:
Stacked on top of #838 (`fix/guardian-rotation-coordinator`). Review/merge that PR first; this PR's base is the stack parent, not `main`.

Records the required #736 Phase 1 product decision. Does not close #736; Phase 2 (native prepare/commit/rollback) remains on that issue.

## Problem

#736 forbids routing Windows managed-enterprise rotation through the Linux/macOS guardian without an explicit, reviewable topology decision. Shipping the coordinator without that decision would either claim false Windows parity or leave operators without a recorded product boundary.

## Current Situation

`setup rotate-token` already fail-closes on Windows `managed_enterprise` and refuses the POSIX `rotate-prepare|commit|rollback` journal. Closed draft PR #658 is not a merged contract. The shipped `DefenseClawHookGuardian` LocalSystem service installs and repairs hooks; it is not a rotation participant. #704 stays open for Windows until a certified native adapter lands.

## Solution

Record the Phase 1 decision in spec 007 and the enterprise deployment guide, and pin the existing fail-closed coordinator message as the approved contract.

- Future Windows rotation uses the shipped LocalSystem guardian topology, not the POSIX journal and not #658.
- Qualified connectors are only families that already pass the native Windows managed-hook filter.
- Per-user writes require the exact active non-elevated token for the manifest-pinned SID.
- Missing session, untrusted DACL, or reparse/hard-link conditions fail closed.
- Certification requires native Windows CI plus the disposable-host harness. Cross-compilation is not coverage.

## Architecture Diagram

```
Linux/macOS managed_enterprise
  setup rotate-token ──▶ POSIX guardian rotate-prepare/commit/rollback

Windows managed_enterprise (this PR)
  setup rotate-token ──▶ fail closed (spec 007)
  DefenseClawHookGuardian ──▶ hook install/repair only
       │
       └── Phase 2 (still #736): native rotate-prepare/commit/rollback
           with SID/DACL/no-reparse handles + current attestations
```

## What Changed (Where & How)

| File | Summary |
| --- | --- |
| `docs/specs/007-windows-managed-rotation-adapter/README.md` | Phase 1 topology, identity, connector, and certification decision |
| `docs-site/content/docs/setup/enterprise-deployment.mdx` | Operator-facing rotation platform boundary |
| `cli/defenseclaw/rotate_token_guardian.py` | Exported fail-closed message bound to spec 007 |
[...]

PR #855, opened 2026-09-09
fix(guardian): consolidate rotation, Windows adapter and doctor binding

vineethsai7:
> Replaces #832. Stacked on #824 — review/merge that first; this PR's base is group 1's branch, not `main`.

## Problem

Twenty PRs were chained one-on-top-of-another from `main` (GitHub stack #852). Nothing could land without all of it, #849 sat 20 deep, and a single `gofmt` slip in #846 turned four PRs red. Two Critical defects in the token-rotation path were sitting unaddressed in the middle of the chain.

## Current Situation

This branch consolidates the guardian/rotation/Windows cluster: #832, #837, #838, #840, #847, #848.

#837 is included because it *creates* `internal/cli/enterprise_hooks_rotate.go` (1,011 lines) that #848 then patches — without it this group does not compile. It was not in the original grouping request; it is not a fourth PR, its commits ride inside this one.

## Solution

```
before                          after
main                            main
 |                               |
 +-- 824 .. 849  (20 deep)       +-- #824  group 1  gateway & OpenClaw
                                 |     |
                                 |     +-- THIS PR  guardian rotation & Windows
                                 |
                                 +-- #841 -> #853   actionfacts
                                 +-- #827 #828 #830 #831  independent singles
```

Alternative considered: keeping the 20-PR chain and fixing forward. Rejected — every PR inherited the one below it's CI failures, so nothing could be merged independently.

## What Changed

| Area | Summary |
| --- | --- |
| `internal/cli/enterprise_hooks*.go` | guardian readiness bound to current per-target attestations; Linux/macOS rotation prepare/commit/rollback; native Windows managed rotation |
| `cli/defenseclaw/rotate_token_guardian.py`, `cmd_setup.py` | guardian participants coordinated during scoped-token rotation |
| `cli/defenseclaw/doctor_exec.py`, `doctor_peer.py` | credential probes and lifecycle exec bound to inspected objects |
| **Critical fix** — `enterprise_hooks_rotate_windows.go` | `committed` now persisted *before* rollback material is retired |
| **Critical fix** — `enterprise_hooks_rotate.go` | directory creation re-validated under lock through an `os.Root` handle |

### The two Critical fixes

**Rollback reported success having restored nothing.** Commit deleted generation-A snapshots before persisting `committed`. If the journal write then failed, the on-disk phase stayed `prepared` with no snapshots; `windowsManagedRotationSnapshottedTargets` skips absent snapshots and `errors.Join()` over an empty slice returns `nil`, so rollback restored nothing and reported success — stranding the host on generation B while the operator was told it reverted to A. Fixed three ways: persist-before-delete, idempotent commit finishes deferred cleanup, and rollback refuses when a *prepared* journal has lost snapshot material. The guard is scoped to `prepared` deliberately: a `preparing` journal was interrupted part-way, where an absent snapshot legitimately means that target was never mutated — an unscoped guard broke `TestWindowsManagedRotationPreparingJournalRefusesResumeThenRollbackRestoresA`, which only surfaced when the suite was run on a real Windows host.

**Check-to-use race on a privileged write.** `refuseEnterpriseHookRotationSymlink` ran in preflight, before the rotation lock, so a target user who swapped their data directory for a symlink redirected the privileged `MkdirAll` outside their own home. Demonstrated: plain `MkdirAll` creates through the swap; a bound root refuses with `mkdirat: path escapes from parent`.

## Open Issues

- The final write still goes through `connector.PublishHookAPIToken`, which takes a path, so a narrow window remains at that last step. Fully closing it needs a descriptor-bound publish API in the connector package — the CodeRabbit thread is deliberately left open rather than marked resolved.
- **#832's two Major threads conflict.** "Allow deferred targets in status" needs a `pending_count` field so the readiness validator can accept `Success + Pending == Target`; "Preserve rollback compatibility" warns that adding fields to exactly those persisted structures makes prior binaries reject the file (three loaders use `DisallowUnknownFields`). This needs a wire-format versioning decision, so neither is applied here.
- Remaining Heavy-lift CodeRabbit threads (rotation lock protocol, descriptor-bound exec on Darwin, Windows ACL ancestor validation) are unaddressed and want design discussion.

## Test Plan

- `go build ./...` and `GOOS=windows go vet ./internal/cli/` clean. Note `enterprise_hooks_rotate_windows.go` is build-constrained to Windows, so a plain macOS build never compiles it.
- **Windows Server 10.0.26100**: all 17 `WindowsManagedRotation` tests pass, including two new regression tests for the Critical. Full `internal/cli` suite shows 30 failures — identical to a `main` baseline run on the same host (`safefile: refusing foreign-owned directory` under `Administrator`'s temp).
- **Linux arm64**: all 8 `EnterpriseHookRotation` tests pass, including two new TOCTOU regression tests. Full `internal/cli` 2 failures, identical to `main` baseline (with `umask 022`; a `002` umask makes the custody check reject group-writable temp dirs).
- macOS: 13 `internal/cli` failures, all pre-existing on `main` (`symbolic link in path: /tmp`).

---

## Verified results

- **GitHub CI: 100 checks, 0 failures.**
- **Windows Server 10.0.26100** (cross-compiled `GOOS=windows GOARCH=amd64` test binary, run on a real host): all **17** `WindowsManagedRotation` tests pass, including the two new regression tests for the commit-ordering Critical. Full `internal/cli`: 30 failures, **identical to a `main` baseline run on the same host** (`safefile: refusing foreign-owned directory` under `Administrator`'s temp).
- **Linux arm64**: all 8 `EnterpriseHookRotation` tests pass, including the two new TOCTOU regression tests. Full `internal/cli`: 2 failures, identical to `main` (with `umask 022` — a `002` umask makes the custody check reject group-writable temp dirs, which alone accounted for 21 spurious failures).
- Note `enterprise_hooks_rotate_windows.go` is build-constrained to Windows, so a plain macOS `go build ./...` never compiles it. Verified with `GOOS=windows go vet` and by cross-compiling the test binary.


vineethsai7:
@coderabbitai full review

PR #857, opened 2026-09-09
chore: consolidate docs, radar CI, routing telemetry and macOS app import

vineethsai7:
> Third of three consolidated PRs. Independent of #824 and #855 — this one bases directly on `main`.

## Problem

Twenty PRs were chained one-on-top-of-another from `main` (GitHub stack #852). Nothing could land without all of it.

## Current Situation

This is the "rest" group: the PRs that were not part of the gateway cluster or the guardian/rotation cluster, and not actionfacts — #827, #828, #830, #831. They share no files with each other and each rebases onto `main` cleanly; they are combined here because the split was specified as three PRs.

## Solution

```
main
 |
 +-- #824   group 1  gateway & OpenClaw interception
 |     |
 |     +-- #855  group 2  guardian rotation & Windows
 |
 +-- THIS PR  group 3  the rest
 |
 +-- #853 -> #856   actionfacts (requested separately)
```

## What Changed

| Origin | Area | Summary |
| --- | --- | --- |
| #827 | `cli/`, `docs-site/` | document atomic invalid-regex rejection |
| #828 | `.github/`, `cli/`, `scripts/` | authorize radar issues from the live job result only |
| #830 | `internal/gateway/`, `schemas/` | canonical per-request semantic-routing decision telemetry |
| #831 | `macos/DefenseClawMac/` | import standalone macOS app v1.1.19 |

10 commits, 89 files.

## Open Issues

- **#831 carries 22 unaddressed CodeRabbit threads** (6 Major) that are deliberately not fixed here. It is a vendored upstream import from `keitheobrien/defenseclaw_mac`; local edits are clobbered by the next sync, and there is no Xcode toolchain available to compile or test Swift changes. Those belong upstream or as tracked issues.
- #830 has one open Minor thread (assert routing metric correlation and values).

Verified for #831: the Cisco overlay survived the three-way merge — bundle ID `com.cisco.defenseclaw.macos`, empty `DEVELOPMENT_TEAM`, `MARKETING_VERSION` 0.8.10 matching repo `VERSION`, and `UpdateChecker` still pinned to `cisco-ai-defense/defenseclaw` with the asset-name gate and host/path pinning intact.

## Test Plan

- `go build ./...` clean, gofmt clean.
- Go and Python suites show no regressions against a `main` baseline (macOS, Linux arm64, Windows Server 10.0.26100, each baselined on the same host).
- Swift changes in #831 are **not** compiled or executed anywhere in this verification — no Xcode available. That gap is real and called out above.

---

[...]

PR #824, opened 2026-09-03
fix(gateway,openclaw): consolidate prompt inspection, tool-verdict fail-closed, fetch init and interceptor self-test

vineethsai7:
> First of three consolidated PRs. Bases on `main`. #855 stacks on this one; review/merge this first.

## Problem

Twenty PRs were chained one-on-top-of-another from `main` as GitHub stack #852. Nothing could land without all of it — #849 sat 20 deep — and a single `gofmt` slip in #846 turned four PRs red at once. The stack mixed four unrelated concerns: gateway prompt inspection, guardian token rotation, actionfacts detection, and a macOS app import.

## Current Situation

This PR consolidates the **gateway and OpenClaw interception** cluster. It carries the commits from four original PRs plus one dependency fix:

| Origin | Closes | Summary |
| --- | --- | --- |
| #824 | #717, #718 | inspect structured system-only chat prompts and Ollama dual-field prompts |
| #825 | #715 | fail closed on structurally invalid tool verdicts |
| #826 | #732, #742 | honor `fetch(request, init)` and unblock large-body peeks |
| #849 | #473 | prove interceptor rewrite; fail Doctor on silent bypass |
| — | — | bump `js-yaml` to 4.3.2 for GHSA-2883-xcg3-v3hh |

20 commits, 39 files, +2852/-302.

**#728 was closed as a duplicate of #825** — same two files, same issue #715, byte-identical logic (identical `INSPECT_ACTIONS`/`INSPECT_MODES`/`INSPECT_SEVERITIES` sets, identical `hasValidInspectActionContract` and `isInspectVerdict`, identical fail-closed vs. unavailable split); only a comment differed. Its line-count gap was just being measured against an older `main`.

## Solution

```
main
 |
 +-- THIS PR (#824)  gateway & OpenClaw interception
 |     |
 |     +-- #855   guardian rotation & Windows adapter
 |
 +-- #857   docs, radar CI, routing telemetry, macOS app
 |
 +-- #853 -> #856   actionfacts on the benchmarks branch
```

#855 stacks here rather than on `main` because both groups need this PR's `cmd_doctor.py` and `capacity_observability_v8_test.go` changes.

Alternative considered: keeping the 20-PR chain and fixing forward. Rejected — every PR inherited the failures of the one beneath it, so nothing could merge independently. Also considered splitting these four into separate PRs onto `main`; rejected because #825, #826 and #849 all touch `extensions/defenseclaw/src/index.ts` and `fetch-interceptor.ts` and would conflict.

[...]

PR #861, opened 2026-09-09
fix(deps): bump js-yaml to 4.3.2 for GHSA-2883-xcg3-v3hh

vineethsai7:
## Problem

`npm audit --omit=dev --audit-level=high` (`.github/workflows/ci.yml:1034`) fails on **every** open PR, including ones that only touch documentation. Advisory [GHSA-2883-xcg3-v3hh](https://github.com/advisories/GHSA-2883-xcg3-v3hh) was published against `js-yaml < 4.3.2` — *maxTotalMergeKeys does not limit CPU use for empty merge sources* — and the OpenClaw extension pins `js-yaml` as a direct production dependency.

## Current Situation

`extensions/defenseclaw/package.json` declared `^4.3.1` and the lockfile pinned exactly `4.3.1`. The advisory's affected range is `>= 4.0.0, < 4.3.2`; first patched version is `4.3.2`.

This is unrelated to any in-flight change — it started failing when the advisory published, so it blocks the whole queue rather than one branch.

## Solution

Bump to `4.3.2`. The declared range already permitted it, so this is a lockfile bump plus raising the floor so a fresh `npm install` cannot resolve back into the affected range.

Alternative considered: `npm audit fix --force`. Rejected — it pulls unrelated major bumps for no benefit here.

## What Changed

| File | Summary |
| --- | --- |
| `extensions/defenseclaw/package.json` | `js-yaml` `^4.3.1` -> `^4.3.2` (one line) |
| `extensions/defenseclaw/package-lock.json` | locked `js-yaml` 4.3.1 -> 4.3.2 |

No code change.

## Open Issues

None.

## Test Plan

- `npm audit --omit=dev --audit-level=high` now reports **found 0 vulnerabilities** (was `1 high severity vulnerability`).
- Extension suite unchanged: 257 tests pass, 14 skipped. The 3 unrunnable suites (`codeguard-scan`, `correlation.integration`, `inspect-hook`) bind local HTTP servers and fail identically before and after; they pass on a host where socket binding is permitted.

### DIFF
--- a/extensions/defenseclaw/package-lock.json
+++ b/extensions/defenseclaw/package-lock.json
@@ -9,7 +9,7 @@
       "version": "0.8.10",
       "license": "Apache-2.0",
       "dependencies": {
-        "js-yaml": "^4.3.1",
+        "js-yaml": "^4.3.2",
         "undici": "^7.10.0"
       },
       "devDependencies": {
@@ -864,6 +864,7 @@
       "integrity": "sha512-jp2P3tQMSxWugkCUKLRPVUpGaL5MVFwF8RDuSRztfwgN1wmqJeMSbKlnEtQqU8UrhTmzEmZdu2I6v2dpp7XIxw==",
       "dev": true,
       "license": "MIT",
+      "peer": true,
       "dependencies": {
         "undici-types": "~7.18.0"
       }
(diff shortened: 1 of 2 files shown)


vineethsai7:
Folded into #824 and #857 instead. The audit gate blocks those PRs directly, and keeping this as a separate PR would have left them red until it merged. The commit is identical in both: `js-yaml` `^4.3.1` -> `^4.3.2` plus the lockfile pin, no code change. #855 inherits it from #824.

PR #864, opened 2026-09-09
feat(ai-discovery): absorb ShadowClaw as the AI Discovery runtime planes

vineethsai7:
> **Consolidation note:** PRs #865 and #833 are incorporated into this branch. PR #864 is now the single merge PR for the complete AI Discovery runtime-plane, DNS-boundary, and end-user identity stack.

## Problem

DefenseClaw inventory answers what AI is present on an endpoint, but it did not establish which process was actively sustaining inference-like compute, which provider that process contacted, whether host actions formed a meaningful chain, or which endpoint user invoked the activity on a shared machine.

The original ShadowClaw integration also lived outside DefenseClaw. That boundary limited correlation, duplicated privileged runtime and packaging work, and prevented full in-process joins against inventory paths, process identity, and telemetry ownership.

## Current Situation

This PR absorbs ShadowClaw's useful runtime-detection behavior into DefenseClaw and consolidates the two former stacked follow-ups:

- #865 bounds decoded DNS names and parser work to protocol wire limits.
- #833 attributes hook activity and per-profile MCP inventory to the invoking endpoint user, with optional connector email evidence.

The three changes now share one linear branch, one review surface, and one CI result.

## Solution

Implement AI Discovery runtime planes in Go, broker privileged acquisition away from the network-facing gateway, correlate runtime observations with the existing inventory, emit canonical v8 telemetry, and attach trusted OS-derived endpoint identity where the execution or profile boundary can establish it.

```text
interactive user / agent hook
          |  uid/SID + account name
          v
DefenseClaw gateway (least privilege)
   |             |                    |
   |             |                    +--> inventory + user join
   |             +--> fixed helper IPC --> process/network/host/DNS planes
   +--> policy, correlation, scoring
                         |
                         v
                 canonical v8 telemetry
           local / OTLP / Splunk / AI Defense
```

### Runtime planes and correlation

- Plane A observes process CPU, memory, argv, ownership, and lineage.
- Plane B attributes connections and provider egress to processes.
- Plane C observes host activity through platform-native mechanisms.
- Native DNS capture and resolution enrich egress without spawning packet-sniffer commands.
- Correlation joins current runtime evidence to AI Discovery inventory while treating missing or incomplete coverage as unknown, never as exoneration.

### Privilege separation

`defenseclaw-sensor-helper` owns privileged acquisition on managed Linux and Windows. Its request protocol is fixed and fieldless: callers cannot select a path, pid, filter, glob, or command. The gateway keeps its existing reduced privilege boundary.

[...]

PR #865, opened 2026-09-11
fix(sensor): bound decoded DNS names to the wire limits

barryqy:
> **Stack note:** This PR is stacked on #864 (`claude/shadowclaw-defenseclaw-integration-734d0c`). PR #833 is stacked above this change through the repository mirror branch `stack/pr865-dns-name-limits`, which points at this PR's reviewed head (`063556bd1`).

## Problem

The DNS wire decoder bounded compression-pointer depth but did not bound the complete expanded name. A hostile packet could therefore amplify a small frame into a multi-kilobyte cached and emitted string. The decoder also accepted reserved DNS label types and `skipName` could advance past a truncated compression pointer.

The first implementation used 255 as the decoded text limit. RFC 1035's 255-octet limit is a wire limit: length octets and the terminating root octet consume two of those bytes, so an uncompressed textual name can contain at most 253 bytes.

## Current Situation

PR #864 adds the AI Discovery runtime DNS plane. DNS frames are untrusted input, and decoded names feed the resolver cache and telemetry, so both total parsing and exact wire-bound enforcement are security boundaries.

## Solution

- Track the expanded wire length, including every label-length octet and the terminating root octet.
- Accept the exact 255-octet wire boundary (253 textual bytes) and reject 256 octets.
- Do not add a separator/root byte when a compression pointer is the complete name.
- Reject reserved `01` and `10` label types.
- Reject truncated compression pointers without returning an offset beyond the payload.

## What Changed (Where & How)

| Path | Change |
| --- | --- |
| `internal/sensor/dnscapture/dns.go` | Enforces exact DNS wire-name and label-type limits during decode and skip operations. |
| `internal/sensor/dnscapture/dns_test.go` | Covers the exact valid boundary, one-byte overflow, reserved types, truncated pointers, and maximum names reached through compression. |

## Breaking Changes

None. The decoder now rejects malformed DNS names that RFC-compliant resolvers cannot produce.

## Open Issues / Follow-ups

None.

## Test Plan

| Command | Observed result |
| --- | --- |
| `go test ./internal/sensor/dnscapture ./internal/sensor/...` | Pass. |
| `go test -race ./internal/sensor/...` | Pass on the original change set. |
| Windows and Linux cross-compilation of the sensor packages | Pass. |
| GitHub Actions on `063556bd1` | 81 passed, 8 intentionally skipped, 0 failed, 0 pending. |


vineethsai7 (approved):
Reviewed the stacked DNS decoder changes at 063556bd1. Exact RFC wire-length handling, compression-pointer boundaries, and malformed-label coverage are sound. Full current-head CI is green and there are no unresolved review threads.

PR #884, opened 2026-09-18
fix(managed): make ancestor trust verdicts advisory on Windows and macOS (AIFW-34262)

cjsanjay:
## AIFW-34262 — DefenseClaw status shows "Unavailable" on Windows

A managed install aborted mid-flight and left the endpoint unprotected: all four services `Disabled`, AVC tile `Unavailable`. The DART shows two trust failures naming the same directory —

```
...\Cisco\Cisco Secure Client\DefenseClaw
untrusted Windows principal S-1-5-21-...-1001 has write-like access mask 0x1f01ff
```

— once from the hook-guardian manifest check and once from the managed config check. That directory is **not** DefenseClaw's to own: Cisco Secure Client / AVC creates the `Cisco Secure Client` tree, shares it with other Cisco software, and re-ACLs it on its own schedule. Something re-introduced a local-user FullControl ACE there while the install was running. Because ancestor permission verdicts were fatal, the install failed the live verify, then **failed the rollback on the same check** — and quiesce had already disabled every service. A transient grant on a parent directory we don't own bricked the endpoint.

## Change

Ancestor directories above the managed roots now emit an advisory and continue instead of failing. Permissions on that shared parent are AVC's responsibility; we log and keep going.

The core of it is splitting one overloaded flag into two independent decisions. `internal/managed/trust_windows.go` previously carried a single `ancestor bool` that meant *both* "use the narrower replacement-rights mask" *and* "don't fail" — so anything needing the narrower mask silently lost the ability to refuse. Now:

```go
type windowsTrustScope struct {
	narrowMask bool
	advisory   bool
}

var (
	windowsTrustLeaf     = windowsTrustScope{}
	windowsTrustNamedDir = windowsTrustScope{narrowMask: true}
	windowsTrustAncestor = windowsTrustScope{narrowMask: true, advisory: true}
)
```

That matters for correctness, not just tidiness: `ValidateTrustedDirectoryAncestor`'s **named** directory is used as a pre-write and pre-delete guard on directories DefenseClaw creates and ACLs itself (hook-guardian manifests, managed policy, namespace purge). Those keep every verdict fatal via `windowsTrustNamedDir`. The AIFW-34262 grant lands one level above them, on `...\Cisco Secure Client\DefenseClaw`, so it is reached as an ancestor and downgraded.

A `trustVerdict` wrapper type separates *permission judgements* from *structural and OS/exec failures*, so only the former are downgradable. `RelaxAncestorTrustJudgement` passes anything else straight through.

### What stays fatal

- the named path itself — owner, mode, DACL, ACL;
- structural failures anywhere: missing path, symlink, reparse point, wrong file type, non-canonical deny ACE;
- OS/API/exec failures anywhere: `GetAce`, security-descriptor reads, attribute reads, `/bin/ls -lde` timing out or being killed, xattr reads;
- in the PowerShell installer: null DACL, untrusted owner on the state root itself, missing required rights;
- ancestors **outside** the platform installer's tree.

### Scope gate

`managed.PlatformInstallerOwnedPath` is a path allowlist — `%ProgramData%\Cisco`, `%ProgramFiles%\Cisco`, `%ProgramFiles(x86)%\Cisco`, `/opt/cisco`, `/Library/Logs/Cisco`. Deliberately not a deployment-mode gate: macOS takes its mode from the protected `config.yaml` rather than an env pin, so a mode check would be dead code there; and a mode gate would relax ancestors that have nothing to do with Cisco (a world-writable temp or home above a per-user runtime path), which is backwards for the unmanaged product.

### Kill switch

`DEFENSECLAW_MANAGED_TRUST_STRICT_ANCESTORS=1|true|yes|on` restores every fatal verdict, on all four surfaces (Go managed, Go connector, psm1, install.sh). It is an **inverted** flag — setting it only hardens — and is registered in `internal/envvars/registry.json` with all five consumers.

## Both platforms, all four trust walks

| Surface | File |
|---|---|
| Go managed trust walk (Windows) | `internal/managed/trust_windows.go` |
| Go managed trust walk (unix/macOS) | `internal/managed/trust_unix.go`, `trust_acl_darwin.go` |
| Windows transactional installer | `packaging/windows/DefenseClawEnterprise.psm1` |
| macOS installer | `packaging/macos/install.sh` |

The second commit extends the same rule to a subsystem with its **own** trust walk: the gateway connector's hook API token path chain. It can't call the managed validators because it has a different owner model (it accepts the invoking uid so unmanaged per-user installs work), but it walks the same AVC-owned ancestors — `hookAPIValidateDirectoryChain` on `/opt/cisco/secureclient` on macOS, and the Windows sibling whose stock-ancestor allowance covers only `BUILTIN\Users`, so the DART's local-user ACE would still have been fatal there. Both now route ancestor verdicts through the exported helpers, scoped by the same allowlist.

Also touched: `packaging/macos/install.sh`'s `ensure_shared_install_parent` for `/opt`, `/opt/cisco`, `/opt/cisco/secureclient` and `/Library/Logs`, `/Library/Logs/Cisco`, `/Library/Logs/Cisco/SecureClient` — the three verdicts (not root-owned, group/other writable, write-capable ACL) became advisories; `not a real directory`, `stat` and `ls -lde` failures still `die`.

Audited and deliberately left alone: `internal/config/env_config_unix.go` (leaf-only `Fstat`, correctly fatal), `packaging/macos/uninstall.sh` and `lib/installer_lib.sh` (already warn), `hook_api_acl_other.go` (capability failure stays fatal).

## Tests

`internal/managed/trust_advisory_test.go`, `trust_windows_test.go`, `internal/gateway/connector/hook_api_token_advisory_unix_test.go` cover: leaf stays fatal; ancestor warns exactly once with path, label and reason; the strict pin restores fatality for `1/true/YES/on` and an explicit `0` does not; structural and exec failures are never swallowed; the path allowlist rejects a prefix sibling (`root + "-evil"`) and the root's own parent; a world-writable non-Cisco ancestor is still fatal and emits zero advisories; and `D:P(A;;GA;;;BU)(A;;GA;;;BA)` is fatal for `windowsTrustNamedDir` but advisory for `windowsTrustAncestor`.

Verified locally: `gofmt` clean, `go build ./...`, `go vet`, `GOOS=windows` and `GOOS=linux` build + vet, and green `go test ./internal/managed/... ./internal/gateway/connector/ ./internal/config/... ./internal/envvars/`. Rebased onto `c9c2df76` and re-verified.

## Please check in CI / on a Windows host

- **`packaging/windows/DefenseClawEnterprise.psm1` has never been syntax-checked** — there is no `pwsh` on the macOS host this was written on. It needs a PowerShell lint/parse and an install smoke run.
- Windows Go tests could not be executed locally either; only `GOOS=windows` build and vet ran. The Windows-only trust tests need a real Windows runner.
- A managed install on a box with a foreign FullControl ACE on `...\Cisco Secure Client\DefenseClaw` is the actual AIFW-34262 repro — it should now complete with a `managed_trust_ancestor_advisory` line in the log instead of aborting.


cjsanjay:
## Added: self-heal for the shared state root ACL (3rd commit)

Making foreign access advisory closed one brick but left two more on the *same* directory, both because AVC can **replace** the canonical DACL we stamp on `...\Cisco Secure Client\DefenseClaw` rather than merely adding an ACE to it:

1. `Set-DefenseClawPathAcl` stamps the canonical descriptor and then demands exact equality with it. An AVC pass landing in that window threw `managed path does not have the exact canonical DACL` and failed the transaction.
2. The post-hardening `Assert-DefenseClawPathAcl` threw on a foreign owner, a null DACL, a deny ACE, or missing required rights **even with** `-AdvisoryUntrustedAccess`.

So: repair rather than relax further. `Set-DefenseClawSharedPathAcl` retries the stamp with a backoff and, if it keeps losing, defers the verdict to the post-hardening assertion, which now takes `-SelfHealKind` — it re-stamps the canonical ACL once and re-reads the path before judging it.

To make a repair possible *between* reading and judging, `Assert-DefenseClawPathAcl` is split: `Get-DefenseClawPathAclVerdicts` reads the DACL and returns verdicts classified by what they actually mean.

| Kind | Meaning | Policy |
|---|---|---|
| `Access` | Somebody else **also** has access (foreign writer/reader, inherited DACL) | Advisory under `-AdvisoryUntrustedAccess`, as before |
| `Contract` | The descriptor on disk isn't the one we stamped (null DACL, foreign owner, deny ACE) | Repaired by re-stamp; advisory only if the repair didn't take |
| `Rights` | **We** lack rights we must have | Never downgraded |

That last row is the boundary, and it's the answer to "does this mean the install never fails on permissions": the relaxation covers *others having access* to a directory AVC owns, never *DefenseClaw lacking the access it needs*. A state root that still denies SYSTEM/Administrators/gateway their required rights after a re-stamp is a broken host, and warning there would only hide a gateway that can't start.

### Preserved boundaries

- `Get-Acl` failures are not verdicts — they propagate, so an unreadable security descriptor can never become a warning.
- `Assert-DefenseClawNoReparsePath` still runs first and stays fatal.
- The repair runs even under `DEFENSECLAW_MANAGED_TRUST_STRICT_ANCESTORS`, because re-stamping only hardens. The pin gates the final downgrade and makes the first lost stamp race fatal again.
- **No behaviour change for the ~40 call sites that pass neither switch:** the first verdict in reading order is thrown with the same message as before, and downgraded verdicts are still logged *before* a fatal one so a DART keeps the context.

Both repair steps log a distinct `managed_acl_self_heal` marker — worth alerting on, unlike an ancestor advisory, because it means AVC and DefenseClaw are contending for one DACL.

Also applied to `Set-DefenseClawPreservedStateAcls`, which stamps the same shared root while preserving state across an uninstall.

macOS needs no equivalent: `install.sh` applies chown/chmod with no stamp-then-verify equality assertion, so there's no race to lose.

### Review emphasis

This commit is **PowerShell-only and unexecuted** — still no `pwsh` on the authoring host. I verified mechanically that bracket balance is unchanged from the baseline, no backtick continuation has trailing whitespace, and each new function is defined once and referenced from the expected call sites. That is not a substitute for a parse. Please prioritise:

- a `pwsh` parse / PSScriptAnalyzer pass on the module;
- the `switch`-in-assignment and `[Collections.Generic.List[object]]` usage in `Get-DefenseClawPathAclVerdicts`;
- `packaging/windows/tests/enterprise-uninstall-transaction-smoke.ps1` — its `Assert-DefenseClawPathAcl` mock gained the new parameters (plus `-AllowedOwnerSIDs`, which it was already missing) so it keeps binding the real call sites.

PR #885, opened 2026-09-19
Send a PreToolUse invocation to the managed lane as a tool call, not as flattened text

elivshit-cisco:
Today a tool call from PreToolUse is sent as free text. `tool_input` is marshalled into a string, `"Tool call: <name>"` is added in front, and the result goes out as a user message. Both inspect clients then drop everything except `role` and `content`.

So a cloud rule that reads `tool_calls[].function.arguments` cannot be evaluated. It finds no tool calls, and the outcome looks the same as "inspected and allowed".

This PR sends the tool call as structure:

```json
{"messages":[{"role":"assistant","content":"",
  "tool_calls":[{"id":"toolu_01U3wpX…","type":"function",
    "function":{"name":"Bash","arguments":"{\"command\": \"rm -rf /\"}"}}]}]}
```

Three details matter here. The role must be `assistant`, because a `tool_calls` array on a user message is not read. `arguments` must be a JSON string, not an object. And `tool_use_id` becomes the tool call id.

`claudeCodeToolArgs` and `codexToolArgs` already return these bytes. They were converted into a string one call later.

### What changed

- both clients carry `tool_calls` on the wire: the managed one and the API-key one
- both deployment modes send the structured form, not only `managed_enterprise`

The scope is small. Only `Direction == "tool_call"` uses the new path. PostToolUse, prompts and events still flatten into content as before. So a `rm -rf /` that a tool only read does not become a command finding.

### Testing

Unit tests pin the wire shape for both clients and both modes: assistant role, the `tool_calls` array, and `arguments` readable as JSON. One test also checks that a message without tool calls is unchanged.

The full `internal/gateway` suite gives the same result as `main`.

It was also tested against a SAM with a real policy. A dangerous shell command was blocked and attributed to the expected rule, and benign commands were allowed.


vineethsai7:
## The diagnosis is right, but the fix is one layer too shallow

The root-cause analysis here is correct and worth keeping: a cloud rule reading `tool_calls[].function.arguments` can't be evaluated today, and the failure is indistinguishable from "inspected and allowed." That's a real hole. `claudeCodeToolArgs` / `codexToolArgs` already producing the right bytes is a nice catch, and the `toolUseID` threading is clean.

But the PR moves DefenseClaw from a representation where **vendor naming didn't matter** (free text, read by a classifier) to one where **vendor naming is load-bearing** (structured fields, read by field selectors) — without adding any normalization. Cross-connector uniformity was never required before; this change requires it and doesn't provide it.

I ran the branch against our own golden payloads in `scripts/live-connector-e2e/golden/` (throwaway probe, not committed). This is what actually reaches AID:

| connector | `tool_calls[].id` | `function.name` | `function.arguments` |
|---|---|---|---|
| claudecode | **`""`** | `Bash` | `{"command":…}` |
| amp | `dc-e2e-amp-tool-allow` (real) | `shell` | `{"command":…}` |
| opencode | `dc-e2e-call-allow` (real) | `bash` | `{"command":…}` |
| cursor | minted UUIDv7 | `run_terminal_cmd` | `{"command":…}` |
| hermes | minted UUIDv7 | `execute_command` | `{"command":…}` |
| openhands | minted UUIDv7 | `terminal` | `{"command":…}` |
| devin | minted UUIDv7 | `exec` | `{"command":…}` |
| antigravity | minted UUIDv7 | `run_command` | **`{"CommandLine":…,"Cwd":…}`** |

### Blocking issues

**1. `function.arguments` key names are vendor-specific, so `arguments.command` does not work everywhere.** Antigravity sends `CommandLine` / `Cwd`. A cloud rule written as `arguments.command` silently misses it — the exact same silent-miss failure mode this PR set out to eliminate, relocated one level down. The new test encodes this divergence in its `argsKey` field rather than surfacing it as a problem.

**2. `function.name` is the raw vendor name — nine spellings for "run a shell command."** `Bash`, `shell`, `bash`, `run_terminal_cmd`, `RunShellCommand`, `execute_command`, `terminal`, `exec`, `run_command`. Every cloud rule keyed on tool name has to enumerate all of them, case-sensitively, and re-enumerate whenever a vendor renames a tool or we add a connector. Worth noting the new test's tool names don't match the goldens (`shell` vs `run_terminal_cmd` for cursor, `run_shell_command` vs `RunShellCommand` for geminicli, `execute_bash` vs `terminal` for openhands) — because the test supplies `tool_name` itself. So the test proves the plumbing, not real-payload fidelity.

**3. `arguments` is sometimes the entire hook envelope.** `agent_hook.go:1594-1596` falls back to `args = payload` when none of `tool_input / toolInput / tool_args / toolArgs / args / arguments / tool_info / toolInfo` is present. Probed with a flat native windsurf `pre_run_command`:

```
name="tool"  args={"command":"rm -rf /","cwd":"/repo","execution_id":"e1",
                   "hook_event_name":"pre_run_command","session_id":"s1"}
```

`function.name` becomes the literal string `"tool"`, and session/execution identifiers are presented to AID as authoritative tool arguments. As free text this was harmless; as structured data it's wrong, and it leaks correlation identifiers into a field that means something specific.

**4. Content is emptied rather than supplemented.** `hookAIDInspectToolCall` sends `{Role: "assistant", ToolCalls: …}` with no `Content`, so the wire carries `"content": ""` (API-key lane) / `"content": {"text": ""}` (managed lane). This *replaces* the text form instead of adding to it. Any AID rule that is a text classifier over content — PII, secrets, prompt injection in tool arguments — now sees an empty string. In `managed_enterprise` the AID lane is the only lane, so if AID's classifiers don't descend into `function.arguments`, this is a silent detection regression on the surface where we have no local fallback. The SAM test confirmed one shell-command policy blocks; that doesn't cover the classifier rule types. Could you confirm with a PII/injection rule on tool args, before and after?

The OpenAI chat schema permits `content` alongside `tool_calls`. Populating both makes the change purely additive and removes this risk entirely.

**5. `tool_calls[].id` is not a usable correlation key, and empty ids ship as `"id": ""`.** Only amp and opencode carry the agent's real id; six connectors get a DefenseClaw-minted UUIDv7; claudecode and windsurf goldens emit empty. `hookAIDInspectToolCall` writes `"id"` into the map unconditionally, so an absent id becomes `"id": ""` rather than being omitted. For claudecode `PermissionRequest` this is permanent — our own manifest records it: *"PermissionRequest has no tool_use_id, so it is a direct decision surface and cannot create pending state."* Since a minted id is indistinguishable on the wire from a real one, a cloud rule can't tell whether it's looking at a stable agent identifier or a synthetic one.

### Smaller items

- **Managed and non-managed lanes disagree on when to use the structured form.** `inspectTrustedToolPolicyCtx` gates on `req.Direction == "tool_call" && len(req.Args) > 0`, but `hookAIDInspectTool` ignores `Direction` entirely. A `POST /api/v1/inspect/tool` with `tool` + `args` and no `direction` field gets structured in non-managed mode and flattened text in managed mode. The PR description's "Only `Direction == "tool_call"` uses the new path" holds for the managed lane only.
- **`aidToolCall` is spliced into another function's doc comment** (`inspect.go:222`). Godoc now attributes the `inspectManagedAIDOnly` paragraph to `aidToolCall`, and `inspectManagedAIDOnly` is left undocumented. Move the type above the comment block.
- The stale comment at the second `hookAIDInspectTool` call site ("We send the rule reasons text rather than just the args because AID's classifier reads free-text content") is now doubly wrong and should be updated in the same pass.

### The contract gap underneath all of this

There is no declaration anywhere of *how* a surface is encoded on the wire. `HookContract.AIDSurfaces` (`hook_contract.go:68`, mirrored as `aid_surfaces` in `cli/defenseclaw/inventory/hook_contracts.json`, parity-tested at `hook_contract_test.go:774`) declares *which* surfaces go to AID — `prompt`, `tool_call`, `tool_result`, `event_content` — but says nothing about their shape. `aid_surfaces` reads identically before and after this PR while the bytes change completely, so no test or review gate can catch this class of change.

Worth knowing while you're in here: `AIDSurfaces` is declarative only. `HookProfileAIDSurfaceEnabled` has exactly one caller in the repo and it's a test (`hook_contract_test.go:876`); the send path never consults it. Windsurf and kiro have no entry in `builtinHookContracts` at all, so their `AIDSurfaces` is empty — and this PR's own new test shows both reaching AID. So the registry currently documents our AI Defense contract rather than enforcing it.

### Suggested path

1. Populate both `content` (existing flattened `"Tool call: <name>\n<args>"`) and `tool_calls`. Structured rules become possible; nothing that works today stops working. This alone makes the PR safe to land.
2. Gate the structured form on a trustworthy args projection. Set an explicit flag at the `args = payload` fallback site (`agent_hook.go:1594`) and skip `tool_calls` when it's set — the existing text path already handles this. `ToolArgsProjectionUncertain` is adjacent but not sufficient: it only fires on an *empty* projection, not on the whole-payload fallback.
3. Omit `id` when empty instead of sending `"id": ""`, and don't substitute a minted UUID for an absent vendor id on this surface — a synthetic id that looks real is worse for rule authors than no id.

**The actual fix** — normalize at the registry boundary, which is already the per-vendor shape authority (it owns `ContentEnvelopeKey`, `ResponseFieldName`, the `Correlation` bindings, `ToolCallLifecycle`):

4. Add a per-connector, per-contract-version tool-call normalization spec to `HookContract`: canonical tool name aliases and canonical argument-key aliases. Mirror it into `hook_contracts.json` so the existing parity test covers it, and emit **both** canonical and vendor-native on the wire (e.g. canonical `function.name` with the raw name preserved alongside). Then a cloud rule is written once against the canonical shape and works on every connector, including ones added later.
5. Reuse the taxonomies we already have rather than inventing one: `connector.ToolSurface` is a closed vocabulary (`shell`, `file-read`, `file-write`, `file-edit`, `mcp`, `skills`, `generic`) and `guardrail.ClassifyToolName` already exists for exactly this purpose — its doc comment says it should work "whether the tool was named `run_shell`, `bash`, or anything else." It currently classifies only 3 of the 10 names in the table above (`Bash`, `bash`, `execute_command`); `shell`, `run_terminal_cmd`, `RunShellCommand`, `terminal`, `exec` and `run_command` all return `CapUnknown`. Extending it is probably a prerequisite and may be worth its own PR.
6. Declare the wire shape in the contract (an `aid_wire_version` on `HookContract`, or richer `aid_surfaces` entries) so the manifest parity test catches the next encoding change automatically.

**On test coverage**: the new per-connector table is a genuinely good idea and I'd like to keep it, but it should be driven by the golden payloads in `scripts/live-connector-e2e/golden/` rather than synthesized `tool_name` / `tool_input`. As written it asserts our own inputs back to us — every real divergence in the table above passes the test today.

Verified on the branch: `go vet ./internal/gateway/` clean, all 7 new tests pass (14/14 connector subtests), `-run 'Managed|AID|CiscoInspect|DefenseClawInspect'` green. The failures above are all behavioral, not build or test failures.

vineethsai7:
You must update the Connector registry with the specific hook events and the AI defense contract @elivshit-cisco

